Issue #22, opened 2025-07-25
Support AArch64


ggguoge:
**The feature you'd like**
Build, compile and run

**Implementation you've considered**

**Additional context**

feilongjiang:
Closed issue as #43 was integrated.

PR #43, opened 2025-08-16
feat: Initial support for AArch64

feilongjiang:
## Related issue(s):

issue #22 

This PR contains the initial support for linux-aarch64 for jeandle-jdk. 
Including the following changes:
1. Add the linux-aarch64 workflow, both building and testing. Here is my [local workflow test](https://github.com/feilongjiang/jeandle-jdk/actions/runs/17227185438).
2. Enable `jvm-feature` detection for linux-aarch64.
3. Implement platform-specific methods like `emit_static_call_stub`, `patch_static_call_site`, etc..
4. Make `setup_frame_size` implementation platform dependent, since the `frame_size` read from the `.stack_sizes` section is different on x86 and AArch64.
5. To resolve the macro conflict on `AARCH64`, I added a workaround that employs `push_macro` and `pop_macro` directives around the relevant LLVM includes.

Testing:
- [x] tier1_compiler_jeandle with slowdebug build on linux-aarch64 & linux-x64
- [x] tier1_compiler_jeandle with release build on linux-aarch64 & linux-x64

### DIFF
--- a/jeandle-docs/getting-started.md
+++ b/jeandle-docs/getting-started.md
@@ -19,7 +19,7 @@ cd jeandle-llvm
 mkdir build
 cd build
 cmake -G "Unix Makefiles" -DLLVM_TARGETS_TO_BUILD=X86 -DCMAKE_BUILD_TYPE="Release" -DCMAKE_INSTALL_PREFIX="/home/jeandle-llvm-install" -DLLVM_BUILD_LLVM_DYLIB=On -DLLVM_DYLIB_COMPONENTS=all ../llvm
-cmake --build . --target install
+cmake --build . --target install --parallel
 ```
 
 3. Clone jeandle-jdk:


src/hotspot/cpu/x86/jeandleCompiledCode_x86.cpp
```diff
@@ -0,0 +1,39 @@
...
+#include "jeandle/jeandleCompilation.hpp"
+
+// Get the frame size from .stack_sizes section.
+void JeandleCompiledCode::setup_frame_size() {
+  SectionInfo section_info(".stack_sizes");
+  if (!ReadELF::findSection(*_elf, section_info)) {
+    JeandleCompilation::report_jeandle_error(".stack_sizes section not found");
+    return;
+  }
+  llvm::DataExtractor data_extractor(llvm::StringRef(((char*)_obj->getBufferStart()) + section_info._offset, section_info._size),
```
Cruise20, line 31: Replace every `oopSize` with `BytesPerWord`, both in x86 and aarch64.

feilongjiang: Thanks for the review! Fixed.

src/hotspot/cpu/aarch64/nativeInst_aarch64.cpp
```diff
@@ -418,6 +418,11 @@ bool NativeInstruction::is_adrp_at(address instr) {
   return (Instruction_aarch64::extract(insn, 31, 24) & 0b10011111) == 0b10010000;
 }
 
+bool NativeInstruction::is_ldr_unsigned_at(address instr) {
+  unsigned insn = *(unsigned*)instr;
+  return (Instruction_aarch64::extract(insn, 31, 22) & 0b0011101101) == 0b0011100101;
```
feilongjiang, line 423: For instruction encoding reference: 
1. [LDR (immediate)](https://developer.arm.com/documentation/ddi0602/2025-06/Base-Instructions/LDR--immediate---Load-register--immediate--?lang=en#:~:text=64\)%3B-,Unsigned%20offset,-31)
2. [LDR (immediate, SIMD&FP)](https://developer.arm.com/documentation/ddi0602/2025-06/SIMD-FP-Instructions/LDR--immediate--SIMD-FP---Load-SIMD-FP-register--immediate-offset--?lang=en#:~:text=Unsigned%20offset%0A\(FEAT_FP\))

.github/workflows/build-linux.yml
```diff
@@ -66,7 +66,7 @@ on:
 jobs:
   build-linux:
     name: build
-    runs-on: ubuntu-22.04
+    runs-on: ${{ inputs.platform == 'linux-aarch64' && 'ubuntu-22.04-arm' || 'ubuntu-22.04' }}
```
taofengliu: Does this mean that the action will run on any runner that match this expression?
Is it possible that 'ubuntu-22.04' is still true when the platform is 'linux-aarch64'?

src/hotspot/share/compiler/compileBroker.cpp
```diff
@@ -87,7 +87,10 @@
 #include "opto/c2compiler.hpp"
 #endif
 #ifdef JEANDLE
+#pragma push_macro("AARCH64")
```
taofengliu, line 90: Just like we discussed before, perhaps we can solve this issue together with the assert macro problem through a separate header file.

test/hotspot/jtreg/ProblemList.txt
```diff
@@ -79,6 +79,8 @@ compiler/codecache/CheckLargePages.java 8319795 linux-x64
 compiler/floatingpoint/TestSubnormalFloat.java 8317810 generic-i586
 compiler/floatingpoint/TestSubnormalDouble.java 8317810 generic-i586
 
+compiler/c2/cr7200264/TestIntVect.java 8325277 linux-aarch64
```
taofengliu, line 82: It is a known issue for aarch64, right?

feilongjiang: Yes, here is the reported bug: https://bugs.openjdk.org/browse/JDK-8325277. It's still unresolved.

.github/workflows/build-linux.yml
feilongjiang: It checks if the input parameter platform is set to 'linux-aarch64'. If so, it chooses the runner labeled 'ubuntu-22.04-arm', which is suitable for ARM64 architecture. Otherwise, it defaults to 'ubuntu-22.04', which is the standard x86_64 Ubuntu runner.
I can try `matrix` instead if this checking looks a bit confusing.

taofengliu: Let's try matrix !

feilongjiang: Seems that it is more convenient to add a "runs-on" input to the build workflow. 
Here is my local action test in a separate branch: https://github.com/feilongjiang/jeandle-jdk/actions/runs/17260704082

src/hotspot/share/compiler/compileBroker.cpp
feilongjiang: I have created #52 to track the issue. Let's do the refactoring in a separate PR.

taofengliu (approved):

Issue #7, opened 2025-07-24
Support safepoint poll


taofengliu:
**The feature you'd like**
Add safepoint polls in appropriate code locations and implement safepoint poll.

**Implementation you've considered**
Use ```poll_word``` to do polls.
Implement it as a JavaOp like:
```
define void @jeandle.safepoint_poll() "lower-phase"="0" {
entry:
  %poll_word_value = load volatile i64, ptr poll_word_addr
  %0 = icmp eq i64 %poll_word_value, 0
  br i1 %0, label %return, label %do_safepoint, !prof !2

return:                                      ; preds = %do_safepoint, %entry
  ret void

do_safepoint:                                     ; preds = %entry
  call void @"do_safepoint"()
  br label %return
}
```

**Additional context**

PR #46, opened 2025-08-20
feat: Support safepoint poll

taofengliu:
## Related issue(s):

issue #7

The safepoint poll is defined as JavaOp, and is inlined in the later stage of optimization.

I use the thread local poll word to determine whether to enter a safepoint, and leverage a runtime stub to transfer execution to the C++-implemented safepoint handler.

Currently, only safepoint polls for back edges.

### DIFF
--- a/src/hotspot/cpu/x86/jeandleAssembler_x86.cpp
+++ b/src/hotspot/cpu/x86/jeandleAssembler_x86.cpp
@@ -128,6 +128,34 @@ void JeandleAssembler::emit_oop_reloc(uint32_t offset, jobject oop_handle) {
   __ code_section()->relocate(at_address, rspec, __ disp32_operand);
 }
 
+void JeandleAssembler::patch_call_vm(uint32_t operand_offset, address target) {
+  assert(operand_offset != 0, "invalid operand address");
+
+  address call_pc = __ addr_at(operand_offset - 1);
+
+  // Set insts_end to where to patch.
+  address insts_end = __ code()->insts_end();
+  __ code()->set_insts_end(call_pc);
+
+  // Patch.
+  __ call(AddressLiteral(target, relocInfo::static_call_type));
+
+  // Recover insts_end.
+  __ code()->set_insts_end(insts_end);
+}
+
+uint32_t JeandleAssembler::fixup_call_inst_offset(uint32_t offset) {
+  return offset + 4;
+}
+
 LinkKind JeandleAssembler::get_oop_reloc_kind() {
   return LinkKind_x86_64::RequestGOTAndTransformToPCRel32GOTLoadREXRelaxable;
+}
+
+LinkKind JeandleAssembler::get_call_vm_link_kind() {
+  return LinkKind_x86_64::BranchPCRel32;
+}
+
+LinkKind JeandleAssembler::get_const_link_kind() {
+  return LinkKind_x86_64::Delta32;
 }
--- a/src/hotspot/cpu/x86/jeandleAssembler_x86.cpp
+++ b/src/hotspot/cpu/x86/jeandleAssembler_x86.cpp
@@ -152,10 +152,10 @@ bool JeandleAssembler::is_oop_reloc_kind(LinkKind kind) {
   return kind == LinkKind_x86_64::RequestGOTAndTransformToPCRel32GOTLoadREXRelaxable;
 }
 
-bool JeandleAssembler::is_call_vm_link_kind(LinkKind kind) {
+bool JeandleAssembler::is_call_vm_reloc_kind(LinkKind kind) {
   return kind == LinkKind_x86_64::BranchPCRel32;
 }
 
-bool JeandleAssembler::is_const_link_kind(LinkKind kind) {
+bool JeandleAssembler::is_const_reloc_kind(LinkKind kind) {
   return kind == LinkKind_x86_64::Delta32;
 }
--- a/src/hotspot/cpu/x86/jeandleAssembler_x86.cpp
+++ b/src/hotspot/cpu/x86/jeandleAssembler_x86.cpp
@@ -148,14 +148,14 @@ uint32_t JeandleAssembler::fixup_call_inst_offset(uint32_t offset) {
   return offset + 4;
 }
 
-LinkKind JeandleAssembler::get_oop_reloc_kind() {
-  return LinkKind_x86_64::RequestGOTAndTransformToPCRel32GOTLoadREXRelaxable;
+bool JeandleAssembler::is_oop_reloc_kind(LinkKind kind) {
+  return kind == LinkKind_x86_64::RequestGOTAndTransformToPCRel32GOTLoadREXRelaxable;
 }
 
-LinkKind JeandleAssembler::get_call_vm_link_kind() {
-  return LinkKind_x86_64::BranchPCRel32;
+bool JeandleAssembler::is_call_vm_link_kind(LinkKind kind) {
+  return kind == LinkKind_x86_64::BranchPCRel32;
 }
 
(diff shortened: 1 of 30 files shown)


src/hotspot/share/jeandle/jeandleCompiledCode.cpp
```diff
@@ -200,11 +217,27 @@ void JeandleCompiledCode::resolve_reloc_info(JeandleAssembler& assembler) {
...
+        address target_addr = JeandleRuntimeRoutine::get_stub_entry(*target.getName());
+        JeandleCallVMReloc* call_vm_reloc = new JeandleCallVMReloc(*block, edge, target_addr);
+
+        uint32_t call_inst_offset = JeandleAssembler::fixup_call_inst_offset(call_vm_reloc->offset());
+
+        // TODO: Set the right bci.
+        _safepoints[call_inst_offset] = new CallSiteInfo(0/* statepoint_id */, JeandleJavaCall::STATIC_CALL, target_addr, 0/* bci */);
+        relocs.push_back(call_vm_reloc);
+
+      } else if (target.isDefined() && edge.getKind() == JeandleAssembler::get_const_link_kind()) {
```
Cruise20: Add the else branch to handle the relocation kinds we did not anticipate: skip those that can be ignored, and intercept those that are unexpected.

src/hotspot/cpu/x86/jeandleAssembler_x86.cpp
```diff
@@ -128,6 +128,34 @@ void JeandleAssembler::emit_oop_reloc(uint32_t offset, jobject oop_handle) {
...
+  return offset + 4;
+}
+
 LinkKind JeandleAssembler::get_oop_reloc_kind() {
   return LinkKind_x86_64::RequestGOTAndTransformToPCRel32GOTLoadREXRelaxable;
 }
+
+LinkKind JeandleAssembler::get_call_vm_link_kind() {
+  return LinkKind_x86_64::BranchPCRel32;
+}
```
feilongjiang: It would be better to return a boolean type here.
For AArch64, the single relocation item may contain multiple kinds [1].

https://github.com/jeandle/jeandle-jdk/pull/43/files#diff-b4f785a3efaa27650aa3d8d0d7cac37b81a8a4d7470d6c89bcd109f42d44af96R129-R132

src/hotspot/cpu/x86/jeandleAssembler_x86.cpp
```diff
@@ -128,6 +128,34 @@ void JeandleAssembler::emit_oop_reloc(uint32_t offset, jobject oop_handle) {
...
+
+LinkKind JeandleAssembler::get_const_link_kind() {
+  return LinkKind_x86_64::Delta32;
```
feilongjiang: Same here, we can just check the type.

src/hotspot/share/jeandle/templatemodule/jeandleRuntimeDefinedJavaOps.cpp
```diff
@@ -0,0 +1,135 @@
...
+ *
+ */
+
+#include <cassert>
+#include "llvm/IR/BasicBlock.h"
+#include "llvm/IR/IRBuilder.h"
+
+#include "jeandle/templatemodule/jeandleRuntimeDefinedJavaOps.hpp"
+#include "jeandle/jeandleRuntimeRoutine.hpp"
+#include CPU_HEADER(jeandleRegister)
```
feilongjiang: It would be better to add a new `jeandleRegister.hpp` file and place the `CPU_HEADER` macro within it, just as `register.hpp` does.

src/hotspot/share/jeandle/jeandleCallVM.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+  // Set the last_Java_sp to enable stack unwind.
+  llvm::Value* last_Java_sp_ptr = ir_builder.CreateIntToPtr(ir_builder.getInt64((uint64_t)JavaThread::last_Java_sp_offset()),
+                                                            llvm::PointerType::get(context, llvm::jeandle::AddrSpace::TLSAddrSpace));
+  llvm::MDNode* sp_register = llvm::MDNode::get(context, {llvm::MDString::get(context, JeandleRegister::get_stack_pointer())});
+  llvm::Value* read_sp_args[] = {llvm::MetadataAsValue::get(context, sp_register)};
+  llvm::CallInst* sp_value = ir_builder.CreateIntrinsic(llvm::Intrinsic::read_register,
+                                                        intptr_type,
+                                                        read_sp_args);
+  ir_builder.CreateStore(sp_value, last_Java_sp_ptr);
```
feilongjiang, line 54: IIUC, `generate_call_VM` is similar to `call_VM_base`, should we also store `last_java_pc`, `last_java_fp` if these fields are valid?

src/hotspot/share/jeandle/jeandleCallVM.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+  // Make arguments.
+  llvm::SmallVector<llvm::Value*> args;
+  for (llvm::Value& arg : llvm_func->args()) {
```
feilongjiang, line 58: Do we need to push `java_thread` to `c_rarg0` as `call_VM_base` did?

src/hotspot/share/jeandle/jeandleRuntimeRoutine.hpp
```diff
@@ -0,0 +1,69 @@
...
+#include "utilities/debug.hpp"
+#include "memory/allStatic.hpp"
+#include "runtime/javaThread.hpp"
+#include "utilities/globalDefinitions.hpp"
+
+//------------------------------------------------------------------------------------------------------
+//   |     c_func      |       return_type             |                    args_type
+//------------------------------------------------------------------------------------------------------
+#define ALL_JEANDLE_ROUTINES(def)                                                                                                     \
+  def(safepoint_handler, llvm::Type::getVoidTy(context), {llvm::PointerType::get(context, llvm::jeandle::AddrSpace::CHeapAddrSpace)})
```
feilongjiang, line 38: Seems we do not pass JavaThread to C/C++ functions, that's why `call_VM` didn't push JavaThread as the first arg, right?

src/hotspot/share/jeandle/jeandleCallVM.cpp
taofengliu: Just according to the C2 compiler's runtime stub, storing last_java_sp is enough.

src/hotspot/share/jeandle/jeandleCallVM.cpp
taofengliu: The current_thread is not always the first argument for us. It depends on the C++ function definitions.

src/hotspot/share/jeandle/jeandleRuntimeRoutine.hpp
taofengliu: Yes.

feilongjiang: Ah, I see. JavaThread was passed as a normal arg to the C functions, so we don't have to push JavaThread to c_rarg0 as it is already in the arg list.

src/hotspot/share/jeandle/templatemodule/jeandleRuntimeDefinedJavaOps.cpp
feilongjiang:
```suggestion
#include "jeandle/jeandleRegister.hpp"
```

Cruise20 (approved):

Cruise20 (approved):

Issue #50, opened 2025-08-27
We should use --parallel parameter when build jeandle-llvm


sendaoYan:
**Goals**
Update the getting-started.md, use --parallel to accelator the jeandle-llvm build

**Motivation**

**Description**

I found that building the jeandle-llvm is very slowly, because cmake only use 1 core cpu when execute the below cmake command:

```shell
cmake --build . --target install
```

I think it's better to use `--parallel` parameter explicitly to accelator the jeandle-llvm build.

Env info:

```
> cat /etc/os-release 
PRETTY_NAME="Ubuntu 22.04.5 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.5 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy

> cmake --version
cmake version 3.22.1

CMake suite maintained and supported by Kitware (kitware.com/cmake).
```

PR #51, opened 2025-08-27
docs: Use --parallel parameter when build jeandle-llvm

sendaoYan:
## Related issue(s):

issue #50

Update the getting-started.md about how the build the jeandle-llvm


taofengliu (commented):
Good idea! We can also make the compilation of jeandle-jdk parallel in this document.

sendaoYan:
> We can also make the compilation of jeandle-jdk parallel in this document.

In jeandle-jdk, configure will pick up the number of cpu cores automatically, and then build the jdk parallelly by default. See https://github.com/jeandle/jeandle-jdk/blob/main/make/autoconf/build-performance.m4#L134

So I think we can leave the doc compilation of jeandle-jdk unchange.

taofengliu (approved):

Issue #59, opened 2025-09-02
Support inline intrinsic functions


kuaiwei:
When jeandle compiler parse bytecode to invoke some java functions, it can recognize functions which can be inlined as intrinsic code.

**Additional context**
We can start from some simple functions like Math.abs(double), Math.sqrt(double), and add more intrinsic in future.

PR #69, opened 2025-09-08
feat: Support inline intrinsic functions

kuaiwei:
## Related issue(s):

issue #59

Initial support for inline intrinsic functions.
During abstract interpreter, jeandle can recognize methods which can be converted as llvm intrinsic.
Implement intrinsic for Math.abs(double) and Math.abs(float)

Testing:
- test/hotspot/jtreg/compiler/jeandle/intrinsic/TestAbsDouble.java
- test/hotspot/jtreg/compiler/jeandle/intrinsic/TestAbsFloat.java

### DIFF
--- a/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
+++ b/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
@@ -26,6 +26,7 @@
 #include "jeandle/jeandleType.hpp"
 #include "jeandle/jeandleUtils.hpp"
 
+#include "logging/log.hpp"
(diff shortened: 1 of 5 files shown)


taofengliu:
Seems we need a test tool to check LLVM IR, so I created an issue #70 . This will be implemented these days, and then we can refactor the test in this PR.

taofengliu:
#71 has been merged, we can refactor the test now. And the dco check failed, we should use ```git commit -s``` to commit code.

kuaiwei:
It merged other commits. Close it and create a new PR.

PR #77, opened 2025-09-09
feat: Implement verified entry

minchao1024:
## Related issue(s):

issue #5 

1. add `patchable-function-entry` attribute to each java method, which reserved extra space for `patch_verified_entry` function.
2. add `NoCfCheck` attribute to each compilation, which can prevent `ENDBR` from being generated at the beginning of the code in x86_64 .


src/hotspot/share/jeandle/jeandleCompilation.cpp
```diff
@@ -282,6 +284,19 @@ void JeandleCompilation::compile_module() {
...
+    func.addFnAttr(llvm::Attribute::NoCfCheck);
+    if (_method != nullptr) {
+#if defined(AMD64)
+      func.addFnAttr("patchable-function-entry", "5");
+#elif defined(AARCH64)
+      func.addFnAttr("patchable-function-entry", "1");
+#endif
+    }
+  }
+}
```
taofengliu: We don't need to add these two attributes for every function.

test/hotspot/jtreg/compiler/jeandle/TestMakeNotEntrant.java
```diff
@@ -0,0 +1,129 @@
...
+            if (add(1, 2) != 3) {
+                Asserts.fail();
+            }
+        }
+    }
+
+    public static int add(int a, int b) {
+        return a + b;
+    }
+}
```
taofengliu: Add a blank line here.

src/hotspot/share/jeandle/jeandleCompilation.cpp
```diff
@@ -282,6 +284,19 @@ void JeandleCompilation::compile_module() {
   _code.install_obj(std::move(object));
 }
 
+void JeandleCompilation::add_llvm_attributes() {
+  for (auto& func : _llvm_module->functions()) {
+    func.addFnAttr(llvm::Attribute::NoCfCheck);
+    if (_method != nullptr) {
+#if defined(AMD64)
+      func.addFnAttr("patchable-function-entry", "5");
```
feilongjiang: Instead of hard-coding the entry size with platform guard here, maybe we can add a method like `get_verified_entry_size` for each platform.

src/hotspot/share/jeandle/jeandleCompilation.cpp
minchao1024: Fixed, only added to Java method and call_VM func.

test/hotspot/jtreg/compiler/jeandle/TestMakeNotEntrant.java
minchao1024: Done

src/hotspot/share/jeandle/jeandleCompilation.cpp
minchao1024: Fixed

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -435,6 +435,8 @@ JeandleAbstractInterpreter::JeandleAbstractInterpreter(ciMethod* method,
                                                        _ir_builder(_block_builder->entry_block()->llvm_block()),
                                                        _statepoint_id(0),
                                                        _oop_idx(0) {
+  _llvm_func->addFnAttr(llvm::Attribute::NoCfCheck);
+  _llvm_func->addFnAttr("patchable-function-entry", std::to_string(get_verified_entry_size()));
```
taofengliu: It's better to move these two lines to JeandleFuncSig::setup_description, and provide a arch relative implementation

test/hotspot/jtreg/compiler/jeandle/TestMakeNotEntrant.java
```diff
@@ -0,0 +1,132 @@
...
+        Thread loopThread = new Thread(() -> {
+            loopTest();
+        });
+        loopThread.setDaemon(true);
+        loopThread.start();
+    }
+
+    public static void loopTest() {
+        while (!finish) {
+            add(2, 2); // update return value
```
taofengliu: Add more comments to explain why we do this.

test/hotspot/jtreg/compiler/jeandle/TestMakeNotEntrant.java
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        Method m = getMethod("add", int.class, int.class);
+        int count = makeNotEntrantCount;
+        while(count > 0) {
+            if (wb.isMethodCompiled(m)) {
+                makeNotEntrant(m);
+                count--;
+            }
+        }
```
taofengliu, line 95: I think we should figure out why Jeandle's behavior differs from C2

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
minchao1024: Fixed

test/hotspot/jtreg/compiler/jeandle/TestMakeNotEntrant.java
minchao1024: done

test/hotspot/jtreg/compiler/jeandle/TestMakeNotEntrant.java
minchao1024: `make_not_entrant` will guide thread to `SharedRuntime::handle_wrong_method`, which will recompile method automatically, no need to enqueue method to CompileQueue.
The reason why C2 compiler needs enqueue method is that C2 will inline the `add` method, so `make_not_entrant` will not affect the `loopTest` method. 

Add dontinline compilepolicy into the test `TestMakeNotEntrant.java` already.

minchao1024:
To patch verified_entry, BranchTarget instructions must be prevented from being generated.
1、In x86_64, `endbr` can be avoid by adding `NoCfCheck` attribute.
2、In aarch64, to avoid `bti`, we must ensure that the function is not marked with `branch-target-enforcement` attribute, as `"branch-target-enforcement"="false"` was removed.
https://github.com/jeandle/jeandle-llvm/blob/main/llvm/lib/Target/AArch64/AArch64MachineFunctionInfo.cpp#L100

test/hotspot/jtreg/compiler/jeandle/fileCheck/TestFileCheck.java
```diff
@@ -39,7 +39,7 @@ public static void main(String[] args) throws Exception {
             FileCheck fileCheck = new FileCheck(currentDir,
                                                 TestFileCheck.class.getDeclaredMethod("add", int.class, int.class),
                                                 false);
-            fileCheck.check("define hotspotcc i32 @\"compiler_jeandle_fileCheck_TestFileCheck_add_(II)I\"(i32 %0, i32 %1) gc \"hotspotgc\" {");
+            fileCheck.check("define hotspotcc i32 @\"compiler_jeandle_fileCheck_TestFileCheck_add_(II)I\"(i32 %0, i32 %1) #3 gc \"hotspotgc\" {");
```
taofengliu:
```suggestion
            fileCheck.check("define hotspotcc i32 @\"compiler_jeandle_fileCheck_TestFileCheck_add_(II)I\"(i32 %0, i32 %1)");
```
We can simply don't check the following attribute.

test/hotspot/jtreg/compiler/jeandle/TestMakeNotEntrant.java
```diff
@@ -0,0 +1,148 @@
...
+        loopThread.start();
+    }
+
+    public static void loopTest() {
+        while (!finish) {
+            /* patching verified_entry without reserved 5-byte header in x86_64, the following cases will occur.
+             * 1、SIGILL: exit_value not zero, can be observed.
+             * 2、SIGSEGV: exit_value not zero, can be observed.
+             * 3、Nothing happened but hurt header code. In "TestMakeNotEntrant::add" method, it will lose the add operation.
+             *    Thus, modify the return value, and then check it.
```
taofengliu:
```suggestion
            /* patching verified_entry without reserved 5-byte header in x86_64, the following cases will occur.
             * 1、SIGILL: The exit value of current process is not zero, can be observed.
             * 2、SIGSEGV: The exit value of current process is not zero, can be observed.
             * 3、Nothing happened but hurt header code. "TestMakeNotEntrant::add" method will lose the add operation.
             *    Thus, change the return value, and then check it.
```
I think this is a more clear version. But I still don't understand why we should change the return value. If patching verified_entry without reserved 5-byte, what will happen to the add instruction? Please explain it more.

src/hotspot/share/jeandle/jeandleCallVM.cpp
```diff
@@ -37,6 +37,7 @@ void JeandleCallVM::generate_call_VM(const char* name, address c_func, llvm::Fun
                                                      target_module);
   llvm_func->setCallingConv(llvm::CallingConv::Hotspot_JIT);
   llvm_func->setGC(llvm::jeandle::JeandleGC);
+  llvm_func->addFnAttr(llvm::Attribute::NoCfCheck);
```
taofengliu:
```suggestion
  JeandleFuncSig::setup_description(llvm_func, true /*is_stub*/);
```
Maybe we can call setup_description here. But for stub functions, we don't need to add patchable-function-entry, so we can add an argument for setup_description to indicate if we are setting up a stub function.
By this way, we can make all llvm function setup process in one place.

taofengliu (commented):
It looks like we're almost done. Here are a few minor suggestions that we might want to consider.

test/hotspot/jtreg/compiler/jeandle/TestMakeNotEntrant.java
minchao1024: The compiled code of Method "TestMakeNotEntrant::add" is
```
  0x0000:   push   %rbp
  0x0001:   lea    (%rsi,%rdx,1),%eax
  0x0004:   pop    %rbp
  0x0005:   ret
```
so, the instruction `lea    (%rsi,%rdx,1)` will definitely be modified when patching verified_entry, and we can check the `eax` register to determine whether this instruction has been corrupted. However, a single check is not sufficient because no other instructions in this loop will modify the `eax` register. Therefore, after the first check, we change the return value and check it again.

test/hotspot/jtreg/compiler/jeandle/fileCheck/TestFileCheck.java
minchao1024: Done

src/hotspot/share/jeandle/jeandleCallVM.cpp
minchao1024: Fixed

taofengliu (approved):

PR #83, opened 2025-09-11
refactor: Refactor call vm for better readability

taofengliu:
- Add more comments for JeandleRuntimeRoutine.
- Rename some fields and functions related to call vm and refactor code to eliminate ambiguity between VM calls and routine calls
- Merge 'JeandleCallReloc' and 'JeandleCallVMReloc', add a call type in 'JeandleCompiledCall' for call vm.
- Add 'jeandle' to compilertype2name_tab, then we can see Jeandle compiled code are tagged with 'jeandle' in hs_err files.
- Split 'JeandleCompiledCall::call_site_size' to 'JeandleCompiledCall::call_site_size' and 'JeandleCompiledCall::call_site_patch_size'.

### DIFF
--- a/src/hotspot/share/jeandle/jeandleCompiledCode.hpp
+++ b/src/hotspot/share/jeandle/jeandleCompiledCode.hpp
@@ -50,10 +50,12 @@ class CallSiteInfo : public JeandleCompilationResourceObj {
                _target(target),
                _bci(bci),
                _statepoint_id(statepoint_id) {
+#ifdef ASSERT
     // We don't need to assign a unique statepoint id for each routine call site, only call type and target is used.
     bool use_default_statepoint_id = (statepoint_id == llvm::StatepointDirectives::DefaultStatepointID);
     bool is_routine_call = (type == JeandleCompiledCall::ROUTINE_CALL);
     assert(use_default_statepoint_id == is_routine_call, "routine calls should use the default statepoint id");
+#endif // ASSERT
   }
 
   JeandleCompiledCall::Type type() const { return _type; }


src/hotspot/share/jeandle/jeandleCompiledCode.cpp
```diff
@@ -294,14 +278,16 @@ void JeandleCompiledCode::resolve_reloc_info(JeandleAssembler& assembler) {
...
       auto& target = edge.getTarget();
 
-      if (!target.isDefined() && JeandleAssembler::is_call_vm_reloc_kind(edge.getKind())) {
-        // Call VM relocations.
+      if (!target.isDefined() && JeandleAssembler::is_routine_call_reloc_kind(edge.getKind())) {
+        // Routine call relocations.
         address target_addr = JeandleRuntimeRoutine::get_routine_entry(*target.getName());
 
-        int inst_end_offset = JeandleAssembler::fixup_call_inst_offset(static_cast<int>(block->getAddress().getValue() + edge.getOffset()));
+        int inst_end_offset = JeandleAssembler::fixup_native_call_inst_offset(static_cast<int>(block->getAddress().getValue() + edge.getOffset()));
```
feilongjiang: Should we name it `fixup_routine_call_inst_offset` to make it more consistent with the reloc kind?

taofengliu: Good idea!

src/hotspot/share/jeandle/jeandleCompiledCall.hpp
```diff
@@ -29,30 +29,37 @@
...
-    // Dynamic calls dispatch to the unverified entry point of a method and are
+    // To call a Java method, dynamic calls dispatch to the unverified entry point of a method and are
     // preceded by an instruction that places an inline cache holder in a register.
     DYNAMIC_CALL,
 
-    // For call vm stub, will call runtime routines
-    VM_CALL,
+    // Call a runtime routine. Different from other calls, routine calls are not generated as nops by LLVM,
+    // because we don't need extra instructions and any alignment for routine call sites. That means,
+    // call_site_patch_size(ROUTINE_CALL) is zero. So we have relocation information about routine calls in compiled object,
```
Cruise20, line 46: It seems there's some confusion in `call_site_patch_size`: the patch size of `ROUTINE_CALL` is not zero. The `call_site_size` is equivalent to `call_site_patch_size` for `ROUTINE_CALL`, maybe they should be different ?

src/hotspot/share/jeandle/jeandleCompiledCode.cpp
```diff
@@ -91,58 +91,86 @@ class JeandleConstReloc : public JeandleReloc {
   address _target;
 };
 
-class JeandleSafepointReloc : public JeandleReloc {
+class JeandleCallReloc : public JeandleReloc {
  public:
-  JeandleSafepointReloc(int inst_end_offset, ciEnv* env, ciMethod* method, OopMap* oop_map, CallSiteInfo* call) :
-    JeandleReloc(inst_end_offset - JeandleJavaCall::call_site_size(call->type())/* beginning of a call instruction */),
+  JeandleCallReloc(int inst_end_offset, ciEnv* env, ciMethod* method, OopMap* oop_map, CallSiteInfo* call) :
+    JeandleReloc(inst_end_offset - JeandleCompiledCall::call_site_size(call->type())/* beginning of a call instruction */),
```
Cruise20: Insert a `space` char before `/* beginning of a call instruction */` annotation.

src/hotspot/share/jeandle/jeandleCompiledCode.cpp
```diff
@@ -294,14 +278,16 @@ void JeandleCompiledCode::resolve_reloc_info(JeandleAssembler& assembler) {
...
         address target_addr = JeandleRuntimeRoutine::get_routine_entry(*target.getName());
 
-        int inst_end_offset = JeandleAssembler::fixup_call_inst_offset(static_cast<int>(block->getAddress().getValue() + edge.getOffset()));
+        int inst_end_offset = JeandleAssembler::fixup_routine_call_inst_offset(static_cast<int>(block->getAddress().getValue() + edge.getOffset()));
 
         // TODO: Set the right bci.
-        _vm_call_sites[inst_end_offset] = new CallSiteInfo(0/* statepoint_id */, JeandleJavaCall::STATIC_CALL, target_addr, -1/* bci */);
+        _routine_call_sites[inst_end_offset] = new CallSiteInfo(JeandleCompiledCall::ROUTINE_CALL,
+                                                                target_addr,
+                                                                -1/* bci */);
```
Cruise20, line 290: Insert a space before annotation

src/hotspot/share/jeandle/jeandleCallVM.cpp
```diff
@@ -70,14 +70,14 @@ void JeandleCallVM::generate_call_VM(const char* name, address c_func, llvm::Fun
   llvm::Value* c_func_ptr = ir_builder.CreateIntToPtr(c_func_addr, c_func_ptr_type);
   llvm::CallInst* call_c_func = ir_builder.CreateCall(func_type, c_func_ptr, args);
   call_c_func->setCallingConv(llvm::CallingConv::C);
-  JeandleJavaCall::Type call_type = JeandleJavaCall::Type::VM_CALL;
-  code.call_sites()[0] = new CallSiteInfo(0 /* statepoint id */, call_type, c_func, 0 /* bci */);
+  JeandleCompiledCall::Type call_type = JeandleCompiledCall::STUB_C_CALL;
+  code.push_non_routine_call_site(new CallSiteInfo(call_type, c_func, -1/* bci */, code.next_statepoint_id()));
   llvm::Attribute id_attr = llvm::Attribute::get(context,
                                                  llvm::jeandle::Attribute::StatepointID,
                                                  "0");
```
Cruise20: Update `id_attr`, to be aligned with `statepoint_id` in `CallSiteInfo`.

src/hotspot/share/jeandle/jeandleRuntimeRoutine.hpp
```diff
@@ -46,9 +46,17 @@
   def(exceptional_return)
```
kuaiwei, line 46: We also need define assembly routines with argument type and return type. For example, StubRoutines::_dsin has signature of `double dsin(double)`

taofengliu: Sure we need this, but I think maybe it's better to support it when we need it. If we support it now, we don't have use case and test case.
What do you think?

kuaiwei: I just find Jeandle use stub routines defined by JeandleRuntimeRoutine. How about routines defined by hotspot? Could we use them directly? Or do we need re-implement them in jeandle ? 
I prefer to reuse them. So we may need some class like `CallHotspotVM` or `CallHotspotRoutine`?

taofengliu: We can call them directly, just use a routine call like calling a JeandleRuntimeRoutine.

Issue #84, opened 2025-09-15
Support intrinsic implemented by native function or assembly routine


kuaiwei:
**The feature you'd like**
Jeandle now can handle intrinsic implemented as llvm intrinsic. We need support more intrinsic implemented by native functions or assemble routine provided by hotspot.

**Implementation you've considered**

**Additional context**

PR #88, opened 2025-09-16
feat: Support intrinsic implemented by native function or assembly routine

kuaiwei:
## Related issue(s):

issue #84 

Support intrinsic of Math.sin(double) backed by hotspot runtime and stub.

### DIFF
--- a/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
+++ b/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
@@ -34,6 +34,7 @@
 #include "utilities/debug.hpp"
 #include "ci/ciMethodBlocks.hpp"
 #include "runtime/sharedRuntime.hpp"
+#include "runtime/stubRoutines.hpp"
 #include "utilities/ostream.hpp"
 
 JeandleBasicBlock::JeandleBasicBlock(int block_id,
@@ -941,17 +942,19 @@ void JeandleAbstractInterpreter::invoke() {
   assert(declared_signature != nullptr, "cannot be null");
   assert(will_link == target->is_loaded(), "");
 
-  if (target->is_loaded() && target->check_intrinsic_candidate()) {
-    if (inline_intrinsic(target)) {
-      if (log_is_enabled(Debug, jeandle)) {
-        ResourceMark rm;
-        stringStream ss;
-        target->print_name(&ss);
-        log_debug(jeandle)("Method `%s` is parsed as intrinsic", ss.as_string());
-      }
-      return;
-    };
+  // try inline callee as intrinsic
+  if (target->is_loaded()
+    && target->check_intrinsic_candidate()
+    && inline_intrinsic(target)) {
+    if (log_is_enabled(Debug, jeandle)) {
+      ResourceMark rm;
+      stringStream ss;
+      target->print_name(&ss);
+      log_debug(jeandle)("Method `%s` is parsed as intrinsic", ss.as_string());
+    }
+    return;
   }
+
   const Bytecodes::Code bc = _codes.cur_bc();
 
   // Construct arguments.
@@ -1036,10 +1039,64 @@ bool JeandleAbstractInterpreter::inline_intrinsic(const ciMethod* target) {
       _jvm->fpush(_ir_builder.CreateIntrinsic(JeandleType::java2llvm(BasicType::T_FLOAT, *_context), llvm::Intrinsic::fabs, {_jvm->fpop()}));
       break;
     }
+    case vmIntrinsicID::_dsin: {
+      if (StubRoutines::dsin() != nullptr) {
+        _jvm->dpush(call_vm((address) StubRoutines::dsin(), "StubRoutines::dsin", T_DOUBLE, {T_DOUBLE}, true));
+      } else {
+        _jvm->dpush(call_vm((address) SharedRuntime::dsin, "SharedRuntime::dsin", T_DOUBLE, {T_DOUBLE}, true));
+      }
+      break;
+    }
     default:
       return false;
   }
   return true;
+}
+
+// Generate IR for call into runtime stub
+llvm::CallInst* JeandleAbstractInterpreter::call_vm(address c_func, llvm::StringRef name, BasicType return_type, llvm::ArrayRef<BasicType> arg_types, bool is_leaf) {
+  std::vector<llvm::Value*> args (arg_types.size());
+  std::vector<llvm::Type*>  types(arg_types.size());
+
+  int i=0;
+  for (BasicType type : arg_types) {
+    switch (type) {
+      case BasicType::T_BOOLEAN:
+      case BasicType::T_CHAR:
+      case BasicType::T_BYTE:
+      case BasicType::T_SHORT:
+      case BasicType::T_INT:
+      case BasicType::T_FLOAT:
+      case BasicType::T_DOUBLE:
+      case BasicType::T_LONG:
+      case BasicType::T_OBJECT:
+      case BasicType::T_ARRAY:
+        args [i] = _jvm->pop(type);
+        types[i] = JeandleType::java2llvm(type, *_context);
+        break;
+      case BasicType::T_VOID:
+      case BasicType::T_ADDRESS:
+      case BasicType::T_NARROWOOP:
+      case BasicType::T_METADATA:
+      case BasicType::T_NARROWKLASS:
+      case BasicType::T_CONFLICT:
+      case BasicType::T_ILLEGAL:
+      default:
+        ShouldNotReachHere();
+        return nullptr;
+    }
+    i++;
+  }
+  llvm::Type*         func_return_type = JeandleType::java2llvm(return_type, *_context);
+  llvm::FunctionType* func_type        = llvm::FunctionType::get(func_return_type, types, false);
+  llvm::PointerType*  func_ptr_type    = llvm::PointerType::get(func_type, llvm::jeandle::AddrSpace::CHeapAddrSpace);
+  llvm::Value*        entry_addr       = _ir_builder.getInt64((intptr_t)c_func);
+  llvm::Value*        func_ptr         = _ir_builder.CreateIntToPtr(entry_addr, func_ptr_type);
+  llvm::CallInst*     call             = _ir_builder.CreateCall(func_type, func_ptr, args);
+  call->setTailCall(is_leaf);
+  call->setCallingConv(llvm::CallingConv::C);
+  call->setName(name);
+  return call;
 }
 
 void JeandleAbstractInterpreter::stack_op(Bytecodes::Code code) {
--- a/src/hotspot/share/jeandle/jeandleAbstractInterpreter.hpp
+++ b/src/hotspot/share/jeandle/jeandleAbstractInterpreter.hpp
@@ -252,6 +252,7 @@ class JeandleAbstractInterpreter : public StackObj {
   void rem_op(BasicType type, Bytecodes::Code code);
 
   llvm::CallInst* call_java_op(llvm::StringRef java_op, llvm::ArrayRef<llvm::Value*> args);
+  llvm::CallInst* call_vm(address c_func, llvm::StringRef name, BasicType return_type, llvm::ArrayRef<BasicType> arg_types, bool is_leaf = false);
 
   void add_safepoint_poll();
 
--- a/test/hotspot/jtreg/compiler/jeandle/intrinsic/TestSinDouble.java
+++ b/test/hotspot/jtreg/compiler/jeandle/intrinsic/TestSinDouble.java
@@ -1,0 +1,129 @@
+/*
+ * Copyright (c) 2025, the Jeandle-JDK Authors. All Rights Reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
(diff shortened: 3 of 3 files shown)


taofengliu (commented):
Can we use `llvm.sin.*` ?

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1036,12 +1039,66 @@ bool JeandleAbstractInterpreter::inline_intrinsic(const ciMethod* target) {
...
+      break;
+    }
     default:
       return false;
   }
   return true;
 }
 
+// Generate IR for call into runtime stub
+llvm::CallInst* JeandleAbstractInterpreter::call_vm(address c_func, llvm::StringRef name, BasicType return_type, llvm::ArrayRef<BasicType> arg_types, bool is_leaf) {
```
feilongjiang: We can further simplify the call vm logic in `JeandleAbstractInterpreter::rem_op` by calling `call_vm` directly.

kuaiwei: Yes, it can use call_vm as well. It can be done as a new PR.

kuaiwei:
> Can we use `llvm.sin.*` ?

Yes, it also can be used. This PR show we can call hotspot runtime stub for intrinsic. We may add a new flag to prefer llvm intrinsic.

taofengliu:
> > Can we use `llvm.sin.*` ?
> 
> Yes, it also can be used. This PR show we can call hotspot runtime stub for intrinsic. We may add a new flag to prefer llvm intrinsic.

I think we should always use llvm intrinsics, because llvm always has newer hardware support, more backend support and llvm intrinsics may expose more opportunities to do optimization in backend.
Calling hotspot runtime may be used on some java-specific intrinsics.
What do you think of this?

kuaiwei:
> > > Can we use `llvm.sin.*` ?
> > 
> > 
> > Yes, it also can be used. This PR show we can call hotspot runtime stub for intrinsic. We may add a new flag to prefer llvm intrinsic.
> 
> I think we should always use llvm intrinsics, because llvm always has newer hardware support, more backend support and llvm intrinsics may expose more opportunities to do optimization in backend. Calling hotspot runtime may be used on some java-specific intrinsics. What do you think of this?

My opnion is we should use hotspot stubs. They have some benifits:
* They are verified by openjdk comunity and we needn't care implementation difference between hotspot and llvm.
* Hotspot keeps update with new hardware. Many cpu vendor contribute support for new instructions.
* The hotspot can detect running cpu and choose better instrcution for stubs. In my test, llvm intrinsic in jeandle doesn't optimize like it. I think we need ask llvm to optimize for native cpu arch.
* Stubs locate in managed code cache. It provides optunity to optimize code layout. Sometimes we find it is useful to change stub location for better performance.

taofengliu:
> My opnion is we should use hotspot stubs. They have some benifits:
> 
> * They are verified by openjdk comunity and we needn't care implementation difference between hotspot and llvm.
> * Hotspot keeps update with new hardware. Many cpu vendor contribute support for new instructions.
> * The hotspot can detect running cpu and choose better instrcution for stubs. In my test, llvm intrinsic in jeandle doesn't optimize like it. I think we need ask llvm to optimize for native cpu arch.
> * Stubs locate in managed code cache. It provides optunity to optimize code layout. Sometimes we find it is useful to change stub location for better performance.

I noticed that LLVM transforms sin and similar intrinsics into library calls, so in this scenario, using Hotspot's provided intrinsics would actually be more efficient.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
taofengliu: How about add Jeandle used hotspot routines to `JeandleRuntimeRoutine`, their addresses will be resolved by `JeandleCompiledCode`. Through this approach, we can uniformly manage all external routines of Jeandle, and in LLVM IR, each routine will be declared as a function, allowing us to determine which routine is being called by its function name.

taofengliu: We can merge this pr first, then do this in another pr, this depends on you.

kuaiwei: They are changed as calling `call_vm`

kuaiwei: Hotspot routines are wrapped as JeandleRuntimeRoutine. 

But I find I can not assign `gc-leaf-function` attribute to the call. It looks it is not a standard llvm attribute. We can resolve it in future PR.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1065,12 +1068,26 @@ bool JeandleAbstractInterpreter::inline_intrinsic(const ciMethod* target) {
...
       return false;
   }
   return true;
 }
 
+// Generate IR for call into runtime stub
+llvm::CallInst* JeandleAbstractInterpreter::call_vm(llvm::FunctionCallee callee, llvm::ArrayRef<llvm::Value *> args, bool is_leaf) {
+  llvm::CallInst *call = _ir_builder.CreateCall(callee, args);
+  call->setCallingConv(llvm::CallingConv::C);
+  call->setTailCall(is_leaf);
```
taofengliu: Tail call and leaf call maybe not one thing.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1249,32 +1266,16 @@ void JeandleAbstractInterpreter::arith_op(BasicType type, Bytecodes::Code code)
...
+    llvm::Value* l = _jvm->fpop();
+    _jvm->fpush(call_vm(JeandleRuntimeRoutine::hotspot_SharedRuntime_frem_callee(_module), {l, r}, /* is_leaf */ true));
   } else {
-    llvm::Type* double_type = llvm::Type::getDoubleTy(*_context);
-    return_type = double_type;
-    args_type = {double_type, double_type};
-    call_addr = CAST_FROM_FN_PTR(address, SharedRuntime::drem);
+    llvm::Value* r = _jvm->dpop();
+    llvm::Value* l = _jvm->dpop();
+    _jvm->dpush(call_vm(JeandleRuntimeRoutine::hotspot_SharedRuntime_drem_callee(_module), {l, r}, /* is_leaf */ true));
```
taofengliu: This function becomes simple now. We can merge this into function `arith_op`.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1065,12 +1068,26 @@ bool JeandleAbstractInterpreter::inline_intrinsic(const ciMethod* target) {
...
+      break;
+    }
     default:
       return false;
   }
   return true;
 }
 
+// Generate IR for call into runtime stub
+llvm::CallInst* JeandleAbstractInterpreter::call_vm(llvm::FunctionCallee callee, llvm::ArrayRef<llvm::Value *> args, bool is_leaf) {
```
```suggestion
// Generate IR for calling into JeandleRuntimeRoutine
llvm::CallInst* JeandleAbstractInterpreter::call_runtime_routine(llvm::FunctionCallee callee, llvm::ArrayRef<llvm::Value *> args, bool is_leaf) {
```
`call_vm` maybe confused with `JeandleCallVM`.

src/hotspot/share/jeandle/jeandleRuntimeRoutine.hpp
```diff
@@ -45,6 +45,18 @@
...
 
+//-----------------------------------------------------------------------------------------------------------------------------------
+//  name                     | func_entry            | return_type                        | arg_types
+//-----------------------------------------------------------------------------------------------------------------------------------
+#define ALL_HOTSPOT_STUBS(def)                                                                                                       \
+  def(SharedRuntime_dsin,    SharedRuntime::dsin,     llvm::Type::getDoubleTy(context),    llvm::Type::getDoubleTy(context))         \
+  def(StubRoutines_dsin,     StubRoutines::dsin(),    llvm::Type::getDoubleTy(context),    llvm::Type::getDoubleTy(context))         \
+  def(SharedRuntime_drem,    SharedRuntime::drem,     llvm::Type::getDoubleTy(context),    llvm::Type::getDoubleTy(context),         \
+                                                                                           llvm::Type::getDoubleTy(context))         \
+  def(SharedRuntime_frem,    SharedRuntime::frem,     llvm::Type::getFloatTy(context),     llvm::Type::getFloatTy(context),          \
```
taofengliu: I think we can define this macro like `ALL_JEANDLE_C_ROUTINES`, func names are also the func entries.

And introducing hotspot routines making the comments below no longer valid, we can update them, by adding the description of calling hotspot routines.

src/hotspot/share/jeandle/jeandleRuntimeRoutine.hpp
```diff
@@ -76,6 +88,17 @@ class JeandleRuntimeRoutine : public AllStatic {
     return _routine_entry.lookup(name);
   }
 
+#define DEF_HOTSPOT_STUB_CALLEE(name, func_entry, return_type, ...) \
+  static llvm::FunctionCallee hotspot_##name##_callee(llvm::Module& target_module) {                    \
+    llvm::LLVMContext& context = target_module.getContext();                                            \
+    llvm::FunctionType* func_type = llvm::FunctionType::get(return_type, {__VA_ARGS__}, false);         \
+    llvm::FunctionCallee callee = target_module.getOrInsertFunction("hotspot_stub_"  #name, func_type); \
```
taofengliu:
```suggestion
    llvm::FunctionCallee callee = target_module.getOrInsertFunction(#name, func_type); \
```
Minor suggestion. We can already distinguish hotspot routines by their name.

src/hotspot/share/jeandle/jeandleRuntimeRoutine.hpp
```diff
@@ -76,6 +88,17 @@ class JeandleRuntimeRoutine : public AllStatic {
     return _routine_entry.lookup(name);
   }
 
+#define DEF_HOTSPOT_STUB_CALLEE(name, func_entry, return_type, ...) \
```
taofengliu:
```suggestion
#define DEF_HOTSPOT_ROUTINE_CALLEE(name, func_entry, return_type, ...) \
```
Minor suggestion

src/hotspot/share/jeandle/jeandleRuntimeRoutine.hpp
```diff
@@ -45,6 +45,18 @@
 #define ALL_JEANDLE_ASSEMBLY_ROUTINES(def) \
   def(exceptional_return)
 
+//-----------------------------------------------------------------------------------------------------------------------------------
+//  name                     | func_entry            | return_type                        | arg_types
+//-----------------------------------------------------------------------------------------------------------------------------------
+#define ALL_HOTSPOT_STUBS(def)                                                                                                       \
```
taofengliu:
```suggestion
#define ALL_HOTSPOT_ROUTINES(def)                                                                                                       \
```
Like `ALL_JEANDLE_C_ROUTINES` and `ALL_JEANDLE_ASSEMBLY_ROUTINES`

src/hotspot/share/jeandle/jeandleRuntimeRoutine.cpp
```diff
@@ -55,6 +60,9 @@ bool JeandleRuntimeRoutine::generate(llvm::TargetMachine* target_machine, llvm::
   // Generate assembly routines.
   ALL_JEANDLE_ASSEMBLY_ROUTINES(GEN_ASSEMBLY_ROUTINE_BLOB);
 
+  // Register hotspot stub
+  ALL_HOTSPOT_STUBS(REGISTER_HOTSPOT_STUBS);
```
taofengliu:
```suggestion
  // Register hotspot routines
  ALL_HOTSPOT_ROUTINES(REGISTER_HOTSPOT_ROUTINES);
```
Minor suggestion

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
taofengliu: This attribute is used in [llvm::callsGCLeafFunction](https://github.com/jeandle/jeandle-llvm/blob/21dafc65f5cdaee6a3da9b8b25b69a677d6ec147/llvm/lib/Transforms/Utils/Local.cpp#L3607).
We can add it by using `Function::addFnAttr` to each declared routine function. And it's better to add it to [Attributes.h](https://github.com/jeandle/jeandle-llvm/blob/21dafc65f5cdaee6a3da9b8b25b69a677d6ec147/llvm/include/llvm/IR/Jeandle/Attributes.h), too.

kuaiwei: I tried with `Attritute::isExistingAttribute(“gc-leaf-function");` , and it reported false.
I think this is out of scope of this feature. We may refine it in future.

taofengliu: LLVM's attributes are designed to be very flexible - we just need to add this string, and it can be processed by other passes. If it cannot be recognized, there must be some other issues.
But we can surely make it in another pr.

src/hotspot/share/jeandle/jeandleRuntimeRoutine.hpp
kuaiwei: The func name has different pattern, for example, `SharedRuntime::dsin` and `StubRoutins::dsin()` , so I added a param for name.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
kuaiwei: move them to arith_op

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
kuaiwei: I'm not clear about the difference. Now I remove the `TailCall` flag, `is_leaf` is reserved for future.

taofengliu: OK, I will try to express something like leaf call in LLVM IR in the future.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
taofengliu: A new issue is created to trace this, #92

Issue #18, opened 2025-07-24
Report source code line number and stack status when Jeandle is crashed


taofengliu:
**The feature you'd like**
Report source code line number and stack status when Jeandle is crashed。

**Implementation you've considered**
In hs_err file.

**Additional context**

PR #96, opened 2025-09-22
feat: Report source code line number and stack status when Jeandle is crashed

minchao1024:
## Related issue(s):

issue #18

1、collect registers info and stack info when calling `report_vm_error`.
2、when the compiler thread crashed, skip `dump_compile_data` to avoid another crash. 
3、add a test case `TestHsErrFile.java` to verify whether the hs_err_pid.log is generated correctly.


minchao1024:
Before JDK 11, the hs-err file lacked registers information because the VM state was recorded by calling `report_vm_error` function, which itself may destroy registers information (unrecoverable). 
Since [JDK-8191101](https://bugs.openjdk.org/browse/JDK-8191101), VM can proactively trigger a SIGSEGV signal to have the OS record the user context containing accurate register information, and VM save it when handling signal in the `JVM_handle_linux_signal`. 
https://github.com/jeandle/jeandle-jdk/blob/19bec1d4b6e4fcad5f0800396fbdbd2a5a7e6584/src/hotspot/share/utilities/debug.cpp#L748-L770
The subsequent `report_vm_error` function directly uses the previously user context to generate the hs_err file, so, it would like:
```
*((address)unwritable_page) = a
report_vm_error(__FILE__, __LINE__, "crash here");
```
which is the macro expansion of `assert`
https://github.com/jeandle/jeandle-jdk/blob/19bec1d4b6e4fcad5f0800396fbdbd2a5a7e6584/src/hotspot/share/utilities/debug.hpp#L30
https://github.com/jeandle/jeandle-jdk/blob/19bec1d4b6e4fcad5f0800396fbdbd2a5a7e6584/src/hotspot/share/utilities/debug.hpp#L152-L166
https://github.com/jeandle/jeandle-jdk/blob/19bec1d4b6e4fcad5f0800396fbdbd2a5a7e6584/src/hotspot/share/utilities/debug.hpp#L51

minchao1024:
Temporarily skip `ciEnv::dump_compile_data` because it will cast `ciEnv::compiler_data()` to `Compile*`, and the type of compiler_data in JeandleCompiler is `JeandleCompilation`, which is not a subclass of `Compile`.
https://github.com/jeandle/jeandle-jdk/blob/c60a4a9190a9227ba691117f9fed83b19558defc/src/hotspot/share/ci/ciEnv.cpp#L1619-L1649

test/hotspot/jtreg/compiler/jeandle/vmCrash/TestHsErrFile.java
```diff
@@ -0,0 +1,217 @@
+/*
```
Cruise20, line 1: The test case can't cover this issue. It made jvm crash on jeandle compiled code or native code. While the source code line number and stack status are recorded for crash on `jeandle compilation`.
It may be a good idea to construct some `bad bytecodes` to trigger crash on `jeandle compilation`.

src/hotspot/share/jeandle/jeandleCompilation.cpp
```diff
@@ -80,8 +80,10 @@ JeandleCompilation::JeandleCompilation(llvm::TargetMachine* target_machine,
 
   if (error_occurred()) {
 #ifdef ASSERT
-    if (JeandleCrashOnError)
+    if (JeandleCrashOnError) {
+      TOUCH_ASSERT_POISON
```
taofengliu: Can we use `fatal` here?

src/hotspot/share/jeandle/jeandleCompilation.cpp
```diff
@@ -118,8 +120,10 @@ JeandleCompilation::JeandleCompilation(llvm::TargetMachine* target_machine,
 #ifdef ASSERT
   // Verify.
   if (llvm::verifyModule(*_llvm_module, &llvm::errs())) {
-    if (JeandleCrashOnError)
+    if (JeandleCrashOnError) {
+      TOUCH_ASSERT_POISON
```
taofengliu: I think we can call `report_error` here.

src/hotspot/share/ci/ciEnv.cpp
```diff
@@ -1617,6 +1617,10 @@ void ciEnv::find_dynamic_call_sites() {
 }
 
 void ciEnv::dump_compile_data(outputStream* out) {
+  if (UseJeandleCompiler) {
+    // TODO: handle dump_compile_data
+    return;
+  }
```
taofengliu: This may prevent c1 and c2 from performing dump operations.

test/hotspot/jtreg/compiler/jeandle/vmCrash/TestHsErrFile.java
minchao1024: done, add a testcase expect crash when jeandleCompilation, which will failed if we don't apply this pr.

test/hotspot/jtreg/compiler/jeandle/vmCrash/TestHsErrFile.java
```diff
@@ -0,0 +1,256 @@
...
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ */
+
+/**
+ * @test
+ * @library /test/lib
```
taofengliu, line 23: Add a require for linux

test/hotspot/jtreg/compiler/jeandle/vmCrash/TestHsErrFile.java
```diff
@@ -0,0 +1,256 @@
...
+            """
+            C  Java_TestHsErrFile_crashInNative
+            j  TestHsErrFile.crashInNative()V
+            J  jeandle TestHsErrFile.nativeMethodWrapper()V (4 bytes)
+            j  TestHsErrFile.lambda$makeProcess
+            j  TestHsErrFile$$Lambda
+            j  java.lang.Thread.runWith(Ljava/lang/Object;Ljava/lang/Runnable;)V
+            j  java.lang.Thread.run()V
+            """;
+        checkHsErrFile("hs_err_pid" + pid + ".log", anchor, expected);
```
taofengliu: We can just don't check the content of hs error file for native call.

test/hotspot/jtreg/compiler/jeandle/vmCrash/TestHsErrFile.java
```diff
@@ -0,0 +1,256 @@
...
+    public static long crashWhenJeandleCompilation() throws Exception {
+        List<String> cmdLine = new ArrayList<>();
+        cmdLine.add("-Xcomp");
+        cmdLine.add("-XX:-TieredCompilation");
+        cmdLine.add("-Xbatch");
+        cmdLine.add("-noverify");
+        cmdLine.add("-XX:-Inline");
+        cmdLine.add("-XX:CompileCommand=compileonly,Crash::doCrash");
+        cmdLine.add("-XX:+UnlockDiagnosticVMOptions");
+        cmdLine.add("-XX:+AbortVMOnCompilationFailure");
```
taofengliu: Currently, we don't need this.

test/hotspot/jtreg/compiler/jeandle/vmCrash/Crash.class
taofengliu, line 1: Let's try to use jasm to write this file

src/hotspot/share/jeandle/jeandleCompilation.cpp
minchao1024: Fixed

src/hotspot/share/jeandle/jeandleCompilation.cpp
minchao1024: here is `JeandleCompilation::JeandleCompilation` that is the last place wherewe can trigger crash when JeandleCrashOnError is on. I think use `fatal` to crash here is fine.

src/hotspot/share/ci/ciEnv.cpp
minchao1024: Done

test/hotspot/jtreg/compiler/jeandle/vmCrash/TestHsErrFile.java
minchao1024: Fixed

test/hotspot/jtreg/compiler/jeandle/vmCrash/TestHsErrFile.java
minchao1024: Fixed

test/hotspot/jtreg/compiler/jeandle/vmCrash/TestHsErrFile.java
minchao1024: Fixed

test/hotspot/jtreg/compiler/jeandle/vmCrash/Crash.class
minchao1024: Fixed

Issue #52, opened 2025-08-28
Refactor include headers


feilongjiang:
**Goals**
Refactor the current jeandle include headers to resolve macro conflicts.
**Motivation**
We witnessed some macro conflicts between LLVM and HotSpot, e.g., `assert` and `AARCH64`. Currently, we are simply adding extra includes or pragmas to bypass the conflict. 
**Description**
Provide a unified way to fix the macro conflicts. For example, we can add macros like `LLVM_HEADER_BEGIN` and `HOTSPOT_HEADER_BEGIN` to separate the include headers from LLVM and HotSpot.

PR #101, opened 2025-09-24
refactor: refactor header files to resolve macro conflicts between LLVM and Hotspot source code.

taofengliu:
## Related issue(s):

issue #52

Add two new headers `__llvmHeadersBegin__.hpp` and `__hotspotHeadersBegin__.hpp` to resolve macro conflicts.

### DIFF
--- a/src/hotspot/share/jeandle/__hotspotHeadersBegin__.hpp
+++ b/src/hotspot/share/jeandle/__hotspotHeadersBegin__.hpp
@@ -26,6 +26,6 @@
 
 // Resolve conflict with AARCH64 macro in LLVM headers.
 #ifdef SAVED_HOTSPOT_AARCH64
-  #define AARCH64 SAVED_HOTSPOT_AARCH64
+  #define AARCH64
   #undef SAVED_HOTSPOT_AARCH64
 #endif // SAVED_HOTSPOT_AARCH64
--- a/src/hotspot/share/jeandle/__hotspotHeadersBegin__.hpp
+++ b/src/hotspot/share/jeandle/__hotspotHeadersBegin__.hpp
@@ -18,13 +18,17 @@
  *
  */
 
-// Need to undef assert from stdlib. Then redefine our assert.
+// This file is used to resolve macro conflicts between LLVM and HotSpot.
+// For more information on macro conflicts, see __llvmHeadersBegin__.hpp
+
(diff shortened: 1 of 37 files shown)


src/hotspot/share/jeandle/__hotspotHeadersBegin__.hpp
```diff
@@ -0,0 +1,31 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ */
+
```
feilongjiang, line 20: We may need some documentation to guide the developer on the usage of `__hotspotHeadersBegin__.hpp` and `__llvmHeadersBegin__.hpp`. If there are any other macro conflicts, we can add them to including guards too.

src/hotspot/share/jeandle/__llvmHeadersBegin__.hpp
```diff
@@ -0,0 +1,25 @@
...
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ */
+
+#include <cassert>
+#ifdef AARCH64
+  #define SAVED_HOTSPOT_AARCH64
+  #undef AARCH64
```
feilongjiang, line 46: Is there any chance that some hotspot source files include jeandle header files without redefining `AArch64`, i.e., only include `__llvmHeadersBegin__.hpp` but never include `__hotspotHeadersBegin__.hpp`?

taofengliu: I believe there are some files that don't need to resolve macro conflicts, but it's so difficult to find out them.

src/hotspot/share/jeandle/__hotspotHeadersBegin__.hpp
Cruise20: Good idea. I've created a good first issue for it: #102

Issue #12, opened 2025-07-24
Support array related stores and loads.


taofengliu:
**The feature you'd like**
Implement the bytecodes "baload", "caload", "saload", "iaload", "laload", "faload", "daload" and "aaload".

**Implementation you've considered**
Refer to C2's implementation.

**Additional context**

lgxbslgx:
Is anyone working on this issue? I would like to investigate it.

taofengliu:
> Is anyone working on this issue? I would like to investigate it.

You can try! And if you encounter any problems, please feel free to contact us

lgxbslgx:
> You can try! And if you encounter any problems, please feel free to contact us

OK, but I don't have too much time recently. I may start working on October 1st.

PR #119, opened 2025-10-04
feat: Support array related stores and loads

lgxbslgx:
## Related issue(s):

issue #12

This patch implements the bytecodes `*aload` and `*astore`.

The jeandle related test cases passed locally (x86_64&linux).


lgxbslgx:
> * `JeandleAbstractInterpreter::do_array_load_inner` and `JeandleAbstractInterpreter::do_array_store_inner` have some common logic that can be merged.

I extracted the common part to a method `compute_array_element_address`.

> * All stores and loads must be atomic with `Unordered` consistency. Reference to #82 and https://llvm.org/docs/Atomics.html#unordered

I added the `Unordered` consistency.

> * To get better readability, we can define constant values like `arrayOopDesc::base_offset_in_bytes` as global values in the template module. There are already some examples in the template module like `arrayOopDesc.length_offset_in_bytes`.

The global value seems be only used by LLVM IR in `template.ll` which this patch doesn't need. Did I miss anything?

> * We can use the first argument of GEP to indicate element type, avoid computing the element offset in bytes. Just like `getelementptr inbounds i64, element_begin, element_index`

Fixed.

taofengliu:
> > * To get better readability, we can define constant values like `arrayOopDesc::base_offset_in_bytes` as global values in the template module. There are already some examples in the template module like `arrayOopDesc.length_offset_in_bytes`.
> 
> The global value seems be only used by LLVM IR in `template.ll` which this patch doesn't need. Did I miss anything?

Suppose we obtain an address from an array oop through arrayOopDesc::base_offset_in_bytes. In the LLVM IR, only the numeric value is visible, not the variable name. Therefore, we cannot infer from the IR alone what this address from the array oop actually represents.

For example, now we can get LLVM IR like this:
```
%ptr2 = getelementptr inbounds i8, ptr addrspace(1) %ptr1, i32 16
```
It's less readable than:
```
%ptr2 = getelementptr inbounds i8, ptr addrspace(1) %ptr1, i32 %arrayOopDesc.base_offset_in_bytes
```

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1439,3 +1439,119 @@ void JeandleAbstractInterpreter::arraylength() {
     llvm::CallInst* call = call_java_op("jeandle.arraylength", {array_oop});
     _jvm->ipush(call);
 }
+
+llvm::Value* JeandleAbstractInterpreter::compute_array_element_address(BasicType basic_type, llvm::Type* type) {
+  llvm::Value* index = _jvm->ipop();
+  llvm::Value* array_oop = _jvm->apop();
+  llvm::Value* array_base_offset = _ir_builder.getInt32(arrayOopDesc::base_offset_in_bytes(basic_type));
+  llvm::Value* array_base = _ir_builder.CreatePtrAdd(array_oop, array_base_offset);
```
```suggestion
  llvm::Value* array_base = _ir_builder.CreateInBoundsPtrAdd(array_oop, array_base_offset, "array_element_base");
```
Use `CreateInBoundsPtrAdd`, and give it a readable name.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1439,3 +1439,119 @@ void JeandleAbstractInterpreter::arraylength() {
...
+  llvm::Value* element_address = _ir_builder.CreateInBoundsGEP(type, array_base, index);
```
```suggestion
  llvm::Value* element_address = _ir_builder.CreateInBoundsGEP(type, array_base, index, "array_element_address");
```

lgxbslgx: Fixed.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
lgxbslgx: Fixed.

lgxbslgx:
> > > * To get better readability, we can define constant values like `arrayOopDesc::base_offset_in_bytes` as global values in the template module. There are already some examples in the template module like `arrayOopDesc.length_offset_in_bytes`.
> > 
> > 
> > The global value seems be only used by LLVM IR in `template.ll` which this patch doesn't need. Did I miss anything?
> 
> Suppose we obtain an address from an array oop through arrayOopDesc::base_offset_in_bytes. In the LLVM IR, only the numeric value is visible, not the variable name. Therefore, we cannot infer from the IR alone what this address from the array oop actually represents.
> 
> For example, now we can get LLVM IR like this:
> 
> ```
> %ptr2 = getelementptr inbounds i8, ptr addrspace(1) %ptr1, i32 16
> ```
> 
> It's less readable than:
> 
> ```
> %ptr2 = getelementptr inbounds i8, ptr addrspace(1) %ptr1, i32 %arrayOopDesc.base_offset_in_bytes
> ```

I noticed that the method `arrayOopDesc::base_offset_in_bytes` has one parameter., so it is not a global constant. I used the method `setName` explicitly in the newest patch.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1444,8 +1444,9 @@ llvm::Value* JeandleAbstractInterpreter::compute_array_element_address(BasicType
   llvm::Value* index = _jvm->ipop();
   llvm::Value* array_oop = _jvm->apop();
   llvm::Value* array_base_offset = _ir_builder.getInt32(arrayOopDesc::base_offset_in_bytes(basic_type));
-  llvm::Value* array_base = _ir_builder.CreatePtrAdd(array_oop, array_base_offset);
-  llvm::Value* element_address = _ir_builder.CreateInBoundsGEP(type, array_base, index);
+  array_base_offset->setName("arrayOopDesc.base_offset_in_bytes"); // May not take effect sometimes.
```
taofengliu: This is a constant, all other constants that have the same value will reuse the same `llvm::Value`, but in other situations, this name might confuse us.
I noticed that `arrayOopDesc::base_offset_in_bytes`'s value is obtained through calculation. It looks like currently we can only use numbers directly in LLVM IR as you did before.

lgxbslgx: I removed the statement just now.

```C++
  array_base_offset->setName("arrayOopDesc.base_offset_in_bytes"); // May not take effect sometimes.
```

taofengliu (approved):

Issue #109, opened 2025-09-28
CI support for RISC-V


taofengliu:
**The feature you'd like**
Support running Jeandle PR's CI on RISC-V environment.

**Implementation you've considered**
Support for both Jeandle-JDK and Jeandle-LLVM.

**Additional context**
Maybe we can use this GitHub app to get RISC-V environment: https://github.com/apps/riscv-builders

lgxbslgx:
I am investigating this issue. Please assign it to me.

PR #143, opened 2025-10-14
ci: Support for RISC-V

lgxbslgx:
## Related issue(s):

issue #109

This patch adds the CI support for RISC-V.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -39,5 +39,8 @@ queue_rules:
     autoqueue: true
     update_method: rebase
     name: jeandle-jdk
+    draft_bot_account: jeandle-bot
+    merge_bot_account: jeandle-bot
+    update_bot_account: jeandle-bot
 merge_queue:
   max_parallel_checks: 1
--- a/src/hotspot/share/jeandle/jeandleCompilation.cpp
+++ b/src/hotspot/share/jeandle/jeandleCompilation.cpp
@@ -71,6 +71,11 @@ JeandleCompilation::JeandleCompilation(llvm::TargetMachine* target_machine,
                                        _context(std::make_unique<llvm::LLVMContext>()),
                                        _code(env, method),
                                        _error_msg(nullptr) {
+  if (entry_bci != InvocationEntryBci) {
+    env->record_method_not_compilable("OSR not supported");
+    return;
+  }
+
   // Setup compilation.
   initialize();
   setup_llvm_module(template_buffer);
--- a/test/hotspot/jtreg/compiler/jeandle/TestIssue173.java
+++ b/test/hotspot/jtreg/compiler/jeandle/TestIssue173.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright (c) 2025, the Jeandle-JDK Authors. All Rights Reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ */
+
+/**
+ * @test
+ * @summary https://github.com/jeandle/jeandle-jdk/issues/173
+ * @library /test/lib /
+ * @build jdk.test.lib.Asserts
+ * @run main/othervm -Xbatch -Xcomp -XX:-TieredCompilation -XX:+UseJeandleCompiler -XX:CompileCommand=compileonly,TestIssue173::addCounter TestIssue173
+ */
+
+import jdk.test.lib.Asserts;
+
+public class TestIssue173 {
+    private static int counter = 0;
+    private static final Object lock = new Object();
+
+    public static void main(String[] args) throws Exception {
+        testSynchronized();
+    }
+
+    static void addOne() {
+        synchronized (lock) {
+            counter++;
+        }
+    }
+
+    static void addCounter() {
+        for (int j = 0; j < 100000; j++) {
+            addOne();
+        }
+    }
+
+    static void testSynchronized() throws Exception {
+        int threadCount = 3;
+        Thread[] threads = new Thread[threadCount];
+
+        for (int i = 0; i < threadCount; i++) {
+            threads[i] = new Thread(() -> {
+                addCounter();
+            });
+        }
+
+        for (Thread thread : threads) {
+            thread.start();
+        }
+
+        for (Thread thread : threads) {
+            thread.join();
+        }
+
+        Asserts.assertEquals(counter, 300000);
+    }
+}


lgxbslgx:
Should I add the RISC-V related (empty or unimplemented) classes and methods in this patch?

lgxbslgx:
The test workflow of this patch is at https://github.com/lgxbslgx/jeandle-jdk/actions/runs/18896514180 . All the test passed just now.

.github/workflows/main.yml
```diff
@@ -89,6 +90,34 @@ jobs:
...
+    uses: ./.github/workflows/build-cross-compile.yml
+    with:
+      gcc-major-version: '10'
+      target-cpu: aarch64
+      gnu-arch: aarch64
+      debian-arch: arm64
+      debian-repository: http://ports.ubuntu.com/ubuntu-ports/
+      debian-version: jammy
+      llvm-target: AArch64
+    if: needs.prepare.outputs.linux-aarch64 == 'true'
```
taofengliu, line 105: Do we need cross-compile here when there is already a non-cross-compile job for AArch64?

.github/workflows/build-cross-compile.yml
```diff
@@ -150,43 +166,100 @@ jobs:
...
       - name: 'Configure'
         run: >
           bash configure
-          --with-conf-name=linux-${{ matrix.target-cpu }}
+          --with-conf-name=linux-${{ inputs.target-cpu }}
           --with-version-opt=${GITHUB_ACTOR}-${GITHUB_SHA}
           --with-boot-jdk=${{ steps.bootjdk.outputs.path }}
           --with-gtest=${{ steps.gtest.outputs.path }}
           --with-zlib=system
+          --disable-warnings-as-errors
```
taofengliu, line 245: Why we need this? What will happen if we enable it?

.github/workflows/build-cross-compile.yml
```diff
@@ -111,37 +92,72 @@ jobs:
...
-              g++-${{ inputs.gcc-major-version }}-${{ matrix.gnu-arch }}-linux-gnu${{ matrix.gnu-abi}} \
+              gcc-${{ inputs.gcc-major-version }}-${{ inputs.gnu-arch }}-linux-gnu${{ inputs.gnu-abi }} \
+              g++-${{ inputs.gcc-major-version }}-${{ inputs.gnu-arch }}-linux-gnu${{ inputs.gnu-abi }} \
               libxrandr-dev libxtst-dev libcups2-dev libasound2-dev
           sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-${{ inputs.gcc-major-version }} 100 --slave /usr/bin/g++ g++ /usr/bin/g++-${{ inputs.gcc-major-version }}
+          if [[ "${{ inputs.gnu-arch }}" == "aarch64" ]]; then
+            sudo ln -s /usr/aarch64-linux-gnu${{ inputs.gnu-abi }}/lib/ld-linux-aarch64.so.1  /lib/ld-linux-aarch64.so.1
+          elif [[ "${{ inputs.gnu-arch }}" == "riscv64" ]]; then
+            sudo ln -s /usr/riscv64-linux-gnu${{ inputs.gnu-abi }}/lib/ld-linux-riscv64-lp64d.so.1  /lib/ld-linux-riscv64-lp64d.so.1
+          fi
```
taofengliu, line 103: The origin steps did not contain this, please explain why we add this.

.github/workflows/build-cross-compile.yml
```diff
@@ -150,43 +166,100 @@ jobs:
...
           --with-jmod-compress=zip-1
-          CC=${{ matrix.gnu-arch }}-linux-gnu${{ matrix.gnu-abi}}-gcc-${{ inputs.gcc-major-version }}
-          CXX=${{ matrix.gnu-arch }}-linux-gnu${{ matrix.gnu-abi}}-g++-${{ inputs.gcc-major-version }}
+          CC=${{ inputs.gnu-arch }}-linux-gnu${{ inputs.gnu-abi }}-gcc-${{ inputs.gcc-major-version }}
+          CXX=${{ inputs.gnu-arch }}-linux-gnu${{ inputs.gnu-abi }}-g++-${{ inputs.gcc-major-version }}
           ${{ inputs.extra-conf-options }} ${{ inputs.configure-arguments }} || (
           echo "Dumping config.log:" &&
           cat config.log &&
           exit 1)
-        if: steps.create-sysroot.outcome == 'success' || steps.get-cached-sysroot.outputs.cache-hit == 'true'
```
taofengliu, line 184: Is it OK to delete this dependency?

.github/workflows/build-cross-compile.yml
```diff
@@ -150,43 +166,100 @@ jobs:
...
 
       - name: 'Build'
         id: build
         uses: ./.github/actions/do-build
         with:
           make-target: 'hotspot ${{ inputs.make-arguments }}'
-          platform: linux-${{ matrix.target-cpu }}
-        if: steps.create-sysroot.outcome == 'success' || steps.get-cached-sysroot.outputs.cache-hit == 'true'
```
taofengliu, line 192: Is it OK to delete this dependency?

.github/workflows/build-cross-compile.yml
```diff
@@ -111,37 +92,72 @@ jobs:
...
+          elif [[ "${{ inputs.gnu-arch }}" == "riscv64" ]]; then
+            sudo ln -s /usr/riscv64-linux-gnu${{ inputs.gnu-abi }}/lib/ld-linux-riscv64-lp64d.so.1  /lib/ld-linux-riscv64-lp64d.so.1
+          fi
 
-      - name: 'Check cache for sysroot'
-        id: get-cached-sysroot
-        uses: actions/cache@v4
-        with:
-          path: sysroot
-          key: sysroot-${{ matrix.debian-arch }}-${{ hashFiles('./.github/workflows/build-cross-compile.yml') }}
```
taofengliu, line 124: Why can't we use this cache?

.github/workflows/build-cross-compile.yml
```diff
@@ -150,43 +166,100 @@ jobs:
           sudo chown ${USER} -R sysroot
           rm -rf sysroot/{dev,proc,run,sys,var}
           rm -rf sysroot/usr/{sbin,bin,share}
-          rm -rf sysroot/usr/lib/{apt,gcc,udev,systemd}
+          rm -rf sysroot/usr/lib/{apt,udev,systemd}
```
taofengliu, line 169: Why delete gcc from this list, since we don't use sysroot cache.

.github/workflows/main.yml
lgxbslgx: It is good to verify whether the `AArch64` cross-compilation is right. Because many developers use x86_64 computers and change the code of other architectures (and then cross-compile). We need to ensure the development work flow is correct. The issue https://github.com/jeandle/jeandle-jdk/issues/165 is an example.

.github/workflows/build-cross-compile.yml
lgxbslgx: Normal JDK building will treat the warnings as errors, but sometimes the code may have compilation warnings. It depends on the OS, compiler and the JDK code. For example:

```
Compiling vm_version_riscv.cpp (for libjvm.so)
/home/runner/work/jeandle-jdk/jeandle-jdk/src/hotspot/os_cpu/linux_riscv/vm_version_linux_riscv.cpp: In static member function 'static uint32_t VM_Version::cpu_vector_length()':
/home/runner/work/jeandle-jdk/jeandle-jdk/src/hotspot/os_cpu/linux_riscv/vm_version_linux_riscv.cpp:77:32: warning: ISO C++17 does not allow 'register' storage class specifier [-Wregister]
   77 |         register unsigned long __v;                             \
      |                                ^~~
/home/runner/work/jeandle-jdk/jeandle-jdk/src/hotspot/os_cpu/linux_riscv/vm_version_linux_riscv.cpp:87:20: note: in expansion of macro 'read_csr'
   87 |   return (uint32_t)read_csr(CSR_VLENB);
      |                    ^~~~~~~~
At global scope:
cc1plus: note: unrecognized command-line option '-Wno-template-id-cdtor' may have been intended to silence earlier diagnostics
```

> What will happen if we enable it?

The building will fail. The warnings, just as I show above, will be treated as errors, and then the building will fail.

.github/workflows/build-cross-compile.yml
lgxbslgx: When I cross-compiled the code, it reported `can't find /lib/ld-linux-riscv64-lp64d.so.1`. So I create such symbolic link to avoid the problem.

.github/workflows/build-cross-compile.yml
lgxbslgx: When a step fails, the subsequent steps do not run. So if we don't need a branch(success and fail), we don't need the `if` expression.

.github/workflows/build-cross-compile.yml
lgxbslgx: Generally, we don't use such command locally after building the first time. I would like to ensure the complete building steps are right and the newbie can always build the `jeandle-jdk` successfully.

.github/workflows/build-cross-compile.yml
lgxbslgx: I add this directory to the linking flags to avoid the linking error sometimes. If we decide not to cache the `sysroot`, we don't need to remove any directories actually.

```
-DCMAKE_EXE_LINKER_FLAGS="-L$GITHUB_WORKSPACE/sysroot/usr/lib/${{ inputs.gnu-arch }}-linux-gnu${{ inputs.gnu-abi }} -L$GITHUB_WORKSPACE/sysroot/usr/lib/gcc/${{ inputs.gnu-arch }}-linux-gnu${{ inputs.gnu-abi }}/${{ inputs.gcc-major-version }} -latomic -lpthread"
```

taofengliu:
@Mergify queue

taofengliu:
@mergifyio queue

taofengliu:
@Mergifyio queue

taofengliu:
@Mergifyio queue

taofengliu:
@Mergifyio queue

taofengliu:
@Mergifyio queue

taofengliu:
@Mergifyio queue

taofengliu:
@Mergifyio queue

taofengliu:
@Mergifyio queue

taofengliu:
@Mergifyio queue

taofengliu:
@Mergifyio queue

PR #150, opened 2025-10-20
refactor: JeandleAbstractInterpreter::call_runtime_routine

taofengliu:
- Not every routine can be called through C calling convention
- `is_leaf` is not used
- Add some description for hotspot routine


src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1192,9 +1192,9 @@ bool JeandleAbstractInterpreter::inline_intrinsic(const ciMethod* target) {
 }
 
 // Generate IR for calling into JeandleRuntimeRoutine
-llvm::CallInst* JeandleAbstractInterpreter::call_runtime_routine(llvm::FunctionCallee callee, llvm::ArrayRef<llvm::Value *> args, bool is_leaf) {
+llvm::CallInst* JeandleAbstractInterpreter::call_jeandle_routine(llvm::FunctionCallee callee, llvm::ArrayRef<llvm::Value *> args, llvm::CallingConv::ID calling_conv) {
```
Cruise20, line 1195: Apply this change to `install_exceptional_return`

Issue #167, opened 2025-10-27
Support aconst_null bytecode


Cruise20:
**The feature you'd like**
Support aconst_null bytecode to pass the following case.

```
public class AconstNullDemo {
    public static void main(String[] args) {
        System.out.println("Starting test.");
        test();
    }

    private static void test() {
        String str = null;
        System.out.println(str);
    }
}
```

Current crash log: 
```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  Internal Error (/home/hexuejin/repo/jeandle/jeandle-jdk/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp:606), pid=245328, tid=245342
#  Error: Unimplemented()
#
# JRE version: OpenJDK Runtime Environment (21.0.7) (fastdebug build 21.0.7-internal-adhoc.root.jeandle-jdk)
# Java VM: OpenJDK 64-Bit Server VM (fastdebug 21.0.7-internal-adhoc.root.jeandle-jdk, compiled mode, g1 gc, linux-amd64)
# Problematic frame:
# V  [libjvm.so+0xf36f8f]  JeandleAbstractInterpreter::interpret_block(JeandleBasicBlock*)+0x203f
#
# Core dump will be written. Default location: /home/admin/logs/core.245328
#
# An error report file with more information is saved as:
# /home/hexuejin/work/test_aconst_null/hs_err_pid245328.log
#
# Compiler replay data is saved as:
# /home/hexuejin/work/test_aconst_null/replay_pid245328.log
#
# If you would like to submit a bug report, please visit:
#   https://bugreport.java.com/bugreport/crash.jsp
```

PR #187, opened 2025-10-31
feat: Support aconst_null bytecode

Anjian-Wen:
## Related issue(s):
issue #167 

Support aconst_null bytecode to pass the following case.

```
public class AconstNullDemo {
    public static void main(String[] args) {
        System.out.println("Starting test.");
        test();
    }

    private static void test() {
        String str = null;
        System.out.println(str);
    }
}
```

### DIFF
--- a/.github/workflows/build-cross-compile.yml
+++ b/.github/workflows/build-cross-compile.yml
@@ -1,5 +1,6 @@
 #
 # Copyright (c) 2022, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2025, the Jeandle-JDK Authors. All Rights Reserved.
 # DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
 #
 # This code is free software; you can redistribute it and/or modify it
@@ -31,6 +32,24 @@ on:
       gcc-major-version:
         required: true
         type: string
+      target-cpu:
+        required: true
+        type: string
+      gnu-arch:
+        required: true
+        type: string
+      debian-arch:
+        required: true
+        type: string
+      debian-repository:
+        required: true
+        type: string
+      debian-version:
+        required: true
+        type: string
+      llvm-target:
+        required: true
+        type: string
       extra-conf-options:
         required: false
         type: string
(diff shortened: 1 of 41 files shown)


test/hotspot/jtreg/compiler/jeandle/bytecodeTranslate/field/TestNullAccess.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    public static void main(String[] args) throws Exception {
+        System.out.println("start");
+        testNull();
+        var testNullMethod = TestNullAccess.class.getDeclaredMethod("testNull");
+        if (!wb.isMethodCompiled(testNullMethod)) {
+            throw new Exception("Method testMain should be compiled");
+        }
+    }
```
taofengliu, line 59: It seems that the test just make sure `testNull` can be compiled by Jeandle, but it does not check whether it runs as expected.

taofengliu (commented):
Looks good!
Just a simple suggestion here.

test/hotspot/jtreg/compiler/jeandle/bytecodeTranslate/field/TestNullAccess.java
Anjian-Wen: thanks for the advice, I have added the check.

test/hotspot/jtreg/compiler/jeandle/bytecodeTranslate/field/TestNullAccess.java
```diff
@@ -0,0 +1,61 @@
...
+        String str = null;
+        System.out.println(str);
+        return str;
+    }
+
+    public static void main(String[] args) throws Exception {
+        System.out.println("start");
+        String ret = testNull();
+        boolean flag = (ret == null) ? true : false;
+        Asserts.assertTrue(flag);
```
Cruise20:
```suggestion
        Asserts.assertTrue(ret == null);
```

Anjian-Wen: sure, good idea, I have fixed it

taofengliu:
@mergifyio queue

PR #208, opened 2025-11-11
enhance: Add monitor check for JeandleVMState::match

taofengliu:
## Related issue(s):

issue #206


src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -93,6 +94,16 @@ bool JeandleVMState::match(JeandleVMState* to_match) {
     }
   }
 
+  if (_locks.size() != to_match->_locks.size()) {
```
Cruise20, line 97: Add an null check before `push_lock`.

Cruise20 (approved):

Issue #188, opened 2025-11-04
Support throw/try/catch for AArch64


taofengliu:
**The feature you'd like**
Support throw/try/catch for AArch64
Pass all exception test for jeandle on AArch64

PR #213, opened 2025-11-12
feat: Support throw/try/catch for aarch64

minchao1024:
## Related issue(s):

issue #188 

Support throw/try/catch for aarch64.

### DIFF
--- a/src/hotspot/share/code/exceptionHandlerTable.hpp
+++ b/src/hotspot/share/code/exceptionHandlerTable.hpp
@@ -32,7 +32,6 @@
 // Here is a class to make JeandleExceptionHandlerTable compatible with ExceptionHandlerTable.
 class ExceptionHandlerTableInterface {
  public:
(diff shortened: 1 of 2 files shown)


src/hotspot/cpu/aarch64/jeandleRuntimeRoutine_aarch64.cpp
```diff
@@ -25,14 +25,162 @@
...
+    __ bind(retaddr);
+  }
+
+  OopMapSet* oop_maps = new OopMapSet();
+
+  oop_maps->add_gc_map(__ pc() - start, new OopMap(4 /* frame_size in slot_size(4 bytes) */, 0));
+
+  __ reset_last_Java_frame(false);
+
+  // Now the exception handler is in rscratch1.
```
feilongjiang:
```suggestion
  // Now the exception handler is in retval.
```

taofengliu (approved):

Issue #207, opened 2025-11-11
Add document for using the Dockerfile to build jeandle compilation environment


taofengliu:
**The feature you'd like**
Add document for using the Dockerfile to build jeandle compilation environment.

**Implementation you've considered**
Add it to 'getting started'

**Additional context**
#182 provided the Dockerfile.

PR #218, opened 2025-11-13
enhance: Add Dockerfile build instructions to getting started guide

BruceAko:
## Related issue(s):

issue #207 

Add document for using the Dockerfile to build jeandle compilation environment.

### DIFF
--- a/jeandle-docs/getting-started.md
+++ b/jeandle-docs/getting-started.md
@@ -40,14 +40,17 @@ Then the compiled JDK can be found in a directory like ```build/linux-x86_64-ser
 
 ### Build with Dockerfile
 
-Jeandle-jdk ships with a Dockerfile so you can build inside a reproducible container instead of installing every dependency on the host. Build the image once, then start a container that mirrors your user ID and mounts your home directory so existing SSH keys and sources stay accessible.
+Jeandle-jdk ships with a Dockerfile so you can build inside a reproducible container instead of installing every dependency on the host. Build the image once from the Dockerfile alone, then clone the repositories *inside* the container. This avoids cloning jeandle-jdk twice.
 
-1. Build the image:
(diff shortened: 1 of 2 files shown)


jeandle-docs/getting-started.md
```diff
@@ -38,6 +38,52 @@ make images
...
+   ```
+   cd jeandle-jdk
+   docker build -t jeandle-dev:latest -f Dockerfile .
+   ```
+2. Launch the container (runs as the current user and reuses your home folder):
+   ```
+   docker run -it --rm \
+     --name jeandle-dev \
+     -u $(id -u):$(id -g) \
+     -w /home/$(id -un) \
```
taofengliu, line 58: We can delete the `WORKDIR` command in the dockerfile since here we have pointed the working directory.

jeandle-docs/getting-started.md
```diff
@@ -38,6 +38,52 @@ make images
...
 Then the compiled JDK can be found in a directory like ```build/linux-x86_64-server-release/images/jdk/``` under the jeandle-jdk path.
 
+### Build with Dockerfile
+
+Jeandle-jdk ships with a Dockerfile so you can build inside a reproducible container instead of installing every dependency on the host. Build the image once, then start a container that mirrors your user ID and mounts your home directory so existing SSH keys and sources stay accessible.
+
+1. Build the image:
+   ```
+   cd jeandle-jdk
+   docker build -t jeandle-dev:latest -f Dockerfile .
```
taofengliu: Here, we assume that the jeandle-jdk repository has already been cloned; however, after starting the container, we clone jeandle-jdk again, which introduces a logical inconsistency.

BruceAko:
@taofengliu Fixed, PTAL.

Issue #115, opened 2025-09-28
Support oopmap on RISC-V


taofengliu:
**The feature you'd like**
Support oopmap on RISC-V

**Additional context**
Use `test/hotspot/jtreg/compiler/jeandle/TestSimpleFGC.java` as a test case.

Issue #114, opened 2025-09-28
Support oop handle resolving on RISC-V


taofengliu:
**The feature you'd like**
Support oop handle resolving on RISC-V

**Implementation you've considered**
According to related process on x86 and AArch64.

**Other Context**
Use `test/hotspot/jtreg/compiler/jeandle/bytecodeTranslate/TestFieldAccess.java` as a test case. This test will create an oop handle of class object when accessing static fields.

Issue #113, opened 2025-09-28
Support stack unwinding on RISC-V


taofengliu:
**The feature you'd like**
Hotspot traverses thread stacks in many scenarios, so we need to support this feature on RISC-V.

**Implementation you've considered**
We need to confirm how Hotspot performs stack traversal on the RISC-V architecture, such as which registers are used as stack frame registers, and whether we need to obtain the size of each stack.

Issue #112, opened 2025-09-28
Support safepoint poll on RISC-V


taofengliu:
**The feature you'd like**
Support safepoint poll on RISC-V.

**Implementation you've considered**
There are two things we need to consider:
- Set up the routine invocation process, then we can call safepoint handler in Jeandle compiled code.
- Support thread local storage, then we can access the poll bit in thread local storage.

**Other Context**
Use `test/hotspot/jtreg/compiler/jeandle/TestSafepointPoll.java` as a test case.

PR #226, opened 2025-11-27
feat: Initial support for RISCV

lgxbslgx:
## Related issue(s):

issue #110 #111 #112 #113 #114 #115

 This patch adds the related code for supporting RISCV architecture. 

- Implement most (if not all) features in RISCV architecture.
- Move method signatures and implementations of `JeandleAssembler::is_oop_reloc`, `JeandleAssembler::is_const_reloc`, `JeandleAssembler::emit_section_word_reloc`, `JeandleAssembler::emit_oop_reloc` to arch specific file.
- Move method implementation of `JeandleCompiledCode::build_exception_handler_table`, `JeandleCompiledCode::resolve_reloc_info` to arch specific file.
- Move class `JeandleReloc` and its sub-classes. Add some arch specific fields for `jeandle_oop_Relocation`, `jeandle_section_word_Relocation`. Because we need to add the field `_rel_offset` which records the offset of the instruction pair (`auipc` and `load/float_load`)  in RISCV.
- A simple implementation of the methods `pack_data_to` and `unpack_data` of the classes `jeandle_oop_Relocation` and `jeandle_section_word_Relocation` in RISCV because of the new field `_rel_offset`.
- Adjust several test cases to meet the need of RISCV.

Some TODO items in the future:
- Optimize methods `pack_data_to` and `unpack_data` of `jeandle_oop_Relocation` and `jeandle_section_word_Relocation`.

Testing locally:
- [x] Test group `tier1_compiler_jeandle` in RISCV
- [ ] Test group  `jtreg:hotspot:tier1` in RISCV
- [x] Test group  `jtreg:hotspot:tier1` in X64
- [ ] Test group  `jtreg:hotspot:tier1` in AArch64

### DIFF
--- a/src/hotspot/cpu/aarch64/jeandleCompiledCode_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/jeandleCompiledCode_aarch64.cpp
@@ -18,16 +18,10 @@
  *
  */
 
-
-#include "jeandle/__llvmHeadersBegin__.hpp"
-#include "llvm/BinaryFormat/Dwarf.h"
-
 #include "jeandle/jeandleAssembler.hpp"
 #include "jeandle/jeandleCompiledCode.hpp"
 #include "jeandle/jeandleCompilation.hpp"
-#include "jeandle/jeandleRegister.hpp"
 #include "jeandle/jeandleReloc.hpp"
-#include "jeandle/jeandleRuntimeRoutine.hpp"
 
 // Get the frame size from .stack_sizes section.
 void JeandleCompiledCode::setup_frame_size() {
--- a/src/hotspot/cpu/aarch64/jeandleOopAddrReloc_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/jeandleOopAddrReloc_aarch64.cpp
@@ -22,8 +22,6 @@
  *
  */
 
-#include "jeandle/__llvmHeadersBegin__.hpp"
-
 #include "jeandle/jeandleAssembler.hpp"
 #include "jeandle/jeandleReloc.hpp"
 
--- a/src/hotspot/cpu/riscv/jeandleCompiledCode_riscv.cpp
+++ b/src/hotspot/cpu/riscv/jeandleCompiledCode_riscv.cpp
@@ -24,7 +24,6 @@
 #include "jeandle/jeandleAssembler.hpp"
 #include "jeandle/jeandleCompiledCode.hpp"
 #include "jeandle/jeandleCompilation.hpp"
-#include "jeandle/jeandleRegister.hpp"
 #include "jeandle/jeandleReloc.hpp"
 #include "jeandle/jeandleRuntimeRoutine.hpp"
 
--- a/src/hotspot/share/jeandle/jeandleReloc.cpp
+++ b/src/hotspot/share/jeandle/jeandleReloc.cpp
@@ -22,11 +22,6 @@
  *
  */
 
-#include "jeandle/__llvmHeadersBegin__.hpp"
-#include "llvm/BinaryFormat/Dwarf.h"
-#include "llvm/ExecutionEngine/JITLink/riscv.h"
-#include "llvm/Object/FaultMapParser.h"
-
 #include "jeandle/jeandleAssembler.hpp"
 #include "jeandle/jeandleRuntimeRoutine.hpp"
 #include "jeandle/jeandleReloc.hpp"
--- a/src/hotspot/cpu/aarch64/jeandleOopReloc_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/jeandleOopReloc_aarch64.cpp
@@ -22,8 +22,6 @@
  *
  */
 
-#include "jeandle/__llvmHeadersBegin__.hpp"
-
 #include "jeandle/jeandleAssembler.hpp"
 #include "jeandle/jeandleReloc.hpp"
 
--- a/src/hotspot/cpu/aarch64/jeandleSectionWordReloc_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/jeandleSectionWordReloc_aarch64.cpp
@@ -22,8 +22,6 @@
  *
  */
 
-#include "jeandle/__llvmHeadersBegin__.hpp"
-
 #include "jeandle/jeandleAssembler.hpp"
 #include "jeandle/jeandleReloc.hpp"
 
--- a/src/hotspot/cpu/riscv/jeandleCompiledCall_riscv.cpp
+++ b/src/hotspot/cpu/riscv/jeandleCompiledCall_riscv.cpp
@@ -18,8 +18,6 @@
  *
  */
 
-#include "jeandle/__llvmHeadersBegin__.hpp"
-
 #include "jeandle/jeandleCompiledCall.hpp"
 
 #include "jeandle/__hotspotHeadersBegin__.hpp"
--- a/src/hotspot/cpu/riscv/jeandleOopAddrReloc_riscv.cpp
+++ b/src/hotspot/cpu/riscv/jeandleOopAddrReloc_riscv.cpp
@@ -22,8 +22,6 @@
  *
  */
 
-#include "jeandle/__llvmHeadersBegin__.hpp"
-
 #include "jeandle/jeandleAssembler.hpp"
 #include "jeandle/jeandleReloc.hpp"
 
--- a/src/hotspot/cpu/riscv/jeandleOopReloc_riscv.cpp
+++ b/src/hotspot/cpu/riscv/jeandleOopReloc_riscv.cpp
@@ -22,13 +22,7 @@
  *
  */
 
-#include "jeandle/__llvmHeadersBegin__.hpp"
-#include "llvm/BinaryFormat/Dwarf.h"
-#include "llvm/ExecutionEngine/JITLink/riscv.h"
-#include "llvm/Object/FaultMapParser.h"
-
 #include "jeandle/jeandleAssembler.hpp"
-#include "jeandle/jeandleRuntimeRoutine.hpp"
 #include "jeandle/jeandleReloc.hpp"
 
 bool JeandleOopReloc::pd_emit_reloc(JeandleAssembler& assembler) {
--- a/src/hotspot/cpu/riscv/jeandleSectionWordReloc_riscv.cpp
+++ b/src/hotspot/cpu/riscv/jeandleSectionWordReloc_riscv.cpp
@@ -24,10 +24,8 @@
 
 #include "jeandle/__llvmHeadersBegin__.hpp"
 #include "llvm/ExecutionEngine/JITLink/riscv.h"
-#include "llvm/Object/FaultMapParser.h"
 
 #include "jeandle/jeandleAssembler.hpp"
-#include "jeandle/jeandleRuntimeRoutine.hpp"
 #include "jeandle/jeandleReloc.hpp"
 
 bool JeandleSectionWordReloc::pd_emit_reloc(JeandleAssembler& assembler) {
--- a/src/hotspot/cpu/x86/jeandleCompiledCode_x86.cpp
+++ b/src/hotspot/cpu/x86/jeandleCompiledCode_x86.cpp
@@ -18,15 +18,10 @@
  *
  */
 
-#include "jeandle/__llvmHeadersBegin__.hpp"
-#include "llvm/BinaryFormat/Dwarf.h"
-
 #include "jeandle/jeandleAssembler.hpp"
 #include "jeandle/jeandleCompiledCode.hpp"
 #include "jeandle/jeandleCompilation.hpp"
-#include "jeandle/jeandleRegister.hpp"
 #include "jeandle/jeandleReloc.hpp"
-#include "jeandle/jeandleRuntimeRoutine.hpp"
 
 // Get the frame size from .stack_sizes section.
 void JeandleCompiledCode::setup_frame_size() {
--- a/src/hotspot/cpu/x86/jeandleOopAddrReloc_x86.cpp
+++ b/src/hotspot/cpu/x86/jeandleOopAddrReloc_x86.cpp
@@ -22,8 +22,6 @@
  *
  */
 
-#include "jeandle/__llvmHeadersBegin__.hpp"
-
 #include "jeandle/jeandleAssembler.hpp"
 #include "jeandle/jeandleReloc.hpp"
 
--- a/src/hotspot/cpu/x86/jeandleOopReloc_x86.cpp
+++ b/src/hotspot/cpu/x86/jeandleOopReloc_x86.cpp
@@ -22,8 +22,6 @@
  *
  */
 
-#include "jeandle/__llvmHeadersBegin__.hpp"
-
 #include "jeandle/jeandleAssembler.hpp"
 #include "jeandle/jeandleReloc.hpp"
 
--- a/src/hotspot/cpu/x86/jeandleSectionWordReloc_x86.cpp
+++ b/src/hotspot/cpu/x86/jeandleSectionWordReloc_x86.cpp
@@ -22,8 +22,6 @@
  *
  */
 
-#include "jeandle/__llvmHeadersBegin__.hpp"
-
 #include "jeandle/jeandleAssembler.hpp"
 #include "jeandle/jeandleReloc.hpp"
 


src/hotspot/cpu/aarch64/jeandleCompiledCode_aarch64.cpp
```diff
@@ -35,3 +38,51 @@ void JeandleCompiledCode::setup_frame_size() {
...
+                                       BytesPerWord/* AddressSize */);
+    llvm::DataExtractor::Cursor data_cursor(0 /* Offset */);
+
+    // Now decode exception handler table.
+    // See EHStreamer::emitExceptionTable in Jeandle-LLVM for corresponding encoding.
+
+    uint8_t header_encoding = data_extractor.getU8(data_cursor);
+    assert(data_cursor && header_encoding == llvm::dwarf::DW_EH_PE_omit, "invalid exception handler table header");
+
+    uint8_t type_encoding = data_extractor.getU8(data_cursor);;
```
feilongjiang:
```suggestion
    uint8_t type_encoding = data_extractor.getU8(data_cursor);
```

src/hotspot/cpu/riscv/jeandleAssembler_riscv.cpp
```diff
@@ -18,78 +18,271 @@
...
+  relocInfo::relocType rtype;
+  if (call->target() == SharedRuntime::get_resolve_opt_virtual_call_stub()) {
+    rtype = relocInfo::opt_virtual_call_type;
+  } else {
+    assert(call->target() == SharedRuntime::get_resolve_static_call_stub(), "illegal call target");
+    rtype = relocInfo::static_call_type;
+  }
+  Address call_addr = Address(call->target(), rtype);
+  // emit trampoline call for patch
+  __ trampoline_call(call_addr);
```
feilongjiang: We may need to check the return address of `trampoline_call` and stop compiling when the trampoline stub overflows.

src/hotspot/cpu/riscv/jeandleAssembler_riscv.cpp
```diff
@@ -18,78 +18,271 @@
...
 
 void JeandleAssembler::patch_routine_call_site(int inst_offset, address target) {
-  Unimplemented();
+  // Set insts_end to where to patch
+  address patch_pc = __ addr_at(inst_offset);
+  int insts_end_offset = __ code()->insts_end() - __ code()->insts_begin();
+  __ code()->set_insts_end(patch_pc);
+
+  // Patch
+  __ trampoline_call(Address(target, relocInfo::runtime_call_type));
```
feilongjiang: Check `trampoline_call` returns.

src/hotspot/cpu/riscv/jeandleAssembler_riscv.cpp
```diff
@@ -18,78 +18,271 @@
...
 
 void JeandleAssembler::patch_static_call_site(int inst_offset, CallSiteInfo* call) {
-  Unimplemented();
+  assert(call->type() == JeandleCompiledCall::STATIC_CALL, "illegal call type");
+
+  // The following `set_insts_end` conflicts with code buffer expansion,
+  // we need to confirm that stub code section has enough space before invoking `set_insts_end`.
+  int required_space = __ max_trampoline_stub_size();
+  if (__ code()->stubs()->maybe_expand_to_ensure_remaining(required_space)) {
+    guarantee(__ code()->blob() != nullptr, "CodeCache is full");
```
feilongjiang: should not crash when code cache is full.

src/hotspot/cpu/riscv/jeandleAssembler_riscv.cpp
```diff
@@ -18,78 +18,271 @@
...
 void JeandleAssembler::patch_ic_call_site(int inst_offset, CallSiteInfo* call) {
-  Unimplemented();
+  assert(inst_offset >= 0, "invalid call instruction address");
+  assert(call->type() == JeandleCompiledCall::DYNAMIC_CALL, "illegal call type");
+
+  // The following `set_insts_end` conflicts with code buffer expansion,
+  // we need to confirm that stub code section has enough space before invoking `set_insts_end`.
+  int required_space = __ max_trampoline_stub_size();
+  if (__ code()->stubs()->maybe_expand_to_ensure_remaining(required_space)) {
+    guarantee(__ code()->blob() != nullptr, "CodeCache is full");
```
feilongjiang: Should not crash when code cache is full.

src/hotspot/cpu/riscv/jeandleAssembler_riscv.cpp
```diff
@@ -18,78 +18,271 @@
...
 
 void JeandleAssembler::patch_external_call_site(int inst_offset, CallSiteInfo* call) {
-  Unimplemented();
+  assert(call->type() == JeandleCompiledCall::EXTERNAL_CALL, "illegal call type");
+
+  // The following `set_insts_end` conflicts with code buffer expansion,
+  // we need to confirm that stub code section has enough space before invoking `set_insts_end`.
+  int required_space = __ max_trampoline_stub_size();
+  if (__ code()->stubs()->maybe_expand_to_ensure_remaining(required_space)) {
+    guarantee(__ code()->blob() != nullptr, "CodeCache is full");
```
feilongjiang: Should not crash when code cache is full.

src/hotspot/cpu/riscv/jeandleAssembler_riscv.cpp
```diff
@@ -18,78 +18,271 @@
...
+  }
+
+  // Set insts_end to where to patch
+  address patch_pc = __ addr_at(inst_offset);
+  int insts_end_offset = __ code()->insts_end() - __ code()->insts_begin();
+  __ code()->set_insts_end(patch_pc);
+
+  // Patch
+  __ ic_call(call->target());
```
feilongjiang: Check the return address.

src/hotspot/cpu/riscv/jeandleAssembler_riscv.cpp
```diff
@@ -18,78 +18,271 @@
...
+    guarantee(__ code()->blob() != nullptr, "CodeCache is full");
+  }
+
+  // Set insts_end to where to patch
+  address patch_pc = __ addr_at(inst_offset);
+  int insts_end_offset = __ code()->insts_end() - __ code()->insts_begin();
+  __ code()->set_insts_end(patch_pc);
+
+  // Patch
+  __ trampoline_call(Address(call->target(), relocInfo::runtime_call_type));
```
feilongjiang: Check stub overflow.

src/hotspot/cpu/riscv/jeandleAssembler_riscv.cpp
```diff
@@ -18,78 +18,271 @@
...
+  return 16;
 }
 
 int JeandleAssembler::emit_exception_handler() {
-  Unimplemented();
-  return 0;
+  int stub_size = __ far_branch_size();
+  address stub = __ start_a_stub(stub_size);
+  if (stub == nullptr) {
+    JeandleCompilation::report_jeandle_error("CodeCache is full when emitting exception handler");
```
feilongjiang: Could be replace with `JEANDLE_ERROR_ASSERT_AND_RET_ON_FAIL`

src/hotspot/cpu/riscv/jeandleCompiledCode_riscv.cpp
```diff
@@ -18,8 +18,73 @@
...
+
 #include "jeandle/jeandleCompiledCode.hpp"
+#include "jeandle/jeandleCompilation.hpp"
 
+// Get the frame size from .stack_sizes section.
 void JeandleCompiledCode::setup_frame_size() {
-  Unimplemented();
+  SectionInfo section_info(".stack_sizes");
+  if (!ReadELF::findSection(*_elf, section_info)) {
+    JeandleCompilation::report_jeandle_error(".stack_sizes section not found");
```
feilongjiang: Use `JEANDLE_ERROR_ASSERT_AND_RET_VOID_ON_FAIL` macro instead.

src/hotspot/cpu/riscv/jeandleRuntimeRoutine_riscv.cpp
```diff
@@ -25,18 +25,198 @@
...
+  const Register exception_oop = x10;
+  const Register exception_pc  = x13;
+
+  address start = __ pc();
+
+  // Get the exception pc
+  __ ld(exception_pc, Address(xthread, JavaThread::exception_pc_offset()));
+
+  // Push the exception pc as return address. (for stack unwinding)
+  __ push_reg(exception_pc);
```
feilongjiang: `push_reg` was used to push regs onto `esp`, should we push onto `sp`?

src/hotspot/cpu/riscv/jeandleRuntimeRoutine_riscv.cpp
```diff
@@ -25,18 +25,198 @@
...
+
+  // Push the frame pointer
+  __ push_reg(fp);
```
feilongjiang: Maybe we could use `RegSet` to push `exception_pc` and `fp`.

src/hotspot/cpu/riscv/jeandleRuntimeRoutine_riscv.cpp
```diff
@@ -25,18 +25,198 @@
...
+  }
+
+  OopMapSet* oop_maps = new OopMapSet();
+
+  oop_maps->add_gc_map(__ pc() - start, new OopMap(4 /* frame_size in slot_size(4 bytes) */, 0));
+
+  __ reset_last_Java_frame(false);
+
+  // Now the exception handler is in retval.
+  __ addi(t0, retval, 0);
```
feilongjiang:
```suggestion
  __ mv(t0, retval);
```

src/hotspot/cpu/riscv/jeandleRuntimeRoutine_riscv.cpp
```diff
@@ -25,18 +25,198 @@
...
+
+  // Move the exception oop to x10. Exception handler will use this.
+  __ ld(exception_oop, Address(xthread, JavaThread::exception_oop_offset()));
```
feilongjiang, line 122: `ld` may clobber `t0` when Address mode is `Address::base_plus_offset`, we should move `retval` to `t0` right before `jal`.

src/hotspot/cpu/riscv/jeandleRuntimeRoutine_riscv.cpp
```diff
@@ -25,18 +25,198 @@
...
+
+  // Clear the exception oop so GC no longer processes it as a root.
+  __ sd(zr, Address(xthread, JavaThread::exception_oop_offset()));
+
+  // For not confusing exception handler, clear the exception pc.
+  __ sd(zr, Address(xthread, JavaThread::exception_pc_offset()));
+
+  // Pop the frame pointer
+  __ pop_reg(fp);
```
feilongjiang: `pop_reg` pops reg from `esp`. We should pop from`sp`.

src/hotspot/cpu/riscv/jeandleRuntimeRoutine_riscv.cpp
```diff
@@ -25,18 +25,198 @@
...
+
+  // Pop the exception pc to x13. Exception handler will use this.
+  __ pop_reg(exception_pc);
```
feilongjiang: Ditto.

src/hotspot/cpu/riscv/jeandleRuntimeRoutine_riscv.cpp
```diff
@@ -25,18 +25,198 @@
...
+  const Register retval = x10;
+
+  // incoming parameters
+  const Register exception_oop = x10;
+  const Register exception_pc  = x13;
+
+  address start = __ pc();
+
+  // Push the exception pc as return address. (for stack unwinding)
+  __ push_reg(exception_pc);
```
feilongjiang: push `exception_pc` and `fp` onto `sp`.

src/hotspot/cpu/riscv/jeandleRuntimeRoutine_riscv.cpp
```diff
@@ -25,18 +25,198 @@
...
+
+  // Push the frame pointer
+  __ push_reg(fp);
```
feilongjiang: Ditto.

src/hotspot/cpu/riscv/jeandleRuntimeRoutine_riscv.cpp
```diff
@@ -25,18 +25,198 @@
...
+
+  oop_maps->add_gc_map(__ pc() - start, new OopMap(4 /* frame_size in slot_size(4 bytes) */, 0));
+
+  __ reset_last_Java_frame(false);
+
+  // Clear the exception pc.
+  __ sd(zr, Address(xthread, JavaThread::exception_pc_offset()));
+
+  // Pop the frame pointer
+  __ pop_reg(fp);
```
feilongjiang: pop reg from `sp`.

src/hotspot/cpu/riscv/jeandleRuntimeRoutine_riscv.cpp
```diff
@@ -25,18 +25,198 @@
...
+
+  // Pop the exception pc to x13. Exception handler will use this.
+  __ pop_reg(exception_pc);
```
feilongjiang: Ditto.

src/hotspot/cpu/riscv/jeandleUtils_riscv.cpp
```diff
@@ -20,9 +20,28 @@
...
 #include "jeandle/jeandleUtils.hpp"
 
 void JeandleFuncSig::setup_description(llvm::Function* func, bool is_stub) {
-  Unimplemented();
+  func->setCallingConv(llvm::CallingConv::Hotspot_JIT);
+
+  func->setGC(llvm::jeandle::JeandleGC);
+
+  // RISCV must enable extensions manually.
+  func->addFnAttr("target-features", "+i,+m,+forced-atomics,+a,+f,+d");
```
feilongjiang: Do we need `c` and `v` extensions?

src/hotspot/cpu/riscv/relocInfo_riscv.cpp
```diff
@@ -119,6 +152,12 @@ void metadata_Relocation::pd_fix_value(address x) {
 
 #ifdef JEANDLE
```
feilongjiang, line 176: Is that routine used by Jeandle only? Seems it was defined under `USE_TRAMPOLINE_STUB_FIX_OWNER` macro.

src/hotspot/cpu/aarch64/jeandleCompiledCode_aarch64.cpp
lgxbslgx: Fixed.

src/hotspot/cpu/riscv/jeandleAssembler_riscv.cpp
lgxbslgx: Fixed.

src/hotspot/cpu/riscv/jeandleAssembler_riscv.cpp
lgxbslgx: Fixed.

src/hotspot/cpu/riscv/jeandleAssembler_riscv.cpp
lgxbslgx: Fixed.

src/hotspot/cpu/riscv/jeandleAssembler_riscv.cpp
lgxbslgx: Fixed.

src/hotspot/cpu/riscv/jeandleAssembler_riscv.cpp
lgxbslgx: Fixed.

src/hotspot/cpu/riscv/jeandleAssembler_riscv.cpp
lgxbslgx: Fixed.

src/hotspot/cpu/riscv/jeandleAssembler_riscv.cpp
lgxbslgx: Fixed.

src/hotspot/cpu/riscv/jeandleAssembler_riscv.cpp
lgxbslgx: Fixed.

src/hotspot/cpu/riscv/jeandleCompiledCode_riscv.cpp
lgxbslgx: Fixed.

src/hotspot/cpu/riscv/jeandleRuntimeRoutine_riscv.cpp
lgxbslgx: Fixed.

src/hotspot/cpu/riscv/jeandleRuntimeRoutine_riscv.cpp
lgxbslgx: Fixed.

src/hotspot/cpu/riscv/jeandleRuntimeRoutine_riscv.cpp
lgxbslgx: Fixed.

src/hotspot/cpu/riscv/jeandleRuntimeRoutine_riscv.cpp
lgxbslgx: I use `t1` instead.

src/hotspot/cpu/riscv/jeandleRuntimeRoutine_riscv.cpp
lgxbslgx: Fixed.

src/hotspot/cpu/riscv/jeandleRuntimeRoutine_riscv.cpp
lgxbslgx: Fixed.

src/hotspot/cpu/riscv/jeandleRuntimeRoutine_riscv.cpp
lgxbslgx: Fixed.

src/hotspot/cpu/riscv/jeandleRuntimeRoutine_riscv.cpp
lgxbslgx: Fixed.

src/hotspot/cpu/riscv/jeandleRuntimeRoutine_riscv.cpp
lgxbslgx: Fixed.

src/hotspot/cpu/riscv/jeandleRuntimeRoutine_riscv.cpp
lgxbslgx: Fixed.

src/hotspot/cpu/riscv/jeandleUtils_riscv.cpp
lgxbslgx: The `c` extension is not supported now. Because some code are not aligned.

I had not tested the  `v` extension before. Maybe it is better to add it when we need in the future.

src/hotspot/cpu/riscv/relocInfo_riscv.cpp
lgxbslgx: I just obey the convention of `X86`.

lgxbslgx: The related code in `relocInfo_x86.cpp`:

```C++
#ifdef JEANDLE
// Fix the call destination at the original code location after the code
// containing the trampoline stub has been moved.
void trampoline_stub_Relocation::pd_fix_owner_after_move() {
  address trampoline = addr();
  nativeCall_at(_owner)->set_destination(trampoline);
}
#endif // JEANDLE
```

src/hotspot/cpu/riscv/jeandleRuntimeRoutine_riscv.cpp
feilongjiang: > I use `t1` instead.

That would be better! According to [1], we should not use t0/x5 for JAL.


1. https://github.com/openjdk/jdk/blob/932556026d6d49fe6f74d4ec4afcb72448611766/src/hotspot/cpu/riscv/macroAssembler_riscv.hpp#L692-L714

src/hotspot/cpu/riscv/jeandleUtils_riscv.cpp
feilongjiang: We could enable/disable the RISC-V extension by command line in the future. For now, it's okay.

src/hotspot/cpu/riscv/jeandleRuntimeRoutine_riscv.cpp
```diff
@@ -25,18 +25,186 @@
...
+  __ push_reg(RegSet::of(exception_pc, fp), sp);
+
+  address frame_complete = __ pc();
+
+  {
+    Label retaddr;
+    __ set_last_Java_frame(sp, noreg, retaddr, t0);
+    __ addi(c_rarg0, xthread, 0);
+    __ la(t0, RuntimeAddress(CAST_FROM_FN_PTR(address, JeandleRuntimeRoutine::get_exception_handler)));
+    __ jal(t0);
```
feilongjiang: We should better use x6/t1 instead of x5/t0 for calls and jumps.

src/hotspot/cpu/riscv/jeandleRuntimeRoutine_riscv.cpp
```diff
@@ -25,18 +25,186 @@
...
+  __ sd(exception_pc, Address(xthread, JavaThread::exception_pc_offset()));
+
+  address frame_complete = __ pc();
+
+  {
+    Label retaddr;
+    __ set_last_Java_frame(sp, noreg, retaddr, t0);
+    __ addi(c_rarg0, xthread, 0);
+    __ la(t0, RuntimeAddress(CAST_FROM_FN_PTR(address, JeandleRuntimeRoutine::search_landingpad)));
+    __ jal(t0);
```
feilongjiang: Use t1/x6 instead of t0/x5 for jal.

src/hotspot/cpu/riscv/jeandleRuntimeRoutine_riscv.cpp
lgxbslgx: Fixed.

src/hotspot/cpu/riscv/jeandleRuntimeRoutine_riscv.cpp
lgxbslgx: Fixed.

src/hotspot/share/jeandle/jeandleCompiledCode.cpp
```diff
@@ -366,6 +373,18 @@ void JeandleCompiledCode::resolve_reloc_info(JeandleAssembler& assembler) {
...
+  llvm::DenseMap<std::pair<LinkBlock*, llvm::orc::ExecutorAddrDiff>, LinkEdge*> rel_high_edges;
+
+  for (auto* block : link_graph->blocks()) {
+    for (auto& edge : block->edges()) {
+      if (edge.getKind() == llvm::jitlink::riscv::EdgeKind_riscv::R_RISCV_PCREL_HI20) {
+        rel_high_edges[{block, edge.getOffset()}] = &edge;
+      }
+    }
+  }
+#endif
```
lgxbslgx: This file has three places which use `#if defined(RISCV64)`. It is not so good. I tried to move the related code to the arch specific files, but the refactor failed. Because `JeandleReloc` and its sub-classes (and some other code) prevent the refactor. 

Maybe we should refactor the related code in another patch. For example, we can move `JeandleReloc` and its sub-classes to their own file. The ultimate goal is to avoid and remove `#if defined(RISCV64)` of the shared files.

lgxbslgx:
I refactored the code (move the riscv related code to its own files) in order to avoid adding riscv related code to the shared files.

src/hotspot/share/code/relocInfo.hpp
```diff
@@ -485,7 +485,7 @@ class RelocationHolder {
   // provide that behavior.  If that changes, we can instead add a Relocation*
   // _reloc member to capture the result of the placement new, and use that to
   // access the base subobject.
-  static const size_t _relocbuf_size = 5 * sizeof(void*);
+  static const size_t _relocbuf_size = 6 * sizeof(void*);
```
feilongjiang, line 493: Is this buffer size expanding just for Jeandle? If `UseJeandleCompiler` is not enabled, would it still be true?

lgxbslgx: > Is this buffer size expanding just for Jeandle?

Only RISCV port of Jeandle  needs buffer size expanding. Should I add guard `JEANDLE` and `RISCV` to avoid unnecessary size expanding? Such as:

```C++
#if (defined(JEANDLE) && defined(RISCV64))
  static const size_t _relocbuf_size = 6 * sizeof(void*);
#else
  static const size_t _relocbuf_size = 5 * sizeof(void*);
#endif // JEANDLE && RISCV64
```

src/hotspot/share/jeandle/jeandleCompiledCode.cpp
```diff
@@ -364,74 +200,7 @@ void JeandleCompiledCode::resolve_reloc_info(JeandleAssembler& assembler) {
...
-        // Oop relocations.
-        assert((target_name).starts_with("oop_handle"), "invalid oop relocation name");
-        relocs.push_back(new JeandleOopReloc(static_cast<int>(block->getAddress().getValue() + edge.getOffset()), _oop_handles[(target_name)]));
-      } else {
-        // Unhandled relocations
-        ShouldNotReachHere();
-      }
-    }
-  }
+  pd_resolve_reloc(assembler, relocs, std::move(*graph_or_err));
```
```suggestion
  if (!pd_resolve_reloc(assembler, relocs, std::move(*graph_or_err))) {
    // The original code for both x86 and aarch64.
  }
```
We can keep the original code unchanged through this way.
The implementation of `pd_resolve_reloc` for x86 and aarch64 just return false.

lgxbslgx: The constructors of the class `JeandleSectionWordReloc`, `JeandleOopReloc` and `JeandleOopAddrReloc` have different fields in RISCV. (Please read the code around the macro `#ifndef RISCV` in the file `jeandleReloc.hpp` for more information. ) So we can't place the related code such as `new JeandleOopReloc(......)` in shared code `codejeandleCompiledCode.cpp` here.

src/hotspot/share/jeandle/jeandleReloc.cpp
```diff
@@ -0,0 +1,169 @@
...
+#include "llvm/Object/FaultMapParser.h"
+
+#include "jeandle/jeandleAssembler.hpp"
+#include "jeandle/jeandleRuntimeRoutine.hpp"
+#include "jeandle/jeandleReloc.hpp"
+
+using LinkBlock = llvm::jitlink::Block;
+using LinkEdge = llvm::jitlink::Edge;
+using LinkKind = llvm::jitlink::Edge::Kind;
+using LinkSymbol = llvm::jitlink::Symbol;
```
taofengliu: Already defined in `jeandleReloc.hpp`

src/hotspot/share/jeandle/jeandleReloc.hpp
```diff
@@ -0,0 +1,158 @@
...
+
+  // JeandleReloc should be allocated by arena. Independent from JeandleCompilationResourceObj
+  // to avoid ambiguous behavior during template specialization.
+  void* operator new(size_t size) throw();
+
+  void* operator new(size_t size, Arena* arena) throw() {
+    return arena->Amalloc(size);
+  }
+
+  void  operator delete(void* p) {} // nothing to do
```
taofengliu, line 67: Can we extend JeandleCompilationResourceObj now?

src/hotspot/share/code/relocInfo.cpp
```diff
@@ -817,6 +817,7 @@ address internal_word_Relocation::target() {
   return target;
 }
 
+#ifndef RISCV
```
taofengliu: Consider to use a function with `pd_` prefix to delete this.

src/hotspot/share/code/relocInfo.hpp
```diff
@@ -487,7 +487,11 @@ class RelocationHolder {
   // provide that behavior.  If that changes, we can instead add a Relocation*
   // _reloc member to capture the result of the placement new, and use that to
   // access the base subobject.
+#if (defined(JEANDLE) && defined(RISCV64))
+  static const size_t _relocbuf_size = 6 * sizeof(void*);
```
taofengliu, line 493: Add a comment to explain why we need to expand the size of relocbuf.

src/hotspot/share/jeandle/jeandleReloc.cpp
lgxbslgx: Deleted.

src/hotspot/share/jeandle/jeandleReloc.hpp
lgxbslgx: I just moved the code of `jeandleCompiledCode.cpp` to this file. So I think it is good to adjust it in another PR, which can make this PR clearer.

src/hotspot/share/code/relocInfo.cpp
lgxbslgx: Refactored.

src/hotspot/share/code/relocInfo.hpp
lgxbslgx: Added.

src/hotspot/cpu/aarch64/jeandleCompiledCode_aarch64.cpp
```diff
@@ -18,8 +18,16 @@
...
+
+#include "jeandle/__llvmHeadersBegin__.hpp"
+#include "llvm/BinaryFormat/Dwarf.h"
+
+#include "jeandle/jeandleAssembler.hpp"
 #include "jeandle/jeandleCompiledCode.hpp"
 #include "jeandle/jeandleCompilation.hpp"
+#include "jeandle/jeandleRegister.hpp"
+#include "jeandle/jeandleReloc.hpp"
+#include "jeandle/jeandleRuntimeRoutine.hpp"
```
taofengliu: Some not used headers here.

src/hotspot/cpu/aarch64/jeandleOopAddrReloc_aarch64.cpp
```diff
@@ -0,0 +1,32 @@
...
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ *
+ */
+
+#include "jeandle/__llvmHeadersBegin__.hpp"
```
taofengliu: We may don't need this header file here.
Please check other files to avoid unneeded headers, too.

src/hotspot/cpu/aarch64/jeandleCompiledCode_aarch64.cpp
lgxbslgx: I removed many unused headers.

taofengliu:
@mergifyio queue

PR #228, opened 2025-12-02
feat: Change relocation model to PIC

minchao1024:

Prior to this, Jeandle had been using the small code model and the static relocation model. However, this approach did not fully align with the requirements of the code cache in the JVM. For instance, on RISC-V, it led LLVM to assume that all symbols were located within the lower or upper 2GB address space. Similarly, on x86, LLVM would assume that all symbols resided in the address range from 0x00000000 to 0x7effffff, which could cause potential issues. Therefore, this PR changes the relocation model to PIC (Position-Independent Code), thereby eliminating LLVM's assumptions about the symbol address space.


taofengliu:
@mergifyio queue

PR #247, opened 2025-12-15
fix: confirm the stub code section has enough space to avoid expanding and relocating when patching

lgxbslgx:

When using `set_insts_end` to set the instruction end address, we need to confirm the stub code section has enough space and doesn't expand during patching. This patch invokes `CodeSection::maybe_expand_to_ensure_remaining` at the beginning of some `patch_*` method which can avoid the expansion and relocation of the code buffer.

### DIFF
--- a/src/hotspot/cpu/aarch64/jeandleAssembler_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/jeandleAssembler_aarch64.cpp
@@ -60,7 +60,10 @@ void JeandleAssembler::patch_static_call_site(int inst_offset, CallSiteInfo* cal
 
   // The following `set_insts_end` conflicts with code buffer expansion,
   // we need to confirm that stub code section has enough space before invoking `set_insts_end`.
-  __ code()->stubs()->maybe_expand_to_ensure_remaining(__ max_trampoline_stub_size());
+  int required_space = __ max_trampoline_stub_size();
+  if (__ code()->stubs()->maybe_expand_to_ensure_remaining(required_space)) {
+    guarantee(__ code()->blob() != nullptr, "CodeCache is full");
+  }
 
   address call_address = __ addr_at(inst_offset);
 
@@ -117,7 +120,10 @@ void JeandleAssembler::patch_ic_call_site(int inst_offset, CallSiteInfo* call) {
 
   // The following `set_insts_end` conflicts with code buffer expansion,
   // we need to confirm that stub code section has enough space before invoking `set_insts_end`.
-  __ code()->stubs()->maybe_expand_to_ensure_remaining(__ max_trampoline_stub_size());
+  int required_space = __ max_trampoline_stub_size();
+  if (__ code()->stubs()->maybe_expand_to_ensure_remaining(required_space)) {
+    guarantee(__ code()->blob() != nullptr, "CodeCache is full");
+  }
 
   address call_address = __ addr_at(inst_offset);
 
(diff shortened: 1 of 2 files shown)


src/hotspot/cpu/aarch64/jeandleAssembler_aarch64.cpp
```diff
@@ -57,6 +57,11 @@ void JeandleAssembler::emit_static_call_stub(int inst_offset, CallSiteInfo* call
 
 void JeandleAssembler::patch_static_call_site(int inst_offset, CallSiteInfo* call) {
   assert(call->type() == JeandleCompiledCall::STATIC_CALL, "illegal call type");
+
+  // The following `set_insts_end` conflicts with code buffer expansion,
+  // we need to confirm that stub code section has enough space before invoking `set_insts_end`.
+  __ code()->stubs()->maybe_expand_to_ensure_remaining(__ max_trampoline_stub_size());
```
minchao1024: Expanding memory allocation may fail, check the blob to avoid puzzling crashes.
```suggestion
  int required_space = __ max_trampoline_stub_size();
  if (__ code()->stubs()->maybe_expand_to_ensure_remaining(required_space)) {
    guarantee(__ code()->blob() != nullptr, "CodeCache is full");
  }
```

taofengliu: I think we can report a jeandle compilation error here, just stop this compilation.

minchao1024: Out of codecache was a very serious and unrecoverable error, which could easily lead to SIGSEGV, it needed to be handled immediately.

taofengliu: I noticed that C2 just record a failure when no enough space in codecache. According to src/hotspot/share/opto/output.cpp:3231

minchao1024: > I noticed that C2 just record a failure when no enough space in codecache. According to src/hotspot/share/opto/output.cpp:3231

it is a pure memory allocation, and C2 will check the error immediately
```
  // Initialize code buffer
  estimate_buffer_size(_buf_sizes._const);
  if (C->failing()) return;
```
but here in `patch_static_call_site`, we will write the codeblob, which could be `nullptr` after memory allocation.

-----------------------------------------------------

or we can return after report jeandle error, and check the error in every step.

src/hotspot/cpu/aarch64/jeandleAssembler_aarch64.cpp
```diff
@@ -94,6 +99,10 @@ void JeandleAssembler::patch_stub_C_call_site(int inst_offset, CallSiteInfo* cal
 }
 
 void JeandleAssembler::patch_routine_call_site(int inst_offset, address target) {
+  // The following `set_insts_end` conflicts with code buffer expansion,
+  // we need to confirm that stub code section has enough space before invoking `set_insts_end`.
+  __ code()->stubs()->maybe_expand_to_ensure_remaining(__ max_trampoline_stub_size());
+
```
minchao1024: No need to preallocate in `patch_routine_call_site`, because routine_call use the share trampoline stub, which will be generated in `finalize_stubs` later.

lgxbslgx: Fixed.

src/hotspot/cpu/aarch64/jeandleAssembler_aarch64.cpp
lgxbslgx: I don't have any idea about error handling here. I fixed it just now according to the suggestion and will change the code if you have another conclusion later.

taofengliu (approved):
@minchao1024 will handle the unresolved issues we discussed above.

Issue #232, opened 2025-12-04
Let LLVM introduce external symbols on aarch64


Cruise20:
**Goals**
As described in #222 . Support on aarch64

PR #250, opened 2025-12-16
feat: Introduce external symbols on aarch64

minchao1024:
## Related issue(s):

issue #232

1. Implement `JeandleAssembler::patch_external_call_site` on aarch64.
2. Use a more suitable ULP tolerance value `ULP_TOLERANCE`.

### DIFF
--- a/src/hotspot/cpu/aarch64/jeandleCompiledCall_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/jeandleCompiledCall_aarch64.cpp
@@ -24,8 +24,8 @@
 #include "nativeInst_aarch64.hpp"
 
 int JeandleCompiledCall::call_site_size(JeandleCompiledCall::Type call_type) {
-  if (call_type == JeandleCompiledCall::ROUTINE_CALL
-      || call_type == JeandleCompiledCall::EXTERNAL_CALL) {
+  if (call_type == JeandleCompiledCall::ROUTINE_CALL ||
+      call_type == JeandleCompiledCall::EXTERNAL_CALL) {
     return NativeInstruction::instruction_size;
   }
 


src/hotspot/cpu/aarch64/jeandleAssembler_aarch64.cpp
```diff
@@ -139,7 +139,31 @@ void JeandleAssembler::patch_ic_call_site(int inst_offset, CallSiteInfo* call) {
 }
 
 void JeandleAssembler::patch_external_call_site(int inst_offset, CallSiteInfo* call) {
-  Unimplemented();
+  assert(inst_offset >= 0, "invalid call instruction address");
+  assert(call->type() == JeandleCompiledCall::EXTERNAL_CALL, "legal call type");
```
feilongjiang:
```suggestion
  assert(call->type() == JeandleCompiledCall::EXTERNAL_CALL, "illegal call type");
```

src/hotspot/cpu/aarch64/jeandleCompiledCall_aarch64.cpp
```diff
@@ -24,7 +24,8 @@
 #include "nativeInst_aarch64.hpp"
 
 int JeandleCompiledCall::call_site_size(JeandleCompiledCall::Type call_type) {
-  if (call_type == JeandleCompiledCall::ROUTINE_CALL) {
+  if (call_type == JeandleCompiledCall::ROUTINE_CALL
+      || call_type == JeandleCompiledCall::EXTERNAL_CALL) {
```
feilongjiang:
```suggestion
  if (call_type == JeandleCompiledCall::ROUTINE_CALL ||
      call_type == JeandleCompiledCall::EXTERNAL_CALL) {
```

feilongjiang (approved):

Issue #249, opened 2025-12-16
Report a Jeandle error instead of crashing when CodeCache is full


minchao1024:
**Description**
Check the return value of `trampoline_call` or `code()->blob()` to determine if the codecache is full. If so, report a jeandle error and return immediately.

PR #253, opened 2025-12-18
enhance: Add unified macros for Jeandle error handling and early return

minchao1024:
## Related issue(s):

issue #249 

Introduce a set of macros to standardize error checking and early returns in Jeandle compilation:

- `JEANDLE_CRASH_ON_ERROR`: crash in debug builds if `JeandleCrashOnError` enabled.
- `JEANDLE_ERROR_ASSERT_AND_RET_[VOID_]ON_FAIL`: check condition + report + [crash +] return.
- `RETURN_[VOID_]ON_JEANDLE_ERROR`: check jeandle error + return.

### DIFF
--- a/src/hotspot/share/jeandle/jeandleCompilation.hpp
+++ b/src/hotspot/share/jeandle/jeandleCompilation.hpp
@@ -60,14 +60,6 @@ class JeandleCompilation : public StackObj {
 
   static JeandleCompilation* current() { return (JeandleCompilation*) ciEnv::current()->compiler_data(); }
 
-  const char* name() {
-    if (_method != nullptr) {
-      return _method->name()->as_utf8();
-    } else {
-      return _llvm_module->getName().data();
-    }
-  }
-
   // Error related:
   void report_error(const char* msg) {
     if (msg != nullptr) {
@@ -83,12 +75,15 @@ class JeandleCompilation : public StackObj {
 
   Arena* arena() { return _arena; }
 
+  const std::string name() { return _name; }
+
  private:
   Arena* _arena; // Hold compilation life-time objects (JeandleCompilationResourceObj).
   llvm::TargetMachine* _target_machine;
   llvm::DataLayout* _data_layout;
   ciEnv* _env;
   ciMethod* _method;
+  const std::string _name;
   int _entry_bci;
   std::unique_ptr<llvm::LLVMContext> _context;
   std::unique_ptr<llvm::Module> _llvm_module;
@@ -113,14 +108,14 @@ class JeandleCompilation : public StackObj {
 #define JEANDLE_CRASH_ON_ERROR(_error_msg)                            \
 do {                                                                  \
   if (JeandleCrashOnError) {                                          \
-    fatal("Compilation failed in '%s': %s", JeandleCompilation::current()->name(), _error_msg); \
+    fatal("Compilation failed in '%s': %s", JeandleCompilation::current()->name().c_str(), _error_msg); \
   }                                                                   \
 } while (0)
 #else
 #define JEANDLE_CRASH_ON_ERROR(_error_msg) (void)(0)
 #endif
 
-#define CHECK_AND_REPORT_JEANDLE_ERROR_VOID(p, msg)                   \
+#define JEANDLE_ERROR_ASSERT_AND_RET_VOID_ON_FAIL(p, msg)             \
 do {                                                                  \
   if (!(p)) {                                                         \
     JeandleCompilation::report_jeandle_error(msg);                    \
@@ -129,7 +124,7 @@ do {                                                                  \
   }                                                                   \
 } while (0)
 
-#define CHECK_AND_REPORT_JEANDLE_ERROR_RETURNVAL(p, msg, return_val)  \
+#define JEANDLE_ERROR_ASSERT_AND_RET_ON_FAIL(p, msg, return_val)      \
 do {                                                                  \
   if (!(p)) {                                                         \
     JeandleCompilation::report_jeandle_error(msg);                    \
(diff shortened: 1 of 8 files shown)


src/hotspot/share/jeandle/jeandleCompilation.hpp
```diff
@@ -99,4 +107,48 @@ class JeandleCompilation : public StackObj {
...
+} while (0)
+
+#define CHECK_AND_REPORT_JEANDLE_ERROR_RETURNVAL(p, msg, return_val)  \
+do {                                                                  \
+  if (!(p)) {                                                         \
+    JeandleCompilation::report_jeandle_error(msg);                    \
+    JEANDLE_CRASH_ON_ERROR(msg);                                      \
+    return return_val;                                                \
+  }                                                                   \
+} while (0)
```
taofengliu:
```suggestion
#define CHECK_AND_REPORT_JEANDLE_ERROR(p, msg)                   \
do {                                                                  \
  if (!(p)) {                                                         \
    JeandleCompilation::report_jeandle_error(msg);                    \
    JEANDLE_CRASH_ON_ERROR(msg);                                      \
    return;                                                           \
  }                                                                   \
} while (0)

#define CHECK_AND_REPORT_JEANDLE_ERROR(p, msg, return_val)  \
do {                                                                  \
  if (!(p)) {                                                         \
    JeandleCompilation::report_jeandle_error(msg);                    \
    JEANDLE_CRASH_ON_ERROR(msg);                                      \
    return return_val;                                                \
  }                                                                   \
} while (0)
```

src/hotspot/share/jeandle/jeandleCompilation.hpp
```diff
@@ -99,4 +107,48 @@ class JeandleCompilation : public StackObj {
...
+    return;                                                           \
+  }                                                                   \
+} while (0)
+
+#define CHECK_JEANDLE_ERROR_AND_RETURN_VAL(return_val)                \
+do {                                                                  \
+  if (JeandleCompilation::jeandle_error_occurred()) {                 \
+    return return_val;                                                \
+  }                                                                   \
+} while (0)
```
taofengliu:
```suggestion
#define CHECK_JEANDLE_ERROR_AND_RETURN()                         \
do {                                                                  \
  if (JeandleCompilation::jeandle_error_occurred()) {                 \
    return;                                                           \
  }                                                                   \
} while (0)

#define CHECK_JEANDLE_ERROR_AND_RETURN(return_val)                \
do {                                                                  \
  if (JeandleCompilation::jeandle_error_occurred()) {                 \
    return return_val;                                                \
  }                                                                   \
} while (0)
```

minchao1024: > Like C1's `BAILOUT`. But why we use `do {} while(0)` ?

We wrap the macro body in `do { ... } while (0)` to ensure it behaves as a single statement in all contexts.
Without this wrapper, if the macro were defined using just `{ ... }` or a bare `if` it would expand into multiple syntactic statements when followed by a semicolon.
This can break control flow, especially in `if-else` constructs. For example:
```
if (cond)
  if (...) { ... }; // or  {if (...) { ... }};
else
  do_something();
```
same as ->
```
if (cond) {
  if (...) { ... }
}
;
else
  do_something();
```
The trailing semicolon after the block turns into a null statement, causing the `else` to appear without a matching `if`, leading to a compile error like:
```
     error: 'else' without a previous 'if'
```

minchao1024: we need to use different macro names for `void-returning` and `value-returning` cases, because C++ macros do not support overloading

src/hotspot/share/jeandle/jeandleCompilation.hpp
```diff
@@ -60,6 +60,14 @@ class JeandleCompilation : public StackObj {
 
   static JeandleCompilation* current() { return (JeandleCompilation*) ciEnv::current()->compiler_data(); }
 
+  const char* name() {
+    if (_method != nullptr) {
+      return _method->name()->as_utf8();
+    } else {
+      return _llvm_module->getName().data();
+    }
+  }
```
taofengliu: The `_llvm_module` is not always not null. Maybe we can add a `_name` field in `JeandleCompilation`.

src/hotspot/share/jeandle/jeandleCompilation.hpp
taofengliu: OK, but the names now are not perfect. Let's think about it.

taofengliu, line 148:
```suggestion
#define JEANDLE_CRASH_ON_ERROR(_error_msg)                            \
do {                                                                  \
  if (JeandleCrashOnError) {                                          \
    fatal("Compilation failed in '%s': %s", JeandleCompilation::current()->name(), _error_msg); \
  }                                                                   \
} while (0)
#else
#define JEANDLE_CRASH_ON_ERROR(_error_msg) (void)(0)
#endif

#define JEANDLE_ERROR_ASSERT_AND_RET_VOID_ON_FAIL(p, msg)                   \
do {                                                                  \
  if (!(p)) {                                                         \
    JeandleCompilation::report_jeandle_error(msg);                    \
    JEANDLE_CRASH_ON_ERROR(msg);                                      \
    return;                                                           \
  }                                                                   \
} while (0)

#define JEANDLE_ERROR_ASSERT_AND_RET_ON_FAIL(p, msg, return_val)  \
do {                                                                  \
  if (!(p)) {                                                         \
    JeandleCompilation::report_jeandle_error(msg);                    \
    JEANDLE_CRASH_ON_ERROR(msg);                                      \
    return return_val;                                                \
  }                                                                   \
} while (0)

#define RETURN_VOID_ON_JEANDLE_ERROR()                         \
do {                                                                  \
  if (JeandleCompilation::jeandle_error_occurred()) {                 \
    return;                                                           \
  }                                                                   \
} while (0)

#define RETURN_ON_JEANDLE_ERROR(return_val)                \
do {                                                                  \
  if (JeandleCompilation::jeandle_error_occurred()) {                 \
    return return_val;                                                \
  }                                                                   \
} while (0)
```

src/hotspot/share/jeandle/jeandleCompilation.hpp
minchao1024: Fixed

src/hotspot/share/jeandle/jeandleCompilation.hpp
minchao1024: Done

taofengliu (approved):

Issue #256, opened 2025-12-23
Support Class initialization barrier on method entry


feilongjiang:
**The feature you'd like**
Support Class initialization barrier on method entry

**Implementation you've considered**
Implement something like C2 prologue, insert clinit_barrier when a method needs class init barrier on entry.

**Additional context**
Same logic at `MachPrologNode::emit` for C2 compiler.

PR #258, opened 2025-12-23
feat: Support class init barrier on mehtod entry

feilongjiang:
## Related issue(s):

issue #256 

Add class initialization barrier on method entry when needed.
Just like C2 `MachPrologNode::emit`, we should emit the `clinit_barrier` check logic before `build_frame`.
Also, add two new tests in `test/hotspot/jtreg/runtime/clinit/ClassInitBarrier.java` to make sure we hit the slowpath. (For debug, we can add macroAssembler `stop` at the beginning of slowpath, new tests will crash at `stop`.)

Testing:
- [x] ClassInitBarrier.java on aarch64
- [x] ClassInitBarrier.java on x64

### DIFF
--- a/src/hotspot/cpu/x86/jeandleAssembler_x86.cpp
+++ b/src/hotspot/cpu/x86/jeandleAssembler_x86.cpp
@@ -167,18 +167,8 @@ void JeandleAssembler::emit_verified_entry() {
 void JeandleAssembler::emit_clinit_barrier_on_entry(Klass* klass) {
   Label fallthrough;
   __ mov_metadata(rscratch1, klass);
-
-  // Fast path check: class is fully initialized
-  __ cmpb(Address(rscratch1, InstanceKlass::init_state_offset()), InstanceKlass::fully_initialized);
-  __ jcc(Assembler::equal, fallthrough);
-
-  // Fast path check: current thread is initializer thread
-  __ cmpptr(r15_thread, Address(rscratch1, InstanceKlass::init_thread_offset()));
-  __ jcc(Assembler::equal, fallthrough);
-
-  // Slow path
+  __ clinit_barrier(rscratch1, r15_thread, &fallthrough);
   __ jump(RuntimeAddress(SharedRuntime::get_handle_wrong_method_stub()));
-
   __ bind(fallthrough);
 }
 
--- a/src/hotspot/cpu/aarch64/jeandleAssembler_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/jeandleAssembler_aarch64.cpp
@@ -193,20 +193,8 @@ void JeandleAssembler::emit_verified_entry() {
 void JeandleAssembler::emit_clinit_barrier_on_entry(Klass* klass) {
   Label fallthrough;
   __ mov_metadata(rscratch2, klass);
-
-  // Fast path check: class is fully initialized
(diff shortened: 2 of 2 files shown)


test/hotspot/jtreg/runtime/clinit/ClassInitBarrier.java
```diff
@@ -45,6 +45,9 @@
  * @run main/othervm/native -Xbatch -XX:CompileCommand=dontinline,*::test* -XX:TieredStopAtLevel=1 -DTHROW=true  -XX:CompileCommand=exclude,*::static* -Xcheck:jni ClassInitBarrier
  * @run main/othervm/native -Xbatch -XX:CompileCommand=dontinline,*::test* -XX:-TieredCompilation  -DTHROW=false -XX:CompileCommand=exclude,*::static* -Xcheck:jni ClassInitBarrier
  * @run main/othervm/native -Xbatch -XX:CompileCommand=dontinline,*::test* -XX:-TieredCompilation  -DTHROW=true  -XX:CompileCommand=exclude,*::static* -Xcheck:jni ClassInitBarrier
+ *
+ * @run main/othervm/native -Xbatch -XX:CompileCommand=dontinline,*::test* -XX:-TieredCompilation -XX:+UseJeandleCompiler -XX:CompileOnly=.testInvokeStatic -XX:CompileOnly=.staticM -DTHROW=false -XX:CompileCommand=dontinline,*::static* -Xcheck:jni ClassInitBarrier
+ * @run main/othervm/native -Xbatch -XX:CompileCommand=dontinline,*::test* -XX:-TieredCompilation -XX:+UseJeandleCompiler -XX:CompileOnly=.testInvokeStatic -XX:CompileOnly=.staticM -DTHROW=true  -XX:CompileCommand=dontinline,*::static* -Xcheck:jni ClassInitBarrier
```
taofengliu: I suggest copy this file to jeandle's test directory, and make it used by jeandle.

minchao1024: Perhaps we can put this test file in the `test/hotspot/jtreg/compiler/jeandle` directory.

feilongjiang: Fixed. Introduce a new test under jeandle.

src/hotspot/cpu/x86/jeandleAssembler_x86.cpp
```diff
@@ -164,6 +164,24 @@ void JeandleAssembler::emit_verified_entry() {
...
+  // Fast path check: class is fully initialized
+  __ cmpb(Address(rscratch1, InstanceKlass::init_state_offset()), InstanceKlass::fully_initialized);
+  __ jcc(Assembler::equal, fallthrough);
+
+  // Fast path check: current thread is initializer thread
+  __ cmpptr(r15_thread, Address(rscratch1, InstanceKlass::init_thread_offset()));
+  __ jcc(Assembler::equal, fallthrough);
+
+  // Slow path
+  __ jump(RuntimeAddress(SharedRuntime::get_handle_wrong_method_stub()));
```
taofengliu: Can we just invoke `MacroAssembler::clinit_barrier` here?

feilongjiang: Yes, we can reuse `MacroAssembler::clinit_barrier`, fixed.

Issue #105, opened 2025-09-26
Create hs_err log when jeandle compiler abort in llvm code


kuaiwei:
**Goals**
If llvm failed on assertion, hotsport will abort but it can not create a hs_err log like other runtime failure. We need the log file for trouble shooting.

**Motivation**

**Description**

PR #262, opened 2025-12-25
enhancement: Add LLVM fatal error handler

BruceAko:
## Related issue(s):

issue #105 

If llvm failed on assertion, hotsport will abort but it can not create a hs_err log like other runtime failure. We need the log file for trouble shooting.

### DIFF
--- a/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
+++ b/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
@@ -196,10 +196,11 @@ void JeandleVMState::store(BasicType type, int index, llvm::Value* value) {
 }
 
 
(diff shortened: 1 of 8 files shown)


src/hotspot/share/jeandle/jeandleCompiler.cpp
```diff
@@ -34,6 +35,14 @@
 
 #include "jeandle/__hotspotHeadersBegin__.hpp"
 #include "runtime/arguments.hpp"
+#include "utilities/vmError.hpp"
+
+// LLVM fatal error handler that generates hs_err log
```
Cruise20:
```suggestion
// LLVM fatal error handler that dispatches the error to jvm.
```

Issue #260, opened 2025-12-24
Complete uncommon trap support for aarch64


taofengliu:
**The feature you'd like**
Uncommon trap support for aarch64

**Additional context**
For now, `test/hotspot/jtreg/compiler/jeandle/deoptimize/TestUncommonTrap.java` and `test/hotspot/jtreg/compiler/jeandle/TestNullCheck.java` will fail in aarch64.

PR #267, opened 2025-12-29
fix: Set CodeOffsets::Deopt to prevent inst tail and deopt head from overlapping

minchao1024:
## Related issue(s):

issue #260 

Set CodeOffsets::Deopt to prevent inst tail and deopt head from overlapping

The start address of the deopt region has special semantics (see `CompiledMethod::is_deopt_entry`). If the instruction (inst) region ends at the same address where the deopt region begins, it may lead to improper checks and cause a crash inside `frame::adjust_unextended_sp`.

### DIFF
--- a/src/hotspot/share/jeandle/jeandleCompiledCode.cpp
+++ b/src/hotspot/share/jeandle/jeandleCompiledCode.cpp
@@ -570,7 +570,7 @@ JeandleOopMap* JeandleCompiledCode::build_oop_map(StackMapParser& stackmaps, Sta
     num_deopts = third.getSmallConstant();
     assert(num_deopts >= 0, "negative number");
 
-    // bci goes firt in deopt operands
+    // bci goes first in deopt operands
     int bci = (location++)->getSmallConstant();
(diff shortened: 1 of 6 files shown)


feilongjiang (approved):

src/hotspot/share/jeandle/jeandleCompiledCode.cpp
```diff
@@ -297,12 +301,8 @@ void JeandleCompiledCode::finalize() {
 
   build_implicit_exception_table();
 
-  // generate shared trampoline stubs
-  bool success = _code_buffer.finalize_stubs();
-  JEANDLE_ERROR_ASSERT_AND_RET_VOID_ON_FAIL(success, "shared stub overflow");
-
   // No deopt support now.
```
Cruise20: Remove this comment

minchao1024: Fixed

taofengliu (approved):

PR #268, opened 2025-12-29
fix: Fix compiler debugging assertion macro spelling case error and correct some typos

BruceAko:

1. Fix compiler debugging assertion macro spelling case error 
2. Correct some typos


BruceAko:
@taofengliu @Cruise20 PTAL, thanks.

Issue #274, opened 2026-01-06
TargetMachine is not multi thread safe


taofengliu:
**Describe the bug**
TargetMachine in `JeandleCompiler` is not multi thread safe, and will lead to some bugs.

**Minimal reproduce steps**
Run renaissance benchmark with Jeandle, the JVM will crash randomly. If we create a new target machine in every compilation, the crash point will be stable.

PR #275, opened 2026-01-06
fix: Make TargetMachine thread local

taofengliu:
## Related issue(s):

issue #274

### DIFF
--- a/test/hotspot/jtreg/compiler/jeandle/TestStackOverflowCheck.java
+++ b/test/hotspot/jtreg/compiler/jeandle/TestStackOverflowCheck.java
@@ -27,11 +27,7 @@
 package compiler.jeandle;
 
 public class TestStackOverflowCheck {
-  // Volatile sink to keep the recursion side-effectful and prevent dead-code elimination.
-  private static volatile int sink;
-
   private static void recurse(int depth) {
-    sink = depth;
     recurse(depth + 1);
   }
 


src/hotspot/share/jeandle/jeandleCompiler.cpp
```diff
@@ -35,13 +35,9 @@
...
 
-JeandleCompiler::JeandleCompiler(llvm::TargetMachine* target_machine) :
-                                 AbstractCompiler(compiler_jeandle),
-                                 _target_machine(target_machine),
-                                 _data_layout(target_machine->createDataLayout()),
-                                 _template_buffer(nullptr) {}
+THREAD_LOCAL llvm::TargetMachine* JeandleCompiler::_target_machine = nullptr;
 
-JeandleCompiler* JeandleCompiler::create() {
+void JeandleCompiler::initialize_target_machine() {
```
Cruise20: Return a `bool` value to tell the caller whether it is successful.

src/hotspot/share/jeandle/jeandleCompiler.hpp
```diff
@@ -50,11 +50,16 @@ class JeandleCompiler : public AbstractCompiler {
...
+  bool initialize_target_machine();
+
+  static llvm::TargetMachine* target_machine() { return _target_machine; }
+
   llvm::DataLayout* data_layout() { return &_data_layout; }
 
  private:
-  std::unique_ptr<llvm::TargetMachine> _target_machine;
+  // TargetMachine is not multi-thread safe, so make it thread local here.
+  static THREAD_LOCAL llvm::TargetMachine* _target_machine;
```
minchao1024, line 63: Compiler threads may be destroyed, using a smart pointer here is safer.

taofengliu: Unique pointer cannot be thread local because of non-trivial destructor. We can delete it in `stopping_compiler_thread`

Cruise20 (approved):

Issue #273, opened 2026-01-05
Support compilation of synchronized methods


taofengliu:
**The feature you'd like**
Support compilation of synchronized methods

PR #279, opened 2026-01-08
feat: add synchronized method compilation support to Jeandle

feilongjiang:
## Related issue(s):

issue #273

Add synchronized method compilation support to Jeandle.
Implementation is similar to C2, added some lock/unlock logic (only slow path for now) at the beginning and return of the synchronized method.

Testing
- [x] `tier1_compiler_jeandle` on linux-aarch64 with release/slowdebug build
- [x] `tier1_compiler_jeandle` on linux-x64 with release/slowdebug build

### DIFF
--- a/test/hotspot/jtreg/compiler/jeandle/TestSynchronizedMethod.java
+++ b/test/hotspot/jtreg/compiler/jeandle/TestSynchronizedMethod.java
@@ -80,7 +80,11 @@ public class TestSynchronizedMethod {
         Asserts.assertEquals(s, 1000000, "s is not 1000000");
         Asserts.assertEquals(o.i, 1000000, "o.i is not 1000000");
 
+        // Load RuntimeException to avoid unloaded class in compilation. Thus the "throw" will not be an uncommon trap and a real unlock logic will be generated for the "throw".
         preInit();
+        Asserts.assertThrows(RuntimeException.class, () -> incEx(-1));
+
+        // Invoke it again to make sure the monitor has been unlocked.
         Asserts.assertThrows(RuntimeException.class, () -> incEx(-1));
     }
 }


src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -2268,20 +2296,38 @@ void JeandleAbstractInterpreter::monitorenter() {
...
+  JeandleCompilation::current()->set_has_monitors(true);
+  null_check(_jvm->raw_peek().value());
+
+  llvm::Value* obj = _jvm->apop();
+  shared_lock(obj);
+}
+
+void JeandleAbstractInterpreter::monitorexit() {
+  JeandleCompilation::current()->set_has_monitors(true);
+  // TODO: need to check if the monitor is balanced.
```
taofengliu: Since we have `check_can_parse`, this TODO can be removed.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -2252,14 +2273,21 @@ void JeandleAbstractInterpreter::multianewarray() {
...
+llvm::Value* JeandleAbstractInterpreter::shared_lock(llvm::Value* obj) {
+  if (!GenerateSynchronizationCode) {
+    return nullptr;
+  }
   // Allocate a BasicLock on stack.
   // Alloca insts should be in the entry block to be 'StaticAlloca'. Then they could be folded into prologue code.
-  llvm::IRBuilder entry_block_ir_builder(_block_builder->entry_block()->header_llvm_block()->getTerminator());
+  llvm::BasicBlock* header_block = _block_builder->entry_block()->header_llvm_block();
+  llvm::Instruction* terminator = header_block->getTerminator();
+  llvm::IRBuilder<> entry_block_ir_builder(header_block->getContext());
```
```suggestion
  llvm::IRBuilder<> entry_block_ir_builder(*_context);
```

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -2252,14 +2273,21 @@ void JeandleAbstractInterpreter::multianewarray() {
...
 
-void JeandleAbstractInterpreter::monitorenter() {
-  null_check(_jvm->raw_peek().value());
-
-  llvm::Value* obj = _jvm->apop();
-
+llvm::Value* JeandleAbstractInterpreter::shared_lock(llvm::Value* obj) {
+  if (!GenerateSynchronizationCode) {
+    return nullptr;
+  }
```
```
  if (!GenerateSynchronizationCode) {
    return nullptr;
  }
  assert(obj != nullptr, "sanity");
```

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -2268,20 +2296,38 @@ void JeandleAbstractInterpreter::monitorenter() {
...
-  // TODO: need to check if the monitor is balanced.
-  llvm::Value* obj = _jvm->apop();
+  return lock;
+}
 
-  llvm::Value* lock = _jvm->pop_lock();
+void JeandleAbstractInterpreter::shared_unlock(llvm::Value* obj, llvm::Value* lock) {
+  if (!GenerateSynchronizationCode) {
+    return;
+  }
```
```suggestion
  if (!GenerateSynchronizationCode) {
    return;
  }
  assert(obj != nullptr && lock != nullptr, "sanity");
```

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -2344,3 +2390,15 @@ void JeandleAbstractInterpreter::boundary_check(llvm::Value* array_oop, llvm::Va
   _ir_builder.SetInsertPoint(boundary_check_pass);
   _block->set_tail_llvm_block(boundary_check_pass);
 }
+
+void JeandleAbstractInterpreter::return_current(llvm::Value* value) {
+  if (_method && _method->is_synchronized()) {
+    shared_unlock(_sync_lock[0], _sync_lock[1]);
```
feilongjiang, line 2505: As we discussed offline, we should also consider the unlock when an exception is thrown.

```
When invoking a method for which ACC_SYNCHRONIZED is set, the executing thread enters a 
monitor, invokes the method itself, and exits the monitor whether the method invocation 
completes normally or abruptly. During the time the executing thread owns the monitor, no other 
thread may enter it. If an exception is thrown during invocation of the synchronized method and 
the synchronized method does not handle the exception, the monitor for the method is 
automatically exited before the exception is rethrown out of the synchronized method.
```

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -613,11 +614,31 @@ void JeandleAbstractInterpreter::interpret() {
...
+  if (_method && _method->is_synchronized()) {
+    JeandleCompilation::current()->set_has_monitors(true);
+    _jvm = _block_builder->entry_block()->VM_state();
+    // Setup Object Pointer
+    llvm::Value* lock_obj = nullptr;
+    if (_method->is_static()) {
+      llvm::Value* oop_handle = find_or_insert_oop(_method->holder()->java_mirror());
+      lock_obj = _ir_builder.CreateLoad(JeandleType::java2llvm(BasicType::T_OBJECT, *_context), oop_handle);
+    } else {
+      // pass the "this" pointer, which is Param0 from StartNode
```
```suggestion
      // Lock the "this" pointer, which is the first parameter.
```

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -613,11 +614,31 @@ void JeandleAbstractInterpreter::interpret() {
...
+
+    llvm::Value* lock = shared_lock(lock_obj);
+    if (lock != nullptr) {
+      _sync_lock[0] = lock_obj;
+      _sync_lock[1] = lock;
+    }
+  }
+
   // Create branch from the entry block.
   _ir_builder.CreateBr(current->header_llvm_block());
```
taofengliu, line 642: Move this before `initialize_VM_state`, then we don't need to handle the case that `terminator` is nullptr in `shared_lock`

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -2383,14 +2408,22 @@ void JeandleAbstractInterpreter::multianewarray() {
...
-  llvm::IRBuilder entry_block_ir_builder(_block_builder->entry_block()->header_llvm_block()->getTerminator());
+  llvm::BasicBlock* header_block = _block_builder->entry_block()->header_llvm_block();
+  llvm::Instruction* terminator = header_block->getTerminator();
+  llvm::IRBuilder<> entry_block_ir_builder(*_context);
+  if (terminator) {
+    entry_block_ir_builder.SetInsertPoint(terminator);
+  } else {
+    // if terminator is nullptr, we are at empty block
+    entry_block_ir_builder.SetInsertPoint(header_block);
+  }
```
taofengliu: The terminator cannot be nullptr as described above.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -2189,6 +2210,10 @@ void JeandleAbstractInterpreter::dispatch_exception_to_handler(llvm::Value* exce
   for (ciExceptionHandlerStream handlers(_method, _bytecodes.cur_bci()); !handlers.is_done(); handlers.next()) {
     ciExceptionHandler* handler = handlers.handler();
     if (handler->is_rethrow()) {
+      // unlock before the exception is rethrown out of the synchronized method
+      if (_method && _method->is_synchronized()) {
+        shared_unlock(_sync_lock[0], _sync_lock[1]);
+      }
```
taofengliu, line 2218: Can we add a test case for this?

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
feilongjiang: As discussed offline, we need to emit load/lock_slowpath before branching to the bci0 block. Therefore, for a synchronized method, it's not the right way to emit LLVM IR. 
We can eliminate the `terminator` check by refactoring `shared_lock` and adding the `lock` parameter.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
feilongjiang: Add new `incEx` test, fixed.

test/hotspot/jtreg/compiler/jeandle/TestSynchronizedMethod.java
```diff
@@ -0,0 +1,86 @@
...
+
+        for (Thread th : threads) {
+            th.join();
+        }
+
+        Asserts.assertEquals(s, 1000000, "s is not 1000000");
+        Asserts.assertEquals(o.i, 1000000, "o.i is not 1000000");
+
+        preInit();
+        Asserts.assertThrows(RuntimeException.class, () -> incEx(-1));
```
taofengliu, line 85:
```suggestion
        // Load RuntimeException to avoid unloaded class in compilation. Thus the "throw" will not be an uncommon trap and a real unlock logic will be generated for the "throw".
        preInit();
        Asserts.assertThrows(RuntimeException.class, () -> incEx(-1));

        // Invoke it again to make sure the monitor has been unlocked.
        Asserts.assertThrows(RuntimeException.class, () -> incEx(-1));
```

taofengliu (approved):

Issue #298, opened 2026-01-13
Incorrect llvm IR by java.lang.ModuleLayer::defineModules


Cruise20:
Reproducible command: `java -Xcomp -XX:-TieredCompilation -XX:+UseJeandleCompiler -XX:CompileCommand=compileonly,java.lang.ModuleLayer::defineModules -version`

hs_err log:
```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGABRT (0x6) at pc=0x00007f6d1a02a9fc (sent by kill), pid=305822, tid=305831
#
# JRE version: OpenJDK Runtime Environment (21.0.7) (fastdebug build 21.0.7-internal-adhoc.root.jeandle-jdk)
# Java VM: OpenJDK 64-Bit Server VM (fastdebug 21.0.7-internal-adhoc.root.jeandle-jdk, compiled mode, serial gc, linux-amd64)
# Problematic frame:
# C  [libc.so.6+0x969fc]  pthread_kill+0x12c
#
# Core dump will be written. Default location: /home/admin/logs/core.305822
#
# An error report file with more information is saved as:
# /home/hexuejin/repo/jeandle/jeandle-jdk/hs_err_pid305822.log
#
# Compiler replay data is saved as:
# /home/hexuejin/repo/jeandle/jeandle-jdk/replay_pid305822.log
#
# If you would like to submit a bug report, please visit:
#   https://bugreport.java.com/bugreport/crash.jsp
#
```

PR #305, opened 2026-01-14
fix: Add initialized exception handlers to work list even after uncommon traps

minchao1024:
## Related issue(s):

issue #298

Ensure initialized exception handlers are added to the work list even when uncommon traps occur.

### DIFF
--- a/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
+++ b/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
@@ -2090,6 +2090,10 @@ void JeandleAbstractInterpreter::do_array_store(BasicType basic_type) {
   null_check(array_ref);
   boundary_check(array_ref, index);
 
+  if (basic_type == T_OBJECT) {
+    array_store_check(_jvm->raw_peek().value(), array_ref);
+  }
+
   llvm::Value* value = nullptr;
   switch (basic_type) {
     case T_INT: {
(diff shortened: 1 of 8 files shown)


src/hotspot/share/jeandle/jeandleAbstractInterpreter.hpp
```diff
@@ -150,7 +150,8 @@ class JeandleBasicBlock : public JeandleCompilationResourceObj {
     is_compiled                   = 1 << 0,
     is_on_work_list               = 1 << 1,
     is_loop_header                = 1 << 2,
-    always_uncommon_trap          = 1 << 3,
+    is_exception_handler_in_use   = 1 << 3,
```
taofengliu: Can we use `_jvm != nullptr` instead?

minchao1024: Yes, I'll apply this change.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -979,8 +979,15 @@ void JeandleAbstractInterpreter::interpret_block(JeandleBasicBlock* block) {
...
-  // ignore successor blocks of uncommon trap
+  // If the block is marked as always_uncommon_trap, only process its initialized exception handler successors.
   if (block->is_set(JeandleBasicBlock::always_uncommon_trap)) {
+    for (JeandleBasicBlock* suc : block->successors()) {
+      if (suc->is_exception_handler() &&
+          suc->VM_state() != nullptr &&
+          !suc->is_set(JeandleBasicBlock::is_compiled)) {
+        add_to_work_list(suc);
+      }
+    }
```
taofengliu, line 990:
```suggestion
    for (JeandleBasicBlock* suc : block->successors()) {
      if (suc->VM_state() != nullptr &&
          !suc->is_set(JeandleBasicBlock::is_compiled)) {
          assert(suc->is_exception_handler(), "only exception handlers are expected to be initialized");
        add_to_work_list(suc);
      }
    }
```
Maybe this is a better version. All initialized blocks should be compiled, but we only find exception handlers are initialized

minchao1024: We can not assume this assert, because non-exception handlers can be initialized by other blocks, like:
```
{ // Entry block
  start();
}

if (flag) {
  { // Block A
    // uncommon trap
    UninitClass.doSomething();
  }
}

{ // Block B
  done();
}
```
`BlockB` is a successor of `BlockA`, and it will be initialized by `Entry block`.

taofengliu (approved):

Issue #286, opened 2026-01-12
Crash occurred when compiling java.lang.invoke.Method.invoke


minchao1024:
**Describe the bug**
Crash occurred when compiling java.lang.invoke.Method.invoke

**Minimal reproduce steps**
```
javac TestMethodHandle.java
java -Xcomp -XX:-TieredCompilation -XX:CompileCommand=compileonly,TestMethodHandle::testMH -XX:+UseJeandleCompiler TestMethodHandle
```

```
// TestMethodHandle.java
import java.lang.invoke.MethodHandle;
import java.lang.invoke.MethodHandles;
import java.lang.invoke.MethodType;

class TestMethodHandle {
    public static void main(String[] args) throws Throwable {
        Class.forName("java.lang.Integer");
        TestMethodHandle t = new TestMethodHandle();

        MethodHandles.Lookup lookup = MethodHandles.lookup();
        MethodType mt = MethodType.methodType(int.class, int.class, int.class);
        MethodHandle mh = lookup.findVirtual(TestMethodHandle.class, "add", mt);
        testMH(mh, t);
    }

    public static void testMH(MethodHandle mh, TestMethodHandle t) throws Throwable {
        mh.invoke(t, 10, 20);
    }

    public int add(int a, int b) {
        return a + b;
    }
}
```

PR #314, opened 2026-01-16
fix: Special handling for signature-polymorphic methods

minchao1024:
## Related issue(s):

issue #286 

Special handling for signature-polymorphic methods：
1. Mark signature-polymorphic method call with `_invokehandle` bytecode.
2. Use target method's signature instead of declared method's signature to build Fucntion.
3. Propagate the appendix oop from the constant pool.
4. The invocation kind of `invokehandle` is determined by the type of the target method: it is treated as a static call if the target is a static method, and as a dynamic call otherwise.

### DIFF
--- a/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
+++ b/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
@@ -1253,10 +1253,10 @@ void JeandleAbstractInterpreter::table_switch() {
 // Generate call instructions.
 void JeandleAbstractInterpreter::invoke() {
   bool will_link;
-  ciSignature* declared_signature = nullptr;
-  ciMethod* target = _bytecodes.get_method(will_link, &declared_signature);
+  ciSignature* method_signature = nullptr;
+  ciMethod* target = _bytecodes.get_method(will_link, &method_signature);
   ciKlass*  holder = _bytecodes.get_declared_method_holder();
-  assert(declared_signature != nullptr, "cannot be null");
+  assert(method_signature != nullptr, "cannot be null");
   assert(will_link == target->is_loaded(), "");
 
   const Bytecodes::Code bc = _bytecodes.cur_bc_raw();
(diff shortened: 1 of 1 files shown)


src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1292,13 +1292,23 @@ void JeandleAbstractInterpreter::invoke() {
...
 
-  if (bc == Bytecodes::_invokedynamic) {
-    if (_bytecodes.has_appendix()) {
-      llvm::Value* appendix_oop_handle = find_or_insert_oop(_bytecodes.get_appendix());
-      llvm::Value* appendix_oop = _ir_builder.CreateLoad(JeandleType::java2llvm(BasicType::T_OBJECT, *_context), appendix_oop_handle);
-      _jvm->push(T_OBJECT, appendix_oop);
-    }
+  // Push appendix argument (MethodType, CallSite, etc.), if one.
+  if (_bytecodes.has_appendix()) {
+    assert(bc == Bytecodes::_invokedynamic || bc == Bytecodes::_invokehandle,
```
taofengliu: Use Bytecodes::has_optional_appendix

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1292,13 +1292,23 @@ void JeandleAbstractInterpreter::invoke() {
...
+           "appendix only valid for invokedynamic or invokehandle");
+    llvm::Value* appendix_oop_handle = find_or_insert_oop(_bytecodes.get_appendix());
+    llvm::Value* appendix_oop = _ir_builder.CreateLoad(JeandleType::java2llvm(BasicType::T_OBJECT, *_context), appendix_oop_handle);
+    _jvm->push(T_OBJECT, appendix_oop);
+  }
+
+  // Special handling for signature-polymorphic methods
+  if (bc == Bytecodes::_invokedynamic ||
+      bc == Bytecodes::_invokehandle) {
```
taofengliu: Bytecodes::has_optional_appendix

minchao1024: Done

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
minchao1024: `Bytecodes::has_optional_appendix` cannot be used here because `has_optional_appendix` means that an appendix oop may be present.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
taofengliu: `method_signature`

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
minchao1024: Done

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
minchao1024: Done

taofengliu (approved):

PR #317, opened 2026-01-19
refactor: Clean code for uncommon trap

taofengliu:


src/hotspot/share/jeandle/jeandleCompiledCode.cpp
```diff
@@ -499,7 +499,7 @@ void JeandleCompiledCode::fill_one_scope_value(const StackMapParser& stackmaps,
     if (is_constant) {
       array->at_put_grow(index++, new ConstantIntValue(StackMapUtil::getConstantUint(stackmaps, location)));
     } else {
-      array->at_put_grow(index++, new_loc_value(location, Location::normal));
+      array->at_put_grow(index++, new_location_value(location, Location::normal));
```
feilongjiang: Do we need `index` for `locals` and `stack`? Seems it is not used afterwards.

taofengliu: fixed

Issue #323, opened 2026-01-21
Crash on assert(NativeDeoptInstruction::is_deopt_at(pc)) failed: check


Cruise20:
**Describe the bug**
```
# A fatal error has been detected by the Java Runtime Environment:
#
#  Internal Error (/home/hexuejin/repo/jeandle/jeandle-jdk/src/hotspot/share/code/nmethod.cpp:1215), pid=9064, tid=9898
#  assert(NativeDeoptInstruction::is_deopt_at(pc)) failed: check
#
# JRE version: OpenJDK Runtime Environment (21.0.7) (fastdebug build 21.0.7-internal-adhoc.root.jeandle-jdk)
# Java VM: OpenJDK 64-Bit Server VM (fastdebug 21.0.7-internal-adhoc.root.jeandle-jdk, mixed mode, serial gc, linux-amd64)
# Problematic frame:
# V  [libjvm.so+0x14b8316]  nmethod::make_deoptimized()+0x1a6
#
# Core dump will be written. Default location: /home/admin/logs/core.9064
#
```

**Minimal reproduce steps**
`jtreg -jdk:./build/jeandle-fastdebug/images/jdk/ -vmoptions:"-XX:+UseJeandleCompiler" -va test/hotspot/jtreg/compiler/c2/TestDeadDataLoopIGVN.java`

**Expected behavior**
Passed

PR #336, opened 2026-01-22
fix: disable VMContinuations for jeandle compiler

feilongjiang:
## Related issue(s):

issue #323

Jeandle does not currently support Virtual Threads, but `Continuations::enabled()` returns true.
`nmethod::make_deoptimized` may check `PostCallNop` and patch it when `VMContinuations` is `true`. But for Jeandle, we did not emit `PostCallNop` at all.
So we should check `VMContinuations` and set it to false when `UseJeandleCompiler` is enabled.

### DIFF
--- a/src/hotspot/share/runtime/vm_version.cpp
+++ b/src/hotspot/share/runtime/vm_version.cpp
@@ -28,20 +28,8 @@
 #include "memory/resourceArea.hpp"
 #include "runtime/vm_version.hpp"
 
-#ifdef JEANDLE
-#include "compiler/compiler_globals.hpp"
-#include "runtime/globals_extension.hpp"
-#endif // JEANDLE
-
 void VM_Version_init() {
   VM_Version::initialize();
-
-#ifdef JEANDLE
-  if (UseJeandleCompiler && VMContinuations) {
-    warning("VMContinuations is not supported with Jeandle Compiler.");
-    VMContinuations = false;
-  }
-#endif // JEANDLE
 
   if (log_is_enabled(Info, os, cpu)) {
     char buf[1024];
--- a/src/hotspot/share/compiler/compilerDefinitions.cpp
+++ b/src/hotspot/share/compiler/compilerDefinitions.cpp
@@ -663,6 +663,15 @@ void CompilerConfig::ergo_initialize() {
 #endif // PRODUCT
 
     UseCompressedClassPointers = false;
+
+    if (FLAG_IS_DEFAULT(VMContinuations)) {
+      FLAG_SET_DEFAULT(VMContinuations, false);
+    }
+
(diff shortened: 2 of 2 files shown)


src/hotspot/share/runtime/vm_version.cpp
```diff
@@ -28,9 +28,21 @@
...
+#include "compiler/compiler_globals.hpp"
+#include "runtime/globals_extension.hpp"
+#endif // JEANDLE
+
 void VM_Version_init() {
   VM_Version::initialize();
 
+#ifdef JEANDLE
+  if (UseJeandleCompiler && VMContinuations) {
+    warning("VMContinuations is not supported with Jeandle Compiler.");
```
feilongjiang: Do we need this warning message? It will be triggered multiple times when running jtreg tests.
E.g.:
```
$ make test TEST="test/hotspot/jtreg/compiler/codegen/BMI2.java" JTREG="VM_OPTIONS=-XX:+UseJeandleCompiler" CONF=slow
Building target 'test' in configuration 'linux-aarch64-server-slowdebug'
Running tests using JTREG control variable 'VM_OPTIONS=-XX:+UseJeandleCompiler'
Test selection 'test/hotspot/jtreg/compiler/codegen/BMI2.java', will run:
* jtreg:test/hotspot/jtreg/compiler/codegen/BMI2.java

Running test 'jtreg:test/hotspot/jtreg/compiler/codegen/BMI2.java'
OpenJDK 64-Bit Server VM warning: VMContinuations is not supported with Jeandle Compiler.
OpenJDK 64-Bit Server VM warning: VMContinuations is not supported with Jeandle Compiler.
```

taofengliu: We can move this to `src/hotspot/share/compiler/compilerDefinitions.cpp`.
Only print warning when this flag is set by command line.

feilongjiang: > We can move this to `src/hotspot/share/compiler/compilerDefinitions.cpp`.

Yes. Looks like compilerDefinitions.cpp is a better place.

> Only print warning when this flag is set by command line.

Not really, `VMContinuations` is set to `true` as the default value [1][2]. We should make it `false` by default for Jeandle.
1. https://github.com/jeandle/jeandle-jdk/blob/73de7954773edfe853c8fa94a1d3856e87a0bb22/src/hotspot/cpu/aarch64/globals_aarch64.hpp#L58
2. https://github.com/jeandle/jeandle-jdk/blob/73de7954773edfe853c8fa94a1d3856e87a0bb22/src/hotspot/cpu/x86/globals_x86.hpp#L83

feilongjiang: Fixed.

Issue #354, opened 2026-02-02
Crash with assert(address_from_scaled_offset(offset, base) == x, "just checkin'") in AArch64


minchao1024:
**Minimal reproduce steps**
```
wget https://repo1.maven.org/maven2/org/ow2/asm/asm/9.6/asm-9.6.jar
javac -cp .:asm-9.6.jar TriggerReadCode.java
/home/minchao/codehub/jeandle-jdk/build/linux-aarch64-server-fastdebug/images/jdk/bin/java -XX:+UseJeandleCompiler -cp .:asm-9.6.jar -Xbatch -XX:CompileThreshold=90 -XX:-TieredCompilation -XX:CompileCommand=compileonly,org.objectweb.asm.ClassReader::readCode TriggerReadCode
```
```
// TriggerReadCode.java
import org.objectweb.asm.ClassReader;
import org.objectweb.asm.ClassVisitor;
import org.objectweb.asm.MethodVisitor;
import org.objectweb.asm.Opcodes;

import java.io.IOException;

public class TriggerReadCode {

    public static void main(String[] args) {
        try {
            String className = "java.util.ArrayList";
            System.out.println("Loading class: " + className);
            long start = System.currentTimeMillis();
            for (int i = 0; i < 100; i++) {
                ClassReader cr = new ClassReader(className);
                ClassVisitor cv = new ClassVisitor(Opcodes.ASM9) {
                    @Override
                    public MethodVisitor visitMethod(int access, String name, String descriptor, String signature, String[] exceptions) {
                        return new MethodVisitor(Opcodes.ASM9) {};
                    }
                };
                cr.accept(cv, 0);
            }
            long end = System.currentTimeMillis();
            System.out.println("Done. Time taken: " + (end - start) + "ms");
        } catch (IOException e) {
            e.printStackTrace();
        }
    }
}
```

PR #355, opened 2026-02-05
fix: Add jeandle_oop_addr_relocation for switch

minchao1024:
## Related issue(s):

issue #354 

Introduced a new relocation type to support LLVM-generated data segments where the addresses of oops are stored. This ensures the VM can correctly patch these memory locations when the target oop or its container moves, which is essential for LLVM's switch-case table dispatching.

### DIFF
--- a/src/hotspot/share/code/relocInfo.cpp
+++ b/src/hotspot/share/code/relocInfo.cpp
@@ -817,6 +817,43 @@ address internal_word_Relocation::target() {
   return target;
 }
 
+void jeandle_section_word_Relocation::pack_data_to(CodeSection* dest) {
+  // Pack _target.
+  short* p = (short*) dest->locs_end();
+  normalize_address(_target, dest, true);
+
+  int sindex = _section;
+  assert(sindex != CodeBuffer::SECT_NONE, "sanity");
+  assert(_target != nullptr, "sanity");
+
+  // compress target address
+  CodeSection* sect = dest->outer()->code_section(sindex);
+  guarantee(sect->allocates2(_target), "must be in correct section");
+  address base = sect->start();
+  jint encoding = scaled_offset(_target, base);
+  assert((uint)sindex < (uint)CodeBuffer::SECT_LIMIT, "sanity");
+  assert(CodeBuffer::SECT_LIMIT <= (1 << section_width), "section_width++");
+
+  // pack compressed target and _offset
+  p = pack_2_ints_to(p, ((encoding << section_width) | sindex), _offset);
+
+  dest->set_locs_end((relocInfo*) p);
+}
+
+void jeandle_section_word_Relocation::unpack_data() {
+  // unpack compressed target and _offset
+  int compressed_target;
+  unpack_2_ints(compressed_target, _offset);
+
+  // decode to target and section
+  jint    encoding = (compressed_target >> section_width);
+  int     sindex = (compressed_target & ((1<<section_width)-1));
+  address base   = binding()->section_start(sindex);
+
+  _section = sindex;
+  _target  = address_from_scaled_offset(encoding, base);
+}
+
 //---------------------------------------------------------------------------------
 // Non-product code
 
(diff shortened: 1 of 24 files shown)


src/hotspot/share/jeandle/jeandleCompiledCode.cpp
```diff
@@ -206,6 +206,26 @@ class JeandleOopReloc : public JeandleReloc {
...
+    JeandleReloc(offset),
+    _oop_handle(oop_handle) {}
+
+  void emit_reloc(JeandleAssembler& assembler) override {
+    assembler.emit_oop_addr_reloc(offset(), _oop_handle);
+  }
+
+  void fixup_offset(int prolog_length) override {
+#ifdef ASSERT
+    _fixed_up = true;
```
Cruise20, line 226: There is no fixup here, but `_fixed_up` is set to `true`. It's confusing.
I suggest to add a comment to clarify that this reloc is located at const section, where no prolog is put at the beginning of. The `_fixed_up` is just for assertion.

src/hotspot/share/code/relocInfo.cpp
```diff
@@ -804,6 +804,13 @@ void internal_word_Relocation::fix_relocation_after_move(const CodeBuffer* src,
   set_value(target);
 }
 
+void jeandle_oop_addr_Relocation::fix_relocation_after_move(const CodeBuffer* src, CodeBuffer* dest) {
+  assert(addr_in_const(), "must in const section");
+  address old_addr = *(address*)addr();
+  int delta = dest - src;
+  address new_addr = old_addr + delta;
+  set_value(new_addr);
+}
```
Cruise20: It seems only suitable for `Pointer64` relocation kind which is specific for aarch64. Move it to relocInfo_aarch64.cpp ?  For `x86` and `riscv`, let it `Unimplemented()`

src/hotspot/share/jeandle/jeandleCompiledCode.cpp
minchao1024: Done

src/hotspot/share/code/relocInfo.cpp
minchao1024: Fixed

src/hotspot/share/code/relocInfo.cpp
```diff
@@ -804,6 +804,13 @@ void internal_word_Relocation::fix_relocation_after_move(const CodeBuffer* src,
   set_value(target);
 }
 
+void jeandle_oop_addr_Relocation::fix_relocation_after_move(const CodeBuffer* src, CodeBuffer* dest) {
+  assert(addr_in_const(), "must in const section");
+  address old_addr = *(address*)addr();
+  int delta = dest - src;
+  address new_addr = old_addr + delta;
+  set_value(new_addr);
```
Cruise20: The new_addr seems incorrect. Because the initial `old_addr` is always zero. And it relies the linker to set the correct value. The correct addr should get from `oop_Relocation::oop_addr()`.
Additionally, the `fix_oop_relocation()` is always called by GC. And for jeandle, we reference the oop indirectly through oop handle. We don't need to fix oop during GC.

src/hotspot/share/jeandle/jeandleCompiledCode.cpp
```diff
@@ -369,6 +392,7 @@ void JeandleCompiledCode::resolve_reloc_info(JeandleAssembler& assembler) {
   for (auto *block : link_graph->blocks()) {
     // Resolve relocations in the compiled code and constant pool.
     if (block->getSection().getName().compare(".text") != 0 &&
+        block->getSection().getName().compare(".data.rel.ro") != 0 &&
```
Cruise20: Is it better to use `starts_with` here?

minchao1024: Fixed

src/hotspot/share/code/relocInfo.cpp
minchao1024: The `jeandle_oop_addr_relocation` is initialized by `fix_oop_relocation`, then `old_addr` in `fix_relocation_after_move` can get the right value. `oop_addr` can't be accessed in `fix_relocation_after_move` function.
And we don't actually need to fix oop during GC, but the same thing happens in Hotspot C2 with UseZGC , so it's not a new problem. Maybe we can deal with it together later.

Cruise20:
@Mergifyio queue

Issue #360, opened 2026-02-09
Too much overhead of ic_miss_stub in compiled code


minchao1024:
**Describe the bug**

A significant performance regression has been observed where ic_miss_stub occupies an unexpectedly high percentage of CPU time in the flame graph. Preliminary analysis indicates that many virtual/interface call sites are failing to stabilize, leading to a constant "ping-pong" effect between compiled code and the Runtime's resolution logic.

Ideally, hot call sites should transition to a Monomorphic or Megamorphic state to avoid frequent stubs. However, current observations show excessive entries into SharedRuntime::handle_wrong_method_ic_miss, suggesting that the Inline Cache (IC) is being invalidated or mismanaged.

minchao1024:
C3-jit SerialGC 耗时(ms)
757.429
43075.996
11729.201
213826.936
77573.968
43519.378
17746.276
3265.185
1545.864
72906.21
32931.802
41665.679
113325.017
10643.23
23062.003
4788.601
22998.019
47202.767
3854.85
203017.011
6286.907
485288.734
55241.204
10470.236
203225.411

OpenJDK SerialGC 耗时(ms)
383.818
9731.569
2675.842
62414.564
13251.153
3715.392
1374.231
2742.943
276.39
14993.322
2521.699
12237.023
68993.342
1554.824
5511.748
1100.588
2472.956
6234.759
410.192
41808.504
1673.669
9076.755
5944.901
2528.47
74332.912

C3-jit SerialGC 耗时 / OpenJDK SerialGC 耗时
1.973 
4.426 
4.383 
3.426 
5.854 
11.713 
12.914 
1.190 
5.593 
4.863 
13.059 
3.405 
1.643 
6.845 
4.184 
4.351 
9.300 
7.571 
9.398 
4.856 
3.756 
53.465 
9.292 
4.141 
2.734

PR #362, opened 2026-02-09
fix: Enforce static binding for methods that cannot use Inline Caches

minchao1024:
## Related issue(s):

issue #360 

This patch fixes an issue where statically bindable methods were incorrectly treated as DYNAMIC_CALLs, leading to runtime resolution failures.

Methods that return true for `can_be_statically_bound()` (e.g., final, private, or super calls) do not possess a valid `vtable_index` (it is set to `nonvirtual_vtable_index`). Consequently, these methods are incompatible with the Inline Cache (IC) mechanism for the following reasons:

1. No Vtable Index: The IC miss stub relies on a valid vtable index to generate a vtable dispatch stub or to populate the cache.
2. Impossible to Patch: Since there is no vtable index, the runtime cannot update the Inline Cache nor transition the call site to a megamorphic vtable call.
3. Resolution Loop: `resolve_virtual_call_C` -> `fill_ic` -> `ic_miss` -> `ic_set_to_clean(resolve_virtual_call_C)`.

This change mandates `JeandleCompiledCall::STATIC_CALL` (using `resolve_opt_virtual_call_stub`) for all methods that `can_be_statically_bound()`, bypassing the Inline Cache entirely and ensuring the call site is linked directly to the target method.

### DIFF
--- a/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
+++ b/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
@@ -1407,8 +1407,6 @@ void JeandleAbstractInterpreter::invoke() {
     case Bytecodes::_invokevirtual:  // fall through
     case Bytecodes::_invokeinterface: {
       if (target->can_be_statically_bound()) {
-        assert((bc != Bytecodes::_invokeinterface) ||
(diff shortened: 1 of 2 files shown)


taofengliu:
@mergifyio queue

PR #363, opened 2026-02-09
fix: enable Jeandle compiler by default

lfeng14:
## Related issue(s):

#349 

enable Jeandle compiler by default and fix fastdebug build mode missing template.ll


lfeng14:
@Mergifyio rebase

Issue #92, opened 2025-09-20
Restrict rewriting certain calls to JeandleRuntimeRoutine to statepoints


taofengliu:
**Goals**
When defining a JeandleRuntimeRoutine, we can decide whether to rewrite calls into it to a statepoint.

**Motivation**
Some JeandleRuntimeRoutines do not need statepoints when we call them.

**Description**
Maybe we can add a parameter to the defining macro of JeandleRuntimeRoutine.
As described in https://github.com/jeandle/jeandle-jdk/pull/88#discussion_r2354477883

PR #368, opened 2026-02-13
enhance: Restrict rewriting certain calls to JeandleRuntimeRoutine to…

wazlyqa:
… statepoints

## Related issue(s):

issue #92 

Some JeandleRuntimeRoutines do not need statepoints when we call them.

### DIFF
--- a/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
+++ b/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
@@ -19,6 +19,7 @@
  */
 
 #include "jeandle/__llvmHeadersBegin__.hpp"
+#include "llvm/Analysis/ConstantFolding.h"
 #include "llvm/IR/Attributes.h"
 #include "llvm/IR/Jeandle/Attributes.h"
 #include "llvm/IR/Jeandle/GCStrategy.h"
@@ -1577,11 +1578,10 @@ bool JeandleAbstractInterpreter::inline_intrinsic(const ciMethod* target) {
 // Generate IR for calling into llvm FunctionCallee, without exception handling.
 llvm::CallInst* JeandleAbstractInterpreter::create_call(llvm::FunctionCallee callee, llvm::ArrayRef<llvm::Value *> args, llvm::CallingConv::ID calling_conv, llvm::ArrayRef<llvm::OperandBundleDef> deopt_bundle) {
   llvm::CallInst *call = _ir_builder.CreateCall(callee, args, deopt_bundle);
-  if (llvm::isa<llvm::ConstantExpr>(callee.getCallee())) {
-    llvm::ConstantExpr* callee_address = llvm::dyn_cast<llvm::ConstantExpr>(callee.getCallee());
-    llvm::ConstantInt* addr_value = llvm::cast<llvm::ConstantInt>(callee_address->getOperand(0));
-    address target_addr = (address)addr_value->getZExtValue();
-    if (JeandleRuntimeRoutine::is_gc_leaf(target_addr)) {
+  if (auto callee_constant = llvm::dyn_cast<llvm::Constant>(callee.getCallee())) {
+    llvm::ConstantInt* addr_value = llvm::dyn_cast<llvm::ConstantInt>(
+      llvm::ConstantFoldCastOperand(llvm::Instruction::PtrToInt, callee_constant, llvm::Type::getInt64Ty(*_context), _module.getDataLayout()));
+    if (addr_value != nullptr && JeandleRuntimeRoutine::is_gc_leaf((address)addr_value->getZExtValue())) {
       call->addFnAttr(llvm::Attribute::get(call->getContext(), "gc-leaf-function"));
     }
   }
--- a/src/hotspot/share/jeandle/jeandleCallVM.cpp
+++ b/src/hotspot/share/jeandle/jeandleCallVM.cpp
@@ -96,9 +96,9 @@ void JeandleCallVM::generate_call_VM(const char* name, address routine_address, 
   ir_builder.CreateCondBr(if_not_null, forward_exception_block, no_exception_block);
   ir_builder.SetInsertPoint(forward_exception_block);
 
-  llvm::CallInst* call_inst = ir_builder.CreateCall(JeandleRuntimeRoutine::install_exceptional_return_for_call_vm_callee(target_module), {});
-  call_inst->addFnAttr(llvm::Attribute::get(call_inst->getContext(), "gc-leaf-function"));
-  call_inst->setCallingConv(llvm::CallingConv::C);
+  llvm::CallInst* install_exceptional_return_call_inst = ir_builder.CreateCall(JeandleRuntimeRoutine::install_exceptional_return_for_call_vm_callee(target_module), {});
+  install_exceptional_return_call_inst->addFnAttr(llvm::Attribute::get(install_exceptional_return_call_inst->getContext(), "gc-leaf-function"));
+  install_exceptional_return_call_inst->setCallingConv(llvm::CallingConv::C);
 
   // Return
   llvm::Type* ret_type = func_type->getReturnType();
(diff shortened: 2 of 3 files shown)


src/hotspot/share/jeandle/jeandleCompiledCode.cpp
```diff
@@ -386,10 +388,15 @@ void JeandleCompiledCode::resolve_reloc_info(JeandleAssembler& assembler) {
 
         // TODO: Set the right bci.
         // JeandleCallReloc for a routine call site will be created during stackmaps resolving because an oopmap is required.
```
```suggestion
        // JeandleCallReloc for a non-gc-leaf routine call site will be created during stackmaps resolving because an oopmap is required.
```
Move this comment to line 398.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1578,6 +1578,9 @@ bool JeandleAbstractInterpreter::inline_intrinsic(const ciMethod* target) {
 // Generate IR for calling into llvm FunctionCallee, without exception handling.
 llvm::CallInst* JeandleAbstractInterpreter::create_call(llvm::FunctionCallee callee, llvm::ArrayRef<llvm::Value *> args, llvm::CallingConv::ID calling_conv, llvm::ArrayRef<llvm::OperandBundleDef> deopt_bundle) {
   llvm::CallInst *call = _ir_builder.CreateCall(callee, args, deopt_bundle);
+  if (llvm::isa<llvm::ConstantExpr>(callee.getCallee())) {
+    call->addFnAttr(llvm::Attribute::get(call->getContext(), "gc-leaf-function"));
+  }
```
taofengliu: Is this for unreachable direct routine? But this might override other cases and lead to misunderstandings.

wazlyqa: Yes, this is for unreachable direct routines. All unreachable direct routines are inherently gc-leaf. I think by design, any routine that might trigger GC would be implemented as reachable or indirect instead, so there's no risk of overriding non-leaf cases here.

taofengliu: As I understand, this introduces an implicit constraint: every call whose callee is constexpr must be an unreachable indirect routine, a constraint that might be inadvertently broken during subsequent development.

src/hotspot/share/jeandle/jeandleCompiledCode.cpp
wazlyqa: Done

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
wazlyqa: I've added a check using JeandleRuntimeRoutine::is_gc_leaf to determine whether the callee is a gc leaf routine

src/hotspot/share/jeandle/jeandleCallVM.cpp
```diff
@@ -97,6 +97,7 @@ void JeandleCallVM::generate_call_VM(const char* name, address routine_address,
   ir_builder.SetInsertPoint(forward_exception_block);
 
   llvm::CallInst* call_inst = ir_builder.CreateCall(JeandleRuntimeRoutine::install_exceptional_return_for_call_vm_callee(target_module), {});
```
taofengliu:
```suggestion
  llvm::CallInst* install_exceptional_return_call_inst = ir_builder.CreateCall(JeandleRuntimeRoutine::install_exceptional_return_for_call_vm_callee(target_module), {});
```

src/hotspot/share/jeandle/jeandleCompiledCode.cpp
```diff
@@ -116,7 +116,9 @@ class JeandleCallReloc : public JeandleReloc {
 
   void emit_reloc(JeandleAssembler& assembler) override {
     // Each call reloc has an oopmap, except for EXTERNAL_CALL.
-    assert((_call->type() != JeandleCompiledCall::EXTERNAL_CALL && _stack_map != nullptr) ||
+    assert((_call->type() != JeandleCompiledCall::EXTERNAL_CALL && _call->type() != JeandleCompiledCall::ROUTINE_CALL && _stack_map != nullptr) ||
+           (_call->type() == JeandleCompiledCall::ROUTINE_CALL && JeandleRuntimeRoutine::is_gc_leaf(_call->target()) && _stack_map == nullptr) ||
+           (_call->type() == JeandleCompiledCall::ROUTINE_CALL && !JeandleRuntimeRoutine::is_gc_leaf(_call->target()) && _stack_map != nullptr) ||
            (_call->type() == JeandleCompiledCall::EXTERNAL_CALL && _stack_map == nullptr),
            "unmatched call type and oopmap");
```
```suggestion
#ifdef ASSERT
    if (_call->type() == JeandleCompiledCall::ROUTINE_CALL) {
      assert(...);
    } else if (_call->type() == JeandleCompiledCall::EXTERNAL_CALL) {
      assert(...);
    } else {
      assert(...);
    }
#endif // ASSERT
```
This maybe more readable.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1578,6 +1578,14 @@ bool JeandleAbstractInterpreter::inline_intrinsic(const ciMethod* target) {
...
 llvm::CallInst* JeandleAbstractInterpreter::create_call(llvm::FunctionCallee callee, llvm::ArrayRef<llvm::Value *> args, llvm::CallingConv::ID calling_conv, llvm::ArrayRef<llvm::OperandBundleDef> deopt_bundle) {
   llvm::CallInst *call = _ir_builder.CreateCall(callee, args, deopt_bundle);
+  if (llvm::isa<llvm::ConstantExpr>(callee.getCallee())) {
+    llvm::ConstantExpr* callee_address = llvm::dyn_cast<llvm::ConstantExpr>(callee.getCallee());
+    llvm::ConstantInt* addr_value = llvm::cast<llvm::ConstantInt>(callee_address->getOperand(0));
+    address target_addr = (address)addr_value->getZExtValue();
+    if (JeandleRuntimeRoutine::is_gc_leaf(target_addr)) {
+      call->addFnAttr(llvm::Attribute::get(call->getContext(), "gc-leaf-function"));
+    }
+  }
```
```suggestion
  if (auto callee_address = llvm::dyn_cast<llvm::ConstantExpr>(callee.getCallee())) {
    llvm::ConstantInt* addr_value = llvm::dyn_cast<llvm::ConstantInt>(callee_address->getOperand(0));
    if (addr_value != nullptr && JeandleRuntimeRoutine::is_gc_leaf((address)addr_value->getZExtValue())) {
      call->addFnAttr(llvm::Attribute::get(call->getContext(), "gc-leaf-function"));
    }
  }
```
Line 1581 has checked `callee.getCallee()` is `llvm::ConstantExpr`, so line 1582 does not need to use `dyn_cast`.
But `callee_address` maybe not a `llvm::ConstantInt`, so we should use `dyn_cast` for it.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1578,6 +1578,12 @@ bool JeandleAbstractInterpreter::inline_intrinsic(const ciMethod* target) {
 // Generate IR for calling into llvm FunctionCallee, without exception handling.
 llvm::CallInst* JeandleAbstractInterpreter::create_call(llvm::FunctionCallee callee, llvm::ArrayRef<llvm::Value *> args, llvm::CallingConv::ID calling_conv, llvm::ArrayRef<llvm::OperandBundleDef> deopt_bundle) {
   llvm::CallInst *call = _ir_builder.CreateCall(callee, args, deopt_bundle);
+  if (auto callee_address = llvm::dyn_cast<llvm::ConstantExpr>(callee.getCallee())) {
+    llvm::ConstantInt* addr_value = llvm::dyn_cast<llvm::ConstantInt>(callee_address->getOperand(0));
```
taofengliu: We've missed something here. `callee_address` can be various instructions, we cannot assume it's first operand is the callee address.

taofengliu:
@mergifyio queue

Issue #373, opened 2026-02-27
StackMapUtil::getConstantUlong did not correctly handle the conversion from unsigned int to signed long


eapenzhan:
When filling a ScopeValue with a small long constant, the value is read directly as an unsigned integer. If it is then directly cast to a signed long, the resulting value may not be as expected.

This issue can be reproduced with the following test case.
```c++
public class Test {
	public static void main(String[] args) {
		for (String str: args) {
			System.out.println(Long.valueOf(str));
			System.out.println(Integer.valueOf(str));
		}
	}
}
```
The output is as follows.
```bash
$ java -XX:-TieredCompilation -Xcomp -XX:+UseJeandleCompiler -XX:CompileCommand=compileonly,java/lang/Long\$LongCache.*  Test 0 4 121 144
CompileCommand: compileonly java/lang/Long$LongCache.* bool compileonly = true
4294967296
0
4294967300
4
4294967417
121
144
144
```

Why 144 is print collectedly?
Because the LongCache initialization is only effected by the deoptimization , which is in range of -128 ~ 127

PR #369, opened 2026-02-25
fix: enable Jeandle compiler by default and update ProblemList.txt

lfeng14:
## Related issue(s):

issue #349

## What this PR does / why we need it:
Enable the Jeandle compiler by default, fix the missing template.ll in fastdebug build mode, and update ProblemList.txt.

PR #374, opened 2026-02-27
fix: correctly handle the conversion from unsigned int to signed long

eapenzhan:
## Related issue(s):

issue #373 

Fix the bug

### DIFF
--- a/src/hotspot/share/jeandle/jeandleCompiledCode.cpp
+++ b/src/hotspot/share/jeandle/jeandleCompiledCode.cpp
@@ -416,10 +416,7 @@ void JeandleCompiledCode::resolve_reloc_info(JeandleAssembler& assembler) {
         int inst_end_offset = JeandleAssembler::fixup_call_inst_offset(static_cast<int>(block->getAddress().getValue() + edge.getOffset()));
 
         // TODO: Set the right bci.
-        CallSiteInfo* call_info = new CallSiteInfo(JeandleCompiledCall::ROUTINE_CALL,
-                                                    target_addr,
-                                                    -1/* bci */,
-                                                    target_addr == JeandleRuntimeRoutine::get_routine_entry("uncommon_trap")/* has_deopt_operands */);
+        CallSiteInfo* call_info = new CallSiteInfo(JeandleCompiledCall::ROUTINE_CALL, target_addr, -1/* bci */);
         if (JeandleRuntimeRoutine::is_gc_leaf(target_addr)) {
           relocs.push_back(new JeandleCallReloc(inst_end_offset, _env, _method, nullptr /* no oopmap */, call_info));
         } else {
@@ -434,9 +431,7 @@ void JeandleCompiledCode::resolve_reloc_info(JeandleAssembler& assembler) {
         int inst_end_offset = JeandleAssembler::fixup_call_inst_offset(static_cast<int>(block->getAddress().getValue() + edge.getOffset()));
 
         // TODO: Set the right bci.
-        CallSiteInfo* call_info = new CallSiteInfo(JeandleCompiledCall::EXTERNAL_CALL,
-                                                   target_addr,
-                                                   -1/* bci */);
+        CallSiteInfo* call_info = new CallSiteInfo(JeandleCompiledCall::EXTERNAL_CALL, target_addr, -1/* bci */);
         // LLVM doesn't rewrite intrinsic calls to statepoints, so we don't need oopmaps for external calls.
         relocs.push_back(new JeandleCallReloc(inst_end_offset, _env, _method, nullptr /* no oopmap */, call_info));
       } else if (JeandleAssembler::is_section_word_reloc(target, edge.getKind())) {
@@ -685,19 +680,16 @@ JeandleStackMap* JeandleCompiledCode::parse_stackmap(StackMapParser& stackmaps, 
   auto second = *(location++);
   assert(location != record->location_end(), "must be in range");
 
+  auto third = *(location++);
+
   assert(first.getKind() == StackMapParser::LocationKind::Constant, "unexpected kind");
   assert(second.getKind() == StackMapParser::LocationKind::Constant, "unexpected kind");
+  assert(third.getKind() == StackMapParser::LocationKind::Constant, "unexpected kind");
 
-  int num_deopts = 0;
+  int num_deopts = third.getSmallConstant();
   bool reexecute = false;
-  if (call_info->has_deopt_operands()) {
+  if (num_deopts > 0) {
     assert(this->_method != nullptr, "must be method compilation");
-
-    auto third = *(location++);
-    assert(third.getKind() == StackMapParser::LocationKind::Constant, "unexpected kind");
-
-    num_deopts = third.getSmallConstant();
-    assert(num_deopts > 0, "negative number");
 
     // bci goes first in deopt operands
     int bci = (location++)->getSmallConstant();
--- a/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
+++ b/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
@@ -1449,7 +1449,7 @@ void JeandleAbstractInterpreter::invoke() {
 
   // Record this call.
   uint32_t id = _compiled_code.next_statepoint_id();
-  _compiled_code.push_non_routine_call_site(new CallSiteInfo(call_type, dest, _bytecodes.cur_bci(), true /* _has_deopt_operands */, id));
+  _compiled_code.push_non_routine_call_site(new CallSiteInfo(call_type, dest, _bytecodes.cur_bci(), id));
 
   // Every invoke instruction may throw exceptions, handle them here.
   DispatchedDest dispatched = dispatch_exception_for_invoke();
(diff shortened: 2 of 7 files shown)


taofengliu:
Please fix the CI failure

eapenzhan:
> It's better to add a test case for this bug fix

This issue can be reproduced by the jtreg test(**compiler/intrinsic/IntrinsicAvailableTest.java**) with an InitializationError. I simply extracted this issue into a minimal example for the bug report. Since there is already an existing test case that can verify this, I don't think it's necessary to create a separate one. What do you think?

taofengliu:
> > It's better to add a test case for this bug fix
> 
> This issue can be reproduced by the jtreg test(**compiler/intrinsic/IntrinsicAvailableTest.java**) with an InitializationError. I simply extracted this issue into a minimal example for the bug report. Since there is already an existing test case that can verify this, I don't think it's necessary to create a separate one. What do you think?

I agree. We can wait for https://github.com/jeandle/jeandle-jdk/pull/369 to merge first, then we can remove compiler/intrinsic/IntrinsicAvailableTest.java from the problem list in this PR.

eapenzhan:
> > > It's better to add a test case for this bug fix
> > 
> > 
> > This issue can be reproduced by the jtreg test(**compiler/intrinsic/IntrinsicAvailableTest.java**) with an InitializationError. I simply extracted this issue into a minimal example for the bug report. Since there is already an existing test case that can verify this, I don't think it's necessary to create a separate one. What do you think?
> 
> I agree. We can wait for #369 to merge first, then we can remove compiler/intrinsic/IntrinsicAvailableTest.java from the problem list in this PR.

There seems to be a slight misunderstanding here. The current fix only resolves the InitializationError; the jtreg test case will still fail due to the missing corresponding intrinsic method. So I'll go ahead and **add a supplementary test** case here.

eapenzhan:
> > > > It's better to add a test case for this bug fix
> > > 
> > > 
> > > This issue can be reproduced by the jtreg test(**compiler/intrinsic/IntrinsicAvailableTest.java**) with an InitializationError. I simply extracted this issue into a minimal example for the bug report. Since there is already an existing test case that can verify this, I don't think it's necessary to create a separate one. What do you think?
> > 
> > 
> > I agree. We can wait for #369 to merge first, then we can remove compiler/intrinsic/IntrinsicAvailableTest.java from the problem list in this PR.
> 
> There seems to be a slight misunderstanding here. The current fix only resolves the InitializationError; the jtreg test case will still fail due to the missing corresponding intrinsic method. So I'll go ahead and **add a supplementary test** case here.

Testcase is added.

test/hotspot/jtreg/compiler/jeandle/deoptimize/TestLongValueOf.java
```diff
@@ -0,0 +1,69 @@
...
+            long longVal = Long.valueOf(testCases[i]);
+
+            Asserts.assertTrue(longVal == results[i],
+                    "Mismatch for input \"" + testCases[i] + "\": " +
+                    "Long.valueOf = " + longVal + ", " +
+                    "Expected: " + testCases[i]
+            );
+        }
+    }
+}
```
taofengliu: Add a blank line at the end of file.

src/hotspot/share/jeandle/jeandleCompiledCode.cpp
```diff
@@ -883,7 +883,7 @@ uint32_t StackMapUtil::getConstantUint(const StackMapParser& parser, const Stack
 uint64_t StackMapUtil::getConstantUlong(const StackMapParser& parser, const StackMapParser::LocationAccessor& location) {
   switch (location.getKind()) {
   case StackMapParser::LocationKind::Constant:
-    return (uint64_t)location.getSmallConstant();
+    return static_cast<uint64_t>(static_cast<int32_t>(location.getSmallConstant()));
```
taofengliu: The result of casting a uint32_t to int32_t is implementation-defined. Will this lead to some problems here?

eapenzhan: In the vast majority of modern compilers (such as GCC, Clang, and MSVC), this conversion is defined as preserving the bit pattern, meaning the original 32-bit binary data is directly reinterpreted as a signed integer. Moreover, since we currently compile HotSpot using GCC with a specified standard version, this is effectively equivalent to a forced cast and will not introduce any issues.

test/hotspot/jtreg/compiler/jeandle/deoptimize/TestLongValueOf.java
eapenzhan: done

taofengliu (commented):
Please consider my proposal:

I think the core issue is that the behavior of `StackMapUtil::getConstantUlong` doesn’t match how it is actually used. In practice, its result is immediately converted to int64_t, but the function returns a uint64_t.

I believe we should introduce a new function, `StackMapUtil::getConstantUlongAsLong`, which performs the read in the way we expect for a long value and returns an int64_t.

The current behavior of `StackMapUtil::getConstantUlong`—reading a uint64_t—seems correct.

And the change of `StackMapUtil::getConstantUlong` will affect the behavior of `StackMapUtil::getConstantDouble`, we should consider this, too.

By the way, I think using `std::memcpy` is a better way to cast a uint32_t to int32_t

eapenzhan:
> Please consider my proposal:
> 
> I think the core issue is that the behavior of `StackMapUtil::getConstantUlong` doesn’t match how it is actually used. In practice, its result is immediately converted to int64_t, but the function returns a uint64_t.
> 
> I believe we should introduce a new function, `StackMapUtil::getConstantUlongAsLong`, which performs the read in the way we expect for a long value and returns an int64_t.
> 
> The current behavior of `StackMapUtil::getConstantUlong`—reading a uint64_t—seems correct.
> 
> And the change of `StackMapUtil::getConstantUlong` will affect the behavior of `StackMapUtil::getConstantDouble`, we should consider this, too.
> 
> By the way, I think using `std::memcpy` is a better way to cast a uint32_t to int32_t

I found the handle of StackMap Constant in [llvm/lib/CodeGen/StackMaps.cpp](https://github.com/jeandle/jeandle-llvm/blob/b9d6571cf2c30614713300a8d257c41702c5ed4b/llvm/lib/CodeGen/StackMaps.cpp#L236). The constant is treated as a signed number, even though it is stored as unsigned. During the processing of constants, they will be converted to signed numbers.

The actual range of values that are problematic is as follows — specifically, the range of numbers that can be converted from int64 to int32.

|符号|范围|
|  ----  | ----  |
|正数|0x0000_0000_0000_0000 ~ 0x0000_0000_7FFF_FFFF|
|负数|0xFFFF_FFFF_8000_0000 ~ 0xFFFF_FFFF_FFFF_FFFF|

When these values are truncated to int32, the upper 32 bits can be directly discarded. However, when they are restored back to int64, sign extension is not applied. The correct approach should involve sign extension, so an intermediate conversion to int32 followed by a conversion back to int64 is required. Since conversions between signed and unsigned numbers of the same bit width do not have this kind of issue, it is acceptable to return either a signed or unsigned 64-bit value in the end.

I believe the author's original intent was to have `StackMapUtil::getConstantUlong` read a uint64 in the way LLVM stores it, leaving it to the caller to decide whether to perform any further type conversion, without needing to define a separate function for each conversion. The handling for `StackMapUtil::getConstantDouble` is also fixed; otherwise, the original handling of Double would have been problematic as well.

Regarding the conversion from uint32 to int32, I think a direct forced cast should suffice. I've noticed that this is basically how it's implemented throughout the JVM, and I don't want to overcomplicate things. What do you think?

src/hotspot/share/jeandle/jeandleCompiledCode.cpp
```diff
@@ -585,7 +573,12 @@ void JeandleCompiledCode::fill_one_scope_value(const StackMapParser& stackmaps,
   switch (encode._basic_type) {
   case T_INT: {
     if (is_constant) {
+#if ( defined( __cplusplus ) && __cplusplus >= 202002L )
       array->append(new ConstantIntValue(StackMapUtil::getConstantUint(stackmaps, location)));
+#else
+      jint const_int = JeandleBitCast::bit_cast<jint>(StackMapUtil::getConstantUint(stackmaps, location));
+      array->append(new ConstantIntValue(const_int));
+#endif
```
```suggestion
      jint const_int = JeandleBitCast::bit_cast<jint>(StackMapUtil::getConstantUint(stackmaps, location));
      array->append(new ConstantIntValue(const_int));
```
Better not to include these macros.

src/hotspot/share/jeandle/jeandleCompiledCode.cpp
```diff
@@ -595,7 +588,12 @@ void JeandleCompiledCode::fill_one_scope_value(const StackMapParser& stackmaps,
     // 2 stack slots for long type
     array->append(new ConstantIntValue((jint)0));
     if (is_constant) {
+#if ( defined( __cplusplus ) && __cplusplus >= 202002L )
       array->append(new ConstantLongValue(StackMapUtil::getConstantUlong(stackmaps, location)));
+#else
+      jlong const_long = JeandleBitCast::bit_cast<jlong>(StackMapUtil::getConstantUlong(stackmaps, location));
+      array->append(new ConstantLongValue(const_long));
+#endif
```
```suggestion
      jlong const_long = JeandleBitCast::bit_cast<jlong>(StackMapUtil::getConstantUlong(stackmaps, location));
      array->append(new ConstantLongValue(const_long));
```

taofengliu:
@mergifyio queue

taofengliu:
@mergifyio queue

taofengliu:
@mergifyio queue

taofengliu:
@mergifyio requeue

PR #377, opened 2026-03-02
fix: Refine runtime calls for landingpad and monitor unlocking

minchao1024:

- Use `JRT_ENTRY_NO_ASYNC` for `search_landingpad` to prevent async exceptions during exception unwinding.
- Use direct_call for `monitor_unlocking` instead of call_vm to reduce overhead for this `JRT_LEAF` routine.


taofengliu:
@mergifyio queue

Issue #342, opened 2026-01-27
fatal error: Compilation failed in 'java.util.concurrent.ConcurrentHashMap.putVal


Cruise20:
**Describe the bug**
```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  Internal Error (/home/hexuejin/repo/jeandle/jeandle-jdk/src/hotspot/share/jeandle/jeandleCompilation.cpp:287), pid=490304, tid=490862
#  fatal error: Compilation failed in 'java.util.concurrent.ConcurrentHashMap.putVal(Ljava/lang/Object;Ljava/lang/Object;Z)Ljava/lang/Object;': module verify failed in Jeandle compilation
#
# JRE version: OpenJDK Runtime Environment (21.0.7) (fastdebug build 21.0.7-internal-adhoc.root.jeandle-jdk)
# Java VM: OpenJDK 64-Bit Server VM (fastdebug 21.0.7-internal-adhoc.root.jeandle-jdk, compiled mode, tiered, serial gc, linux-amd64)
# Problematic frame:
# V  [libjvm.so+0xf4beeb]  JeandleCompilation::compile_java_method()+0x42b
#
# Core dump will be written. Default location: /home/admin/logs/core.490304
#
# An error report file with more information is saved as:
```

**Minimal reproduce steps**
`serviceability/jvmti/GetLocalVariable/GetLocalVars.java`

PR #382, opened 2026-03-04
fix: erase invalid phi node

wazlyqa:
## Related issue(s):

issue #342 

Fix a bug where PHI nodes in LLVM IR could have fewer incoming edges
than the number of predecessor blocks, causing module verification to
fail.

### DIFF
--- a/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
+++ b/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
@@ -126,6 +126,8 @@ bool JeandleVMState::update_phi_nodes(JeandleVMState* income_jvm, llvm::BasicBlo
     llvm::PHINode* phi_node = llvm::cast<llvm::PHINode>(_locals[i].value());
 
     if (income_locals[i].is_null() || phi_node->getType() != income_locals[i].value()->getType()) {
+      assert(phi_node->use_empty(), "cannot use invalid local variable");
+      phi_node->eraseFromParent();
       invalidate_local(i);
       continue;
     }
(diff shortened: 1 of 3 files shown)


wazlyqa:
Before fix: PHI nodes with missing incoming edges
```
bci_222:  ; preds = %bci_97, %bci_169_normal_dest, %bci_155_null_check_pass, %bci_209, %bci_186_null_check_pass
  %96 = phi ptr addrspace(1) [ %182, %bci_186_null_check_pass ], [ %182, %bci_209 ]
```
%96 has 2 entries but bci_222 has 5 predecessors

After fix: PHI nodes with safe placeholder values on undefined paths
```
bci_222:  ; preds = %bci_97, %bci_169_normal_dest, %bci_155_null_check_pass, %bci_209, %bci_186_null_check_pass
  %96 = phi ptr addrspace(1) [ %189, %bci_186_null_check_pass ], [ %189, %bci_209 ], [ null, %bci_155_null_check_pass ], [ null, %bci_169_normal_dest ], [ %42, %bci_97 ]
```

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -126,7 +126,16 @@ bool JeandleVMState::update_phi_nodes(JeandleVMState* income_jvm, llvm::BasicBlo
...
+      llvm::Type* phi_type = phi_node->getType();
+      llvm::Value* safe_value;
+      if (phi_type->isPointerTy()) {
+        // Use null for object types.
+        safe_value = llvm::ConstantPointerNull::get(llvm::cast<llvm::PointerType>(phi_type));
+      } else {
+        // Use zero for primitive types.
+        safe_value = llvm::UndefValue::get(phi_type);
+      }
+      phi_node->addIncoming(safe_value, income_block);
```
```suggestion
      assert(phi_node->use_empty(), "cannot use invalid local variable");
      phi_node->eraseFromParent();
      invalidate_local(i);
```
The phi node is invalid, we can just remove it.

wazlyqa: Enabling JVMTI disables method liveness analysis, which leads to the generation of such phi nodes. Removing them will cause test failures in the JVMTI-enabled code paths.

taofengliu:
Remove the solved test from problem list.

taofengliu:
@mergifyio queue

Issue #380, opened 2026-03-04
SerialGC is forced even when explicitly disabled


rcjhd:
**Describe the bug**
```
java.lang.RuntimeException: 'Garbage collector not selected' missing from stdout/stderr
        at jdk.test.lib.process.OutputAnalyzer.shouldMatch(OutputAnalyzer.java:350)
        at gc.arguments.TestDisableDefaultGC.main(TestDisableDefaultGC.java:50)
        at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
        at java.base/java.lang.reflect.Method.invoke(Method.java:580)
        at com.sun.javatest.regtest.agent.MainWrapper$MainTask.run(MainWrapper.java:138)
        at java.base/java.lang.Thread.run(Thread.java:1583)

JavaTest Message: Test threw exception: java.lang.RuntimeException: 'Garbage collector not selected' missing from stdout/stderr
JavaTest Message: shutting down test
```
Test failed in tier1 testcase `test/hotspot/jtreg/gc/arguments/TestDisableDefaultGC.java`. Jeandle force the garbage collector to SerialGC after #198 . However, it's also set even when `-XX:-UseSerialGC` in cmdline.

**Minimal reproduce steps**
`java -XX:+UseJeandleCompiler -XX:-UseSerialGC -version`

Now can normally startup and UseSerialGC is true.

**Expected behavior**
```
Error occurred during initialization of VM
Garbage collector not selected (default collector explicitly disabled)
```

and `test/hotspot/jtreg/gc/arguments/TestDisableDefaultGC.java` should pass.

PR #383, opened 2026-03-04
fix: Do not force-enable UseSerialGC when it is explicitly disabled

rcjhd:
## Related issue(s):

issue #380 

Jeandle only supports Serial GC now, we enabled it by default. However, when `-XX:-UseSerialGC` is set, it is also enabled too that is not match the testcase intent. This PR fix that and set UseSerialGC by default with an ergonomical form.


taofengliu:
@mergifyio queue

Issue #381, opened 2026-03-04
Cannot find objects only on operand stack


wazlyqa:
**Describe the bug**
```
java.lang.RuntimeException: Unexpected Exception
        at jdk.test.lib.Asserts.fail(Asserts.java:717)
        at IterateHeapWithEscapeAnalysisEnabled$TestCase02b.run(IterateHeapWithEscapeAnalysisEnabled.java:547)
        at IterateHeapWithEscapeAnalysisEnabled.runTestCases(IterateHeapWithEscapeAnalysisEnabled.java:218)
        at IterateHeapWithEscapeAnalysisEnabled.main(IterateHeapWithEscapeAnalysisEnabled.java:188)
        at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
        at java.base/java.lang.reflect.Method.invoke(Method.java:580)
        at com.sun.javatest.regtest.agent.MainWrapper$MainTask.run(MainWrapper.java:138)
        at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: java.lang.RuntimeException: unexpected number of instances: expected 0 to equal 1
        at jdk.test.lib.Asserts.fail(Asserts.java:694)
        at jdk.test.lib.Asserts.assertEquals(Asserts.java:207)
        at jdk.test.lib.Asserts.assertEQ(Asserts.java:180)
        at IterateHeapWithEscapeAnalysisEnabled$TestCase02b.runTest(IterateHeapWithEscapeAnalysisEnabled.java:563)
        at IterateHeapWithEscapeAnalysisEnabled$TestCase02b.run(IterateHeapWithEscapeAnalysisEnabled.java:544)
        ... 6 more

JavaTest Message: Test threw exception: java.lang.RuntimeException: Unexpected Exception
JavaTest Message: shutting down test
```

**Minimal reproduce steps**
```
serviceability/jvmti/Heap/IterateHeapWithEscapeAnalysisEnabled.java
```
TestCase02b fails with unexpected number of instances: expected 0 to equal 1. The key difference from TestCase02a (which passes) is that the allocated object is never assigned to a local variable:

```
// TestCase02a — passes (object stored in local variable)
ABBox ab = new ABBox(this);
dontinline_endlessLoop(ab);
ab.synchronizedSlowInc();

// TestCase02b — fails (object only on operand stack)
[...]

PR #390, opened 2026-03-10
fix: add deopt bundle for safepoint

wazlyqa:
## Related issue(s):

issue #381 

Previously, `add_safepoint_poll` generated a call to the `safepoint_poll`
java op without any deopt bundle. This meant the LLVM stackmap record at
the safepoint poll call site had no deopt operands, so `parse_stackmap`
produced an empty `locals` array, and JVMTI could not find objects that
were only reachable via locals at that point.

### DIFF
--- a/src/hotspot/share/jeandle/jeandleCompiledCode.cpp
+++ b/src/hotspot/share/jeandle/jeandleCompiledCode.cpp
@@ -416,11 +416,7 @@ void JeandleCompiledCode::resolve_reloc_info(JeandleAssembler& assembler) {
         int inst_end_offset = JeandleAssembler::fixup_call_inst_offset(static_cast<int>(block->getAddress().getValue() + edge.getOffset()));
 
         // TODO: Set the right bci.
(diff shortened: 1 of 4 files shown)


src/hotspot/share/jeandle/jeandleCompiledCode.cpp
```diff
@@ -419,7 +419,8 @@ void JeandleCompiledCode::resolve_reloc_info(JeandleAssembler& assembler) {
         CallSiteInfo* call_info = new CallSiteInfo(JeandleCompiledCall::ROUTINE_CALL,
                                                     target_addr,
                                                     -1/* bci */,
-                                                    target_addr == JeandleRuntimeRoutine::get_routine_entry("uncommon_trap")/* has_deopt_operands */);
+                                                    target_addr == JeandleRuntimeRoutine::get_routine_entry("uncommon_trap") ||
+                                                    target_addr == JeandleRuntimeRoutine::get_routine_entry("safepoint_handler")/* has_deopt_operands */);
```
taofengliu: Remove has_deopt_operands related code

taofengliu:
@mergifyio queue

Issue #389, opened 2026-03-10
Crash with assert(external_guard || result != nullptr) failed: Invalid JNI handle


wazlyqa:
**Describe the bug**

```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  Internal Error (/home/zhouyuyu/jeandle-jdk/src/hotspot/share/runtime/jniHandles.inline.hpp:86), pid=2180414, tid=2180466
#  assert(external_guard || result != nullptr) failed: Invalid JNI handle
#
# JRE version: OpenJDK Runtime Environment (21.0.7) (fastdebug build 21.0.7-internal-adhoc.zhouyuyu.jeandle-jdk)
# Java VM: OpenJDK 64-Bit Server VM (fastdebug 21.0.7-internal-adhoc.zhouyuyu.jeandle-jdk, mixed mode, sharing, serial gc, linux-amd64)
# Problematic frame:
# V  [libjvm.so+0x93ead8]  JNIHandles::resolve_non_null(_jobject*)+0x268
#
# Core dump will be written. Default location: Core dumps may be processed with "/usr/lib/systemd/systemd-coredump %P %u %g %s %t %c %h" (or dumping to /home/zhouyuyu/jeandle-jdk/JTwork/scratch/core.2180414)
#
# If you would like to submit a bug report, please visit:
#   https://bugreport.java.com/bugreport/crash.jsp
#
```

**Minimal reproduce steps**

```
jtreg -vmoption:"-XX:+UseJeandleCompiler" -va ./test/hotspot/jtreg/compiler/c1/CanonicalizeGetModifiers.java
```

wazlyqa:
The test CanonicalizeGetModifiers.test() calls Class.getModifiers() on a null receiver. Since null_check was commented out in the Jeandle invoke() method, the null receiver was passed directly as an argument to the virtual call. The call dispatched into the interpreter, which executed JVM_GetClassModifiers with a null JNI handle. This triggered an assertion failure in JNIHandles::resolve_non_null(), causing a fatal VM crash.

rcjhd:
`java/util/Base64/TestBase64.java` has a same problem as this issue.

PR #394, opened 2026-03-11
fix: add null check for invoke

wazlyqa:
## Related issue(s):

issue #389 

Add explicit null check for receiver in `JeandleAbstractInterpreter::invoke()`. For bytecodes that have a receiver, we check the receiver for null before further processing. If null, an uncommon trap with `Reason_null_check` is generated to deoptimize back to the interpreter where `NullPointerException` will be thrown.

### DIFF
--- a/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
+++ b/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
@@ -1304,7 +1304,7 @@ void JeandleAbstractInterpreter::invoke() {
 
   llvm::Value* receiver_value = nullptr;
 
-  // If the receiver is a constant null, do not torture the system by attempting to call through it.  
+  // If the receiver is null, do not torture the system by attempting to call through it.  
   if (receiver) {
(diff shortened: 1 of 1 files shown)


src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1304,14 +1304,14 @@ void JeandleAbstractInterpreter::invoke() {
 
   llvm::Value* receiver_value = nullptr;
 
-  // TODO: To keep consistent with C2, but no suitable test case for now.
-  // if (receiver) {
-    // int receiver_depth = target->arg_size() - 1; // Index of stack slots where receiver locates.
-    // receiver_value = _jvm->raw_peek(receiver_depth).value();
+  // If the receiver is a constant null, do not torture the system by attempting to call through it.
```
```suggestion
  // If the receiver is null, do not torture the system by attempting to call through it.  
```
Here 'constant' maybe confused.

taofengliu:
@mergifyio queue

Cruise20:
LGTM

taofengliu:
@mergifyio requeue

Issue #388, opened 2026-03-07
Support deoptimize handler


eapenzhan:
Many jtreg tests (such as #322 and #341 ) include the `DeoptimizeALot` related option, whose implementation relies on the deoptimize handler. In order to pass these tests and verify the robustness of deoptimization, we need to implement this functionality.

**Implementation you've considered**
Add code stub of deopt handler

eapenzhan:
During the implementation, it was discovered that the scope does not carry the `is_method_handle_invoke` information, which makes it impossible to determine whether the corresponding MH stub for the deopt_handle needs to be installed. This issue can be resolved by adding the corresponding attribute to `CallsiteInfo` to record this information and correctly populating it when processing the stack map.

The implementation is with the PR.

PR #396, opened 2026-03-11
feat: Support deopt handler

eapenzhan:
## Related issue(s):

issue #388 

Support deoptimization when code is tagged as deprecated.

### DIFF
--- a/src/hotspot/cpu/aarch64/jeandleRegister_aarch64.hpp
+++ b/src/hotspot/cpu/aarch64/jeandleRegister_aarch64.hpp
@@ -36,7 +36,7 @@ public:
   }
 
   static const bool is_stack_pointer(Register reg) {
-    return reg == sp;
+    return reg == r31_sp;
   }
 
   static const Register decode_dwarf_register(int dwarf_encoding) {
@@ -49,7 +49,7 @@ private:
     r0, r1, r2, r3, r4, r5, r6, r7, r8, r9, r10,
     r11, r12, r13, r14, r15, r16, r17, r18_tls, r19, r20,
     r21, r22, r23, r24, r25, r26, r27, r28, r29, r30,
-    sp
+    r31_sp
   };
 };
 
--- a/src/hotspot/cpu/aarch64/jeandleRegister_aarch64.hpp
+++ b/src/hotspot/cpu/aarch64/jeandleRegister_aarch64.hpp
@@ -39,6 +39,10 @@ public:
     return reg == r31_sp;
   }
 
+  static const bool is_frame_pointer(Register reg) {
+    return reg == r29;
+  }
+
   static const Register decode_dwarf_register(int dwarf_encoding) {
     assert(dwarf_encoding >=0 && dwarf_encoding < Register::number_of_registers, "invalid dwarf register number");
     return _dwarf_registers[dwarf_encoding];
--- a/src/hotspot/cpu/aarch64/jeandleRegister_aarch64.hpp
+++ b/src/hotspot/cpu/aarch64/jeandleRegister_aarch64.hpp
@@ -39,10 +39,6 @@ public:
     return reg == r31_sp;
   }
 
-  static const bool is_frame_pointer(Register reg) {
-    return reg == r29;
-  }
-
   static const Register decode_dwarf_register(int dwarf_encoding) {
     assert(dwarf_encoding >=0 && dwarf_encoding < Register::number_of_registers, "invalid dwarf register number");
     return _dwarf_registers[dwarf_encoding];
--- a/src/hotspot/cpu/x86/jeandleAssembler_x86.cpp
+++ b/src/hotspot/cpu/x86/jeandleAssembler_x86.cpp
@@ -193,6 +193,22 @@ int JeandleAssembler::emit_exception_handler() {
   return offset;
 }
 
+int JeandleAssembler::deopt_handler_size() {
+  return 17;
+}
+
+int JeandleAssembler::emit_deopt_handler() {
+  address base = __ start_a_stub(deopt_handler_size());
+  JEANDLE_ERROR_ASSERT_AND_RET_ON_FAIL(base != nullptr, "deopt handler stub overflow", 0);
+  int offset = __ offset();
+  InternalAddress here(__ pc());
+  __ pushptr(here.addr(), rscratch1);
+  __ jump(RuntimeAddress(JeandleRuntimeRoutine::get_routine_entry(JeandleRuntimeRoutine::_deopt_blob)));
+  assert(__ offset() - offset <= deopt_handler_size(), "overflow");
+  __ end_a_stub();
+  return offset;
+}
+
 using LinkKind_x86_64 = llvm::jitlink::x86_64::EdgeKind_x86_64;
 
 void JeandleAssembler::emit_section_word_reloc(int operand_offset, LinkKind kind, int64_t addend, address target, int reloc_section) {
--- a/src/hotspot/cpu/x86/jeandleAssembler_x86.cpp
+++ b/src/hotspot/cpu/x86/jeandleAssembler_x86.cpp
@@ -205,8 +205,22 @@ int JeandleAssembler::emit_deopt_handler() {
     JEANDLE_REPORT_ERROR_AND_RET("deopt handler stub overflow", 0);
   }
   int offset = __ offset();
+
+#ifdef _LP64
+  address the_pc = (address) __ pc();
+  Label next;
+  // push a "the_pc" on the stack without destroying any registers
+  // as they all may be live.
+
+  // push address of "next"
+  __ call(next, relocInfo::none); // reloc none is fine since it is a disp32
+  __ bind(next);
+  // adjust it so it matches "the_pc"
+  __ subptr(Address(rsp, 0), __ offset() - offset);
+#else
   InternalAddress here(__ pc());
-  __ pushptr(here.addr(), rscratch1);
+  __ pushptr(here.addr(), noreg);
+#endif
   __ jump(RuntimeAddress(JeandleRuntimeRoutine::get_routine_entry(JeandleRuntimeRoutine::_deopt_blob)));
   assert(__ offset() - offset <= deopt_handler_size(), "deopt handler stub overflow");
   __ end_a_stub();
--- a/src/hotspot/cpu/x86/jeandleAssembler_x86.cpp
+++ b/src/hotspot/cpu/x86/jeandleAssembler_x86.cpp
@@ -185,7 +185,9 @@ int JeandleAssembler::interior_entry_alignment() const {
 
 int JeandleAssembler::emit_exception_handler() {
   address base = __ start_a_stub(NativeJump::instruction_size);
-  JEANDLE_ERROR_ASSERT_AND_RET_ON_FAIL(base != nullptr, "exception handler stub overflow", 0);
+  if (base == nullptr) {
+    JEANDLE_REPORT_ERROR_AND_RET("exception handler stub overflow", 0);
+  }
   int offset = __ offset();
   __ jump(RuntimeAddress(JeandleRuntimeRoutine::get_routine_entry(JeandleRuntimeRoutine::_exception_handler)));
   assert(__ offset() - offset <= (int)NativeJump::instruction_size, "overflow");
@@ -199,7 +201,9 @@ int JeandleAssembler::deopt_handler_size() {
 
 int JeandleAssembler::emit_deopt_handler() {
   address base = __ start_a_stub(deopt_handler_size());
-  JEANDLE_ERROR_ASSERT_AND_RET_ON_FAIL(base != nullptr, "deopt handler stub overflow", 0);
+  if (base == nullptr) {
+    JEANDLE_REPORT_ERROR_AND_RET("deopt handler stub overflow", 0);
+  }
   int offset = __ offset();
   InternalAddress here(__ pc());
   __ pushptr(here.addr(), rscratch1);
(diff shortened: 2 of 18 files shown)


taofengliu:
Pease fix the CI failure

taofengliu (commented):
There are some code just for debugging, please remove them.

eapenzhan:
## What is added

- Emit Jeandle deopt handler stubs and install them into `CodeOffsets::Deopt` and, when needed, `CodeOffsets::DeoptMH`.
- Reuse HotSpot `SharedRuntime::deopt_blob()` entry points instead of maintaining a Jeandle-specific deopt blob implementation.
- Propagate `is_method_handle_invoke` through Jeandle call-site metadata so the generated debug information matches HotSpot deoptimization expectations for MethodHandle call sites.
- Resolve and install the real `orig_pc_offset_in_bytes` from LLVM stackmaps instead of using a heuristic or hardcoded value.
- Allocate the Jeandle **OrigPcSlot** lazily, in the same entry-block static-alloca style as BasicLock, so it becomes a stable frame slot and does not interfere with entry CFG construction.

## Design notes

- **OrigPcSlot** is not created eagerly at the beginning of IR construction anymore. It is created on demand when a deopt bundle is formed, and inserted into the entry block before its terminator. This follows the same pattern already used by BasicLock.
- The installed `orig_pc_offset_in_bytes` is now the real stack-slot offset parsed from LLVM stackmaps.
- We only require a resolved orig-pc offset for methods that actually contain deopt payloads in stackmap records. Methods without such deopt records continue to install 0, avoiding unnecessary assertions during code installation.
- The final branch from the Jeandle entry initialization path is emitted from `entry_block()->tail_llvm_block()`, which preserves correctness when the entry path is extended by synchronized method locking or class-initialization guards.

## Tests fixed

- compiler/profiling/spectrapredefineclass_classloaders/Launcher.java
- compiler/c2/Test6910605_1.java
- compiler/uncommontrap/Test8009761.java
- compiler/interpreter/Test6833129.java
- compiler/c2/Test6910605_2.java
- compiler/escapeAnalysis/Test6775880.java
- compiler/jsr292/NonInlinedCall/InvokeTest.java
> compiler/profiling/spectrapredefineclass_classloaders/Launcher.java does not belong to depot handler problem any more.

src/hotspot/share/jeandle/jeandleCompiledCode.hpp
```diff
@@ -254,7 +275,11 @@ class JeandleCompiledCode : public StackObj {
   ciMethod* _method;
   address _routine_entry;
   std::string _func_name;
+  llvm::Value* _orig_pc_slot;
+  int _orig_pc_offset_in_bytes;
   int _interpreter_frame_size_in_bytes;
+  bool _needs_orig_pc_offset;
```
taofengliu: As I understand, this field can be deleted, right? We can just use `_orig_pc_offset_in_bytes` to determine if we need orig_pc_offset

src/hotspot/cpu/x86/jeandleAssembler_x86.cpp
```diff
@@ -185,14 +185,34 @@ int JeandleAssembler::interior_entry_alignment() const {
...
+}
+
+int JeandleAssembler::emit_deopt_handler() {
+  address base = __ start_a_stub(deopt_handler_size());
+  if (base == nullptr) {
+    JEANDLE_REPORT_ERROR_AND_RET("deopt handler stub overflow", 0);
+  }
+  int offset = __ offset();
+  InternalAddress here(__ pc());
+  __ pushptr(here.addr(), rscratch1);
```
taofengliu: It is different from the 64 arch version of `HandlerImpl::emit_deopt_handler` in `src/hotspot/cpu/x86/x86.ad`. Is it OK?

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -255,6 +255,11 @@ llvm::SmallVector<llvm::Value*> JeandleVMState::deopt_args(llvm::IRBuilder<>& bu
     args.push_back(obj.value());
     args.push_back(lock);
   }
+  if (llvm::Value* orig_pc_slot = JeandleCompilation::current()->compiled_code()->orig_pc_slot()) {
+    uint64_t encode = DeoptValueEncoding(0, DeoptValueEncoding::OrigPcSlotType, T_ADDRESS).encode();
+    args.push_back(builder.getInt64(encode));
+    args.push_back(orig_pc_slot);
+  }
```
taofengliu, line 262: Update the comment of depot args's format at the top of this function.

src/hotspot/share/jeandle/jeandleAssembler.hpp
```diff
@@ -47,6 +47,8 @@ class JeandleAssembler : public StackObj {
   void emit_clinit_barrier_on_entry(Klass* klass);
 
   int emit_exception_handler();
+  int emit_deopt_handler();
+  int deopt_handler_size();
```
taofengliu, line 52:
```suggestion
  int emit_exception_handler();

  int emit_deopt_handler();
  int deopt_handler_size();
```

src/hotspot/share/jeandle/jeandleRuntimeRoutine.hpp
```diff
@@ -284,7 +284,10 @@
 
 #define ALL_JEANDLE_ASSEMBLY_ROUTINES(def) \
   def(exceptional_return)                  \
-  def(exception_handler)
+  def(exception_handler)                   \
+  def(deopt_blob)                          \
+  def(deopt_blob_with_reexecution)         \
+  def(deopt_blob_with_exception)
```
taofengliu: Are they used?

src/hotspot/share/jeandle/jeandleCompiledCode.cpp
```diff
@@ -498,14 +504,19 @@ JeandleStackMap* JeandleCompiledCode::parse_stackmap(StackMapParser& stackmaps,
...
   if (num_deopts > 0) {
     assert(this->_method != nullptr, "must be method compilation");
 
     // bci goes first in deopt operands
     int bci = (location++)->getSmallConstant();
     num_deopts--;
-    call_info->set_bci(bci);
+    if (call_info != nullptr) {
+      call_info->set_bci(bci);
+    }
```
taofengliu: In fact, `call_info` cannot be nullptr.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
minchao1024, line 262: Since there's only one `original_pc_slot` in `nmethod`, passing it to all `deopt_args` would incur duplicate overhead. Perhaps we could use the intrinsic `llvm.experimental.stackmap` to implement this.

src/hotspot/cpu/x86/jeandleAssembler_x86.cpp
minchao1024: Don't use `LIR_Assembler::emit_deopt_handler` here, beacase it will destory `rscratch1` register that is an assignable register in C2。

eapenzhan: I double-check the reserved register in jeandle-llvm.  `rscratch1` is not reserved. I need to implement it as C2 in the x86.ad file.

eapenzhan: You're right, i'll change it.

src/hotspot/share/jeandle/jeandleRuntimeRoutine.hpp
eapenzhan: No, they're only used by C1 for now. As I was thinking to remove it, but I left it here to be reviewed to see your opinions. Both entries are used as their literal name. However I didn't find C2 use it any way. So I think it's not needed for jeandle either.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
eapenzhan: I think the orignal pc is for thread. The deopt method may be returned from different sub-call of different threads. So every call need a slot for its thread. It's not the same offset for different threads. What do you think?

src/hotspot/share/jeandle/jeandleCompiledCode.cpp
eapenzhan: Yeah, It's added by AI when I let it check for me. I'll remove it.

src/hotspot/share/jeandle/jeandleCompiledCode.hpp
eapenzhan: Yeah I'll change it to a function.

src/hotspot/share/jeandle/jeandleRuntimeRoutine.hpp
taofengliu: OK, I think we can remove it.

src/hotspot/cpu/aarch64/jeandleRegister_aarch64.hpp
```diff
@@ -49,7 +49,7 @@ class JeandleRegister : public AllStatic {
     r0, r1, r2, r3, r4, r5, r6, r7, r8, r9, r10,
     r11, r12, r13, r14, r15, r16, r17, r18_tls, r19, r20,
     r21, r22, r23, r24, r25, r26, r27, r28, r29, r30,
-    sp
+    r31_sp
```
minchao1024, line 52: It is recommended to keep using `sp` here because in the aarch64 DWARF standard, register 31 is strictly defined as the stack pointer, while `r31_sp` is defined as a physical register, representing both `sp` and `zr`.
https://github.com/ARM-software/abi-aa/blob/main/aadwarf64/aadwarf64.rst#41dwarf-register-names

eapenzhan: I understand. But when reviewing the code, I noticed that there is a conversion involving `VMReg`, and `VMRegImpl` does not define any registers beyond 31 on Aarch64. This `sp` at 33 is a virtual register — although its purpose is apparent on the surface, the conversion does not handle the mapping from virtual registers to physical registers. Therefore, I think it would be better to use 31 instead.

What do you think?

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
minchao1024: My mistake. To be more precise, each frame that may deopt has an original_pc_slot. The position of this slot is fixed relative to the current sp position, and the position only needs to be obtained once.

we can use the intrinsic `llvm.experimental.stackmap` to obtain it in every compiledCode.

src/hotspot/cpu/aarch64/jeandleRegister_aarch64.hpp
minchao1024: Yes，`sp` is not valid when call `as_VMReg`, using `r31_sp` is better.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
minchao1024: Unfortunately, `llvm.experimental.stackmap` has the side effect of forcibly preserving frame points (FPs).

We cannot use it because it conflicts with JVM's `PreserveFramePointer` flag。

eapenzhan: Yes, I think in subsequent optimizations we should develop a dedicated intrinsic for jeandle-llvm for this purpose. For now, let's proceed with this suboptimal solution as a temporary measure.

src/hotspot/share/jeandle/jeandleCompiledCode.cpp
```diff
@@ -680,6 +694,22 @@ int JeandleCompiledCode::frame_size_in_slots() {
...
+    _orig_pc_offset_in_bytes = offset;
+  } else {
+    assert(_orig_pc_offset_in_bytes == offset, "orig pc slot offset must be stable");
+  }
+}
+
+int JeandleCompiledCode::orig_pc_offset_in_bytes() const {
+  if (_orig_pc_offset_in_bytes < 0) {
+    return 0;
+  }
```
taofengliu: Why we return 0 when `_orig_pc_offset_in_bytes < 0` ?

eapenzhan: When it is less than 0, it indicates that the original pc slot has never been initialized(**normally the offset must be greater than 0**), meaning there is no deopt bundle, and this function should also have no possibility of deoptimization. This is simply assigning it a default value from before the deopt handler was implemented, which should be harmless for the time being.

src/hotspot/share/jeandle/jeandleCompiledCode.cpp
```diff
@@ -680,6 +694,22 @@ int JeandleCompiledCode::frame_size_in_slots() {
...
+  return _orig_pc_offset_in_bytes;
+}
```
taofengliu: We can use 0 to represent an uninitialized value for _orig_pc_offset_in_bytes, then we can return immediately here.
Then move this function to jeandleCompiledCode.hpp

src/hotspot/share/jeandle/jeandleCompiledCode.cpp
```diff
@@ -680,6 +694,22 @@ int JeandleCompiledCode::frame_size_in_slots() {
   return _frame_size * sizeof(intptr_t) / VMRegImpl::stack_slot_size;
 }
 
+void JeandleCompiledCode::set_real_orig_pc_offset_in_bytes(int offset) {
+  assert(offset >= 0, "sanity");
+  if (_orig_pc_offset_in_bytes == -1) {
+    _orig_pc_offset_in_bytes = offset;
+  } else {
```
taofengliu, line 701:
```suggestion
  assert(offset > 0, "sanity");
  if (_orig_pc_offset_in_bytes == 0) {
    _orig_pc_offset_in_bytes = offset;
  } else {
```

eapenzhan:
The original pc slot behaves differently in terms of its slot position on the x86 and aarch64 platforms. On x86, it may coincide with the address of unextended_sp, whereas on aarch64 it will not. We need to investigate in detail whether this difference will introduce any issues; follow-up research will be tracked by this issue #422 .

taofengliu (approved):

taofengliu:
@mergifyio queue

Issue #401, opened 2026-03-12
JRE crash due to missing template.ll in java.base module


minchao1024:
**Describe the bug**
The Jeandle LLVM backend requires `template.ll` at runtime for code generation, but this file was missing from the modular JRE image.

**Minimal reproduce steps**
```
make legacy-jre-image CONF=fastdebug
java -Xcomp -XX:-TieredCompilation -verison
```

PR #402, opened 2026-03-12
fix: Include template.ll in java.base jmod

minchao1024:
## Related issue(s):

issue #401 

This fixes the crash occurring when running applications on a 
custom-built JRE without the full JDK environment.


taofengliu:
@mergifyio queue

Issue #386, opened 2026-03-06
The line number is parsed incorrectly in HeapMonitor


xiejf2020:
**Describe the bug**

In test/hotspot/jtreg/serviceability/jvmti/HeapMonitor/MyPackage/HeapMonitorTest.java and other HeapMonitorXxxTest.java:

```
java.lang.RuntimeException: No expected events were found: false, false
        at MyPackage.HeapMonitor.allocateAndCheckFrames(HeapMonitor.java:220)
        at MyPackage.HeapMonitor.allocateAndCheckFrames(HeapMonitor.java:234)
        at MyPackage.HeapMonitorTest.main(HeapMonitorTest.java:46)
        at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
        at java.base/java.lang.reflect.Method.invoke(Method.java:580)
        at com.sun.javatest.regtest.agent.MainWrapper$MainTask.run(MainWrapper.java:138)
        at java.base/java.lang.Thread.run(Thread.java:1583)
```

Enable the print logs in libHeapMonitorTest.cpp and it shows:

```cpp
    if (print_out_comparisons) {
      fprintf(stderr, "\tComparing: (check_lines: %d)\n", check_lines);
      fprintf(stderr, "\t\tNames: %s and %s\n", name, expected[i].name);
      fprintf(stderr, "\t\tSignatures: %s and %s\n", signature, expected[i].signature);
      fprintf(stderr, "\t\tFile name: %s and %s\n", file_name, expected[i].file_name);
      fprintf(stderr, "\t\tLines: %d and %d\n", line_number, expected[i].line_number);
      fprintf(stderr, "\t\tResult is %d\n", differ);
    }
```
↓
```
        Comparing: (check_lines: 1)
                Names: actuallyAllocate and actuallyAllocate
                Signatures: ()I and ()I
                File name: HeapMonitor.java and HeapMonitor.java
                Lines: 87 and 98
                Result is 1
```

The line numbers are different.

**Minimal reproduce steps**

```bash
make test CONF=fast JTREG="VM_OPTIONS='-XX:+UseJeandleCompiler'" TEST=test/hotspot/jtreg/serviceability/jvmti/HeapMonitor/MyPackage/HeapMonitorTest.java
```

**Expected behavior**

The line number should be 98 instead of 87.

xiejf2020:
[...]

PR #406, opened 2026-03-13
fix: bytecode 'new' missing BCI info

xiejf2020:
## Related issue(s):

issue #386

Add deopt bundle for `newinstance`, `newarray` and `multianewarray`.
Therefore, the compiled code can correctly obtain the current BCI.


taofengliu:
@mergifyio queue

Issue #395, opened 2026-03-11
finalize() won't be executed by `System.gc()`


rcjhd:
**Describe the bug**
`Object.finalize()` function won't be executed by `System.gc()`.

**Minimal reproduce steps**
I write a simple testcase to quickly reproduce that
```
public class FinalizeDemo {
    private static boolean isFinalized = false;

    @Override
    protected void finalize() throws Throwable {
        isFinalized = true;
        System.out.println("finalize() is executed！");
    }

    public static void testGCTrigger() {
        isFinalized = false;

        new FinalizeDemo();

        System.gc();

        try {
            Thread.sleep(100);
        } catch (InterruptedException e) {
            Thread.currentThread().interrupt();
        }

        System.out.println("isFinalized：" + isFinalized);
    }

    public static void main(String[] args) {
        testGCTrigger();
    }
}

```
and then `java -Xcomp -XX:-TieredCompilation FinalizeDemo`

**Expected behavior**
```
finalize() is executed！
isFinalized：true
```
But now:
```
isFinalized: false
```
**Additional context**
testcase `java/lang/Object/FinalizationOption.java`、 `java/lang/ref/Basic.java` and maybe more due to this reason.

rcjhd:
Firstly, when I add `-XX:"CompileCommand=compileonly,java.lang.Object::<init>"` for Jeandle compiler, there will be having this problem. On the contrary, the problem disappears when I exclude that function.

It seems like a reference processing lost in GC. When enabled `-Xlog:gc+ref=debug`, it shows a difference in above case.
Expected:
```
[1.366s][debug][gc,ref] GC(0) ReferenceProcessor::execute queues: 1, RefProcThreadModel::Single, marks_oops_alive: false
[1.366s][debug][gc,ref] GC(0) ReferenceProcessor::execute queues: 1, RefProcThreadModel::Single, marks_oops_alive: true
[1.366s][debug][gc,ref] GC(0) Skipped PhantomRefsPhase of Reference Processing: no references
```
Actual:
```
[1.367s][debug][gc,ref] GC(0) Skipped SoftWeakFinalRefsPhase of Reference Processing: no references
[1.367s][debug][gc,ref] GC(0) Skipped KeepAliveFinalRefsPhase of Reference Processing: no references
[1.367s][debug][gc,ref] GC(0) Skipped PhantomRefsPhase of Reference Processing: no references
```
[...]

PR #408, opened 2026-03-16
fix: call_register_finalizer in Object::<init>

rcjhd:
## Related issue(s):

issue #395 

C1/C2 compiler will register finalizer in `Object::<init>`, while Jeandle will not. It causes the finalize() function can't be normally executed. This PR is to add related logic in return_current like C2.


taofengliu:
@mergifyio queue

Issue #378, opened 2026-03-03
Support On-Stack Replacement


minchao1024:
**The feature you'd like**
Implement OSR mechanism to allow the VM to transition from interpreted code to compiled code during a long-running loop.

PR #415, opened 2026-03-27
feat: Support On-Stack Replacement

minchao1024:
## Related issue(s):

issue #378 

This PR implements OSR compilation for Jeandle-JDK and introduces several 
optimizations for monitor management:

1. Support OSR: Implemented the entry logic and state migration for OSR.
2. Monitor PhiNodes: Constructed PhiNodes for monitors during OSR to handle 
   monitor states across loop boundaries, which is unique to OSR compilation.
3. LockRecord Reuse: Optimized stack space by reusing LockRecords based on 
   monitor nesting levels instead of allocating per enter/exit pair.


src/hotspot/share/jeandle/jeandleCompilation.cpp
```diff
@@ -276,6 +274,8 @@ void JeandleCompilation::compile_java_method() {
     JeandleAbstractInterpreter interpret(_method, _entry_bci, *_llvm_module, _code);
   }
 
+  RETURN_VOID_ON_JEANDLE_ERROR();
+
```
taofengliu: Can we return after dump ir? IR is always necessary for debug.

src/hotspot/share/jeandle/jeandleUtils.cpp
```diff
@@ -30,18 +30,23 @@
 #include "compiler/compilerThread.hpp"
 #include "runtime/thread.hpp"
 
-llvm::Function* JeandleFuncSig::create_llvm_func(ciMethod* method, llvm::Module& target_module) {
+llvm::Function* JeandleFuncSig::create_llvm_func(ciMethod* method, llvm::Module& target_module, bool is_osr) {
```
taofengliu:
```suggestion
llvm::Function* JeandleFuncSig::create_llvm_func(ciMethod* method, llvm::Module& target_module, bool is_osr_entry) {
```

src/hotspot/share/jeandle/jeandleUtils.hpp
```diff
@@ -32,7 +32,7 @@
 class JeandleFuncSig : public AllStatic {
  public:
   // Create a llvm function according to the Java method.
-  static llvm::Function* create_llvm_func(ciMethod* method, llvm::Module& target_module);
+  static llvm::Function* create_llvm_func(ciMethod* method, llvm::Module& target_module, bool is_osr);
```
taofengliu:
```suggestion
  static llvm::Function* create_llvm_func(ciMethod* method, llvm::Module& target_module, bool is_osr_entry);
```

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -453,7 +475,13 @@ void BasicBlockBuilder::setup_control_flow() {
   // Connect all basic blocks according to control flow transfer instructions.
   ciBytecodeStream codes(_method);
 
-  JeandleBasicBlock* current = _entry_block;
+  if (!is_osr()) {
+    connect_block(_bci2block[0], entry_block());
+  } else {
+    connect_block(_bci2block[_entry_bci], osr_entry_block());
```
taofengliu: Is it possible that osr bci is not the first bci of a basic block? According to C1's BlockListBuilder, we may need to explicitly create a separate block for osr entry. The Jeandles BasicBlockBuilder is similar to BlockListBuilder.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -573,7 +601,11 @@ void BasicBlockBuilder::mark_loops() {
   _visited.initialize(num_blocks);
   _next_block_order = num_blocks - 1;
 
-  mark_loops(_bci2block[0]);
+  if(!is_osr()) {
+    mark_loops(_bci2block[0]);
+  } else {
+    mark_loops(_bci2block[_entry_bci]);
+  }
```
taofengliu, line 613: Is it needed? BlockListBuilder does not do this.
By the way, we can refer to BlockListBuilder to adjust our code.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -648,80 +680,275 @@ void JeandleAbstractInterpreter::initialize_VM_state() {
...
+    initialize_VM_state_from_osr_buffer(initial_jvm, _llvm_func->getArg(0));
   }
 
   _block_builder->entry_block()->set_VM_state(initial_jvm);
 }
 
+void JeandleAbstractInterpreter::initialize_VM_state_from_osr_buffer(JeandleVMState* initial_jvm, llvm::Value* osr_buffer) {
+  assert(is_osr(), "sanity");
+
+  // Do some special top-level things.
```
```suggestion
  // Do type flow analysis.
```

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -648,80 +680,275 @@ void JeandleAbstractInterpreter::initialize_VM_state() {
...
+  assert(_method != nullptr, "only for Java method compilations");
+  ciTypeFlow* flow = _method->get_osr_flow_analysis(_entry_bci);
+  assert(!flow->failing(), "type flow analysis failed for OSR compilation");
```
taofengliu, line 721: In debug builds, we can crash because assertion failure. In release builds, we should record method_not_compilable.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -648,80 +680,275 @@ void JeandleAbstractInterpreter::initialize_VM_state() {
...
+
+  int max_locals = initial_jvm->max_locals();
+  ciTypeFlow::Block* osr_entry_block = flow->rpo_at(0);
+  assert(osr_entry_block->start() == _entry_bci, "the first rpo block must be osr entry block");
+
+  // Check bailouts.  We currently do not perform on stack replacement
+  // of loops in catch blocks or loops which branch with a non-empty stack.
```
taofengliu: Hard to understand. Explain it more.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -648,80 +680,275 @@ void JeandleAbstractInterpreter::initialize_VM_state() {
...
+  if (osr_entry_block->stack_size() != 0) {
+    JEANDLE_REPORT_ERROR_AND_RET_VOID("OSR starts with non-empty stack");
+  }
+
+  // Commute monitors from interpreter frame to compiler frame.
+  int mcnt = osr_entry_block->monitor_count();
```
```suggestion
  int monitor_count = osr_entry_block->monitor_count();
```
C2's var name is hard to read.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -648,80 +680,275 @@ void JeandleAbstractInterpreter::initialize_VM_state() {
...
+                                                                      osr_buffer,
+                                                                      _ir_builder.getInt64(monitors_addr_offset - (index * 2 + 1) * wordSize));
+      llvm::Value* displaced_hdr = load_from_address(displaced_hdr_addr, T_ADDRESS, false);
+
+      llvm::Value* lock = entry_block_ir_builder.CreateAlloca(_ir_builder.getIntPtrTy(_module.getDataLayout()),
+                                                              llvm::jeandle::AddrSpace::CHeapAddrSpace,
+                                                              nullptr,
+                                                              "BasicLock");
+      push_allocated_basic_lock(lock);
+      assert(allocated_basic_lock_at(initial_jvm->locks_size()) == lock, "unbalanced monitors");
```
taofengliu, line 762: How can this make sure balanced monitors?

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -648,80 +680,275 @@ void JeandleAbstractInterpreter::initialize_VM_state() {
...
+
+    if (!live_oops->at(index)) {
+      continue;
+    }
+
+    if (osr_entry_trap_block == nullptr) {
+      osr_entry_trap_block = llvm::BasicBlock::Create(*_context, "osr_entry_trap_block", _llvm_func);
+      _bytecodes.force_bci(_entry_bci);
+      uncommon_trap(Deoptimization::Reason_constraint, Deoptimization::Action_reinterpret, osr_entry_trap_block);
+    }
```
taofengliu, line 882: Better to move these out of the loop.


```suggestion
  if (_jvm->max_locals() > 0) {
    // Initialize osr_entry_trap_block
  }
```

src/hotspot/share/jeandle/jeandleAbstractInterpreter.hpp
```diff
@@ -288,7 +293,30 @@ class JeandleAbstractInterpreter : public StackObj {
   // Object & Lock for synchronized method
   LockValue _sync_lock;
 
+  // Reuse stack allocation for monitor
```
taofengliu:
```suggestion
  // All allocated basic lock on stack.
```

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -2637,10 +2864,20 @@ void JeandleAbstractInterpreter::shared_lock(LockValue lock) {
...
+      // Alloca insts should be in the entry block to be 'StaticAlloca'. Then they could be folded into prologue code.
+      llvm::IRBuilder entry_block_ir_builder(_block_builder->entry_block()->header_llvm_block()->getTerminator());
+      basic_lock = entry_block_ir_builder.CreateAlloca(_ir_builder.getIntPtrTy(_module.getDataLayout()),
+                                                       llvm::jeandle::AddrSpace::CHeapAddrSpace, nullptr, "BasicLock");
+      // Save the basic_lock for later reuse.
+      push_allocated_basic_lock(basic_lock);
+      assert(allocated_basic_lock_at(monitor_nest_level) == basic_lock, "unbalanced monitors");
+    } else {
+      basic_lock = allocated_basic_lock_at(monitor_nest_level);
+    }
```
taofengliu, line 3082: I am wondering will this path taken when running? Because the initial vm state has contained the allocated basic locks, thus all successors' vm state will have the allocated basic locks.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -110,14 +111,27 @@ bool JeandleVMState::match(JeandleVMState* to_match) {
...
+    for (size_t i = 0; i < income_jvm->locks_size(); i++) {
+      assert(!income_jvm->lock_at(i).is_null(), "null lock");
+      assert(!lock_at(i).is_null(), "null lock");
+      assert(lock_at(i).lock() == income_jvm->lock_at(i).lock(), "unbalanced monitors");
+
+      llvm::PHINode* phi_node = llvm::cast<llvm::PHINode>(lock_at(i).object().value());
+
+      phi_node->addIncoming(income_jvm->lock_at(i).object().value(), income_block);
+    }
+  }
```
taofengliu, line 138: Why osr compilation need to create phi nodes for locks?

src/hotspot/share/jeandle/jeandleType.hpp
```diff
@@ -119,7 +121,7 @@ class LockValue {
   LockValue() : _object(TypedValue()), _basic_lock(nullptr) { }
 
   bool equals(const LockValue& rhs) {
-    return _object.value() == rhs._object.value() && _basic_lock == rhs._basic_lock;
+    return _object.value()->getType() == rhs._object.value()->getType() && _basic_lock == rhs._basic_lock;
```
taofengliu: Still cannot understand why, could you explain more on how locks are handled?

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
minchao1024: The OSR BCI is always the target of a backedge, so it has to be at the start of a block. And we also have `osr_entry_block()` for osr setup code.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
minchao1024: During OSR compilation, marking loops starting from the OSR BCI helps us locate the true loop headers, thereby reducing the number of Phi nodes.
```
while (FlagA) { // This is the loop header for normal compilation, but not for OSR compilation.
    while (FlagB) { // osr entry
        doSomething();
    }
}
```

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
minchao1024: Code comments updated.
```
  // OSR Compilation Bailouts:
  // In HotSpot, OSR is restricted to loop headers where the operand stack is empty.
  // This is because SharedRuntime::OSR_migration_begin is designed to migrate 
  // only locals and monitors from the interpreter frame; it does not currently account for 
  // copying operand stack slots into the OSR buffer.
  if (osr_entry_block->stack_size() != 0) {
    JEANDLE_REPORT_ERROR_AND_RET_VOID("OSR starts with non-empty stack");
  }
```

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
minchao1024: This assertion alone does not guarantee balanced monitors; it is only a necessary condition for internal state consistency.
The logic here reuses `lock_record` based on the `nest_level`. This check simply verifies that the `lock_record` correctly maps to the current nesting level.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
minchao1024: We only need `osr_entry_trap_block` after it's determined that the `check_locals` logic needs to be generated. Since `max_local > 0` doesn't always trigger the `check_locals` logic, we still need to handle this situation inside the loop to avoid unnecessary blocks.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.hpp
minchao1024: This comment indicates that the `basic_block` will be reused, which I think is fine.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
minchao1024: This is the original problem: we allocated a `lock_record` on the stack for each `monitorenter`, but this can cause redundancy in some cases, such as the following:

```
synchronized(this) {       // monitor_nest_level = 0
    synchronized(objA) {   // monitor_nest_level = 1
        doSomething();
    }
    synchronized(objB) {   // monitor_nest_level = 1
        doSomething();
    }
}
```
Since the monitors for `objA` and `objB` operate at the same `monitor_nest_level` and have disjoint lifecycles, they can share the same `lock_record` slot.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
minchao1024: Because the locks of the `osr_entry_bci` block may come from different blocks and have different values in OSR compilation, they need to be merged by Phi nodes. For instance:
```
while (flagA) {
    synchronized(objA) {   // There is a lock obj value (lock_objA) created by JeandleAbstractInterpreter in blockA.
        while (flagB) {    // [osr entry] There is a lock obj value (lock_objB) created by osr_migration in osr_entry_block.
            doSomething(); // %lock_obj = phi [%lock_objA, %blockA], [%lock_objB, %osr_entry_block]
        }
    }
}
```

src/hotspot/share/jeandle/jeandleType.hpp
minchao1024: `_object.value()` may come from different blocks and have different values in OSR compilation

src/hotspot/share/jeandle/jeandleCompilation.cpp
minchao1024: It looks we can delay the return operation.

src/hotspot/share/jeandle/jeandleUtils.cpp
minchao1024: Done

src/hotspot/share/jeandle/jeandleUtils.hpp
minchao1024: Done

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
minchao1024: Fixed

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
minchao1024: Fixed

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
minchao1024: Fixed

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
taofengliu: `osr_entry_block()` does not check the osr bci always starts a basic block. Can we make an assertion here?
And `osr_entry_block()` is same as `entry_block()`, can we just use `entry_block` ?

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -648,80 +680,297 @@ void JeandleAbstractInterpreter::initialize_VM_state() {
...
+      llvm::Value* lock = entry_block_ir_builder.CreateAlloca(_ir_builder.getIntPtrTy(_module.getDataLayout()),
+                                                              llvm::jeandle::AddrSpace::CHeapAddrSpace,
+                                                              nullptr,
+                                                              "BasicLock");
+      push_allocated_basic_lock(lock);
+      assert(allocated_basic_lock_at(initial_jvm->locks_size()) == lock, "unbalanced monitors");
+      store_to_address(lock, displaced_hdr, T_ADDRESS, false);
+
+      if (index == 0 && _method->is_synchronized()) {
+        JeandleCompilation::current()->set_has_monitors(true);
```
taofengliu: Do we need to set_has_monitors not only for sync_lock? Is it better to set_has_monitors immediately when `if (monitor_count != 0)` ?

src/hotspot/share/jeandle/jeandleType.hpp
taofengliu: But this check is too fragile here. Let's think how to improve.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -648,80 +680,297 @@ void JeandleAbstractInterpreter::initialize_VM_state() {
...
+  }
+
+  assert(!skip_next_slot, "broken double word");
+
+  // Release osr buffer
+  llvm::FunctionCallee OSR_migration_end_callee = JeandleRuntimeRoutine::SharedRuntime_OSR_migration_end_callee(_module);
+  llvm::CallInst* call_OSR_migration_end = _ir_builder.CreateCall(OSR_migration_end_callee, {osr_buffer});
+  call_OSR_migration_end->setCallingConv(llvm::CallingConv::C);
+
+  // Initilaize vm_state for incoming uncommon_trap
```
```suggestion
  // Initialize vm_state for incoming uncommon_trap
```

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
taofengliu: We can add a comment to explain this.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
minchao1024: There is an assertion in `connect_block`, which can ensure the `_entry_bci` is a `start_bci` of basic block.
```
  static void connect_block(JeandleBasicBlock* child_block, JeandleBasicBlock* parent_block) {
    assert(child_block != nullptr && parent_block != nullptr, "connecting nullptr");
    child_block->add_predecessor(parent_block);
    parent_block->add_successor(child_block);
  }
```

And `osr_entry_block()` removed.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
minchao1024: Comment added.

src/hotspot/share/jeandle/jeandleType.hpp
minchao1024: Since the monitor phi node only exists in the OSR compilation, it has been modified to retain the original logic in the normal compilation.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
minchao1024: Fixed

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
minchao1024: Fixed

src/hotspot/share/jeandle/templatemodule/template.ll
```diff
@@ -196,6 +196,21 @@ check_subtype:
...
+entry:
+  %is_null = icmp eq ptr addrspace(1) %oop, null
+  br i1 %is_null, label %return_true, label %check_subtype
+
+return_true:
+  ret i1 true
+
+check_subtype:
+  %sub_klass = call hotspotcc ptr addrspace(0) @jeandle.load_klass(ptr addrspace(1) %oop)
+  %is_subtype = call hotspotcc i1 @jeandle.check_klass_subtype(ptr addrspace(0) %sub_klass, ptr addrspace(0) %super_klass)
```
taofengliu: We can use `jeandle.check_instanceof`, which can be eliminated by jeandle-llvm. And use nonnull attribute.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -651,80 +685,298 @@ void JeandleAbstractInterpreter::initialize_VM_state() {
...
-    locals_idx += type->size();
+    initialize_VM_state_from_osr_buffer(initial_jvm, _llvm_func->getArg(0));
   }
 
   _block_builder->entry_block()->set_VM_state(initial_jvm);
 }
 
-void JeandleAbstractInterpreter::interpret() {
-  JeandleBasicBlock* current = bci2block()[0];
+void JeandleAbstractInterpreter::initialize_VM_state_from_osr_buffer(JeandleVMState* initial_jvm, llvm::Value* osr_buffer) {
```
taofengliu, line 715: Is it possible to attach type information to the loaded locals and locks' objects?

src/hotspot/share/jeandle/jeandleType.hpp
```diff
@@ -119,6 +122,9 @@ class LockValue {
   LockValue() : _object(TypedValue()), _basic_lock(nullptr) { }
 
   bool equals(const LockValue& rhs) {
+    if (JeandleCompilation::current()->is_osr_compilation()) {
+      return _object.value()->getType() == rhs._object.value()->getType() && _basic_lock == rhs._basic_lock;
+    }
```
taofengliu, line 130: Add a comment to explain why we treat OSR differently.

src/hotspot/share/jeandle/templatemodule/template.ll
minchao1024: Fixed

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
minchao1024: Oops are verified via `instanceof_or_null`, which automatically attaches type information whenever possible.

src/hotspot/share/jeandle/jeandleType.hpp
minchao1024: Done

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
taofengliu: But these can only be valid through context-sensitive query. We can attach metadata to load instructions to make them accessible through context-insensitive query.

minchao1024: `instanceof_or_null` provides type sharpening only within specific control-flow branches, it cannot resolve types for context-insensitive queries due to the potential null case.

src/hotspot/share/jeandle/templatemodule/template.ll
```diff
@@ -196,6 +196,20 @@ check_subtype:
   ret i32 %is_subtype_ext
 }
 
+; Implementation of Java instanceof_or_null operation.
+define hotspotcc i1 @jeandle.instanceof_or_null(ptr addrspace(0) nocapture %super_klass, ptr addrspace(1) nocapture %oop) noinline "lower-phase"="0" {
```
taofengliu: Can we use `jeandle.checkcast`?

src/hotspot/share/jeandle/jeandleAbstractInterpreter.hpp
```diff
@@ -288,7 +292,30 @@ class JeandleAbstractInterpreter : public StackObj {
...
+  llvm::SmallVector<llvm::Value*> _allocated_basic_lock;
```
taofengliu, line 300: Add a comment to explain why we can reuse basic locks

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -651,80 +685,298 @@ void JeandleAbstractInterpreter::initialize_VM_state() {
...
+    }
+
+    if (!live_oops->at(index)) {
+      continue;
+    }
+
+    if (osr_entry_trap_block == nullptr) {
+      osr_entry_trap_block = llvm::BasicBlock::Create(*_context, "osr_entry_trap_block", _llvm_func);
+      _bytecodes.force_bci(_entry_bci);
+      uncommon_trap(Deoptimization::Reason_constraint, Deoptimization::Action_reinterpret, osr_entry_trap_block);
```
taofengliu, line 881: Will this happen in a real Java program?

taofengliu (commented):
We can use AI to add more comprehensive tests for OSR, including synchronized methods and monitors and so on.

src/hotspot/share/jeandle/templatemodule/template.ll
minchao1024: Yes, they are the same implementation; I will remove the redundant `instanceof_or_null`.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.hpp
minchao1024: Done

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
minchao1024: Yes, the corresponding test case `testOSRTypeState` have been added.

taofengliu (commented):
We can test synchronization by having multiple threads contend for the same lock.

taofengliu (approved):

taofengliu:
@mergifyio queue

Issue #425, opened 2026-04-11
Jeandle JIT TargetMachine initialization does not propagate host CPU / host feature information


eapenzhan:
Jeandle currently creates its LLVM `TargetMachine` using the process triple only, but without propagating the host CPU name or the host CPU feature set.

In `src/hotspot/share/jeandle/jeandleCompiler.cpp`, `JeandleCompiler::initialize_target_machine()` currently does:

```cpp
llvm::Triple target_triple = llvm::Triple(llvm::sys::getProcessTriple());

llvm::TargetOptions options;
llvm::SubtargetFeatures features;
options.EmitStackSizeSection = true;

_target_machine = target->createTargetMachine(target_triple,
                                              "" /* CPU */,
                                              features.getString(),
                                              options,
                                              llvm::Reloc::Model::PIC_,
                                              llvm::CodeModel::Model::Small,
                                              llvm::CodeGenOptLevel::Aggressive,
                                              true /* JIT */);
```

Here, both:
- `CPU` is empty
- `features` is an empty `llvm::SubtargetFeatures`

This means Jeandle is currently target-triple-aware, but not host-feature-aware, when constructing the JIT `TargetMachine`.

**Why this matters**

On host JIT compilation, LLVM instruction selection depends not only on the target triple, but also on the subtarget CPU/features attached to the `TargetMachine`. Without explicitly passing host CPU / host feature information, LLVM may select a conservative generic subtarget instead of using the actual capabilities of the running machine.

This is particularly visible on x86. For example, lowering paths for operations such as `llvm.floor`, `llvm.ceil`, and `llvm.rint` depend on whether LLVM sees `SSE4.1` as available. If Jeandle does not provide host CPU/features, these operations may not be legalized and selected through the expected `ROUNDSD`-based path, even on machines that do support SSE4.1.

Note that `llvm::TargetParser::SubtargetFeatures` itself does not automatically discover host CPU capabilities. It is only a feature string container. If Jeandle wants host-accurate JIT code generation, it needs to explicitly query and pass host information when creating the `TargetMachine`.

**Expected behavior**

When Jeandle creates the LLVM JIT `TargetMachine`, it should propagate:
- host CPU name
- host CPU features

so that LLVM can generate code against the actual host subtarget instead of a conservative default.

**Proposed fix**

During `TargetMachine` construction, Jeandle should use LLVM host queries such as:
- `llvm::sys::getHostCPUName()`
[...]

PR #427, opened 2026-04-13
fix: TargetMachine initialization does not propagate host CPU / host feature information

eapenzhan:
## Related issue(s): 

issue #425 

This would allow Jeandle to construct a host-accurate JIT TargetMachine, improving code generation quality and enabling expected target-specific instruction selection for host-supported features.

### DIFF
--- a/src/hotspot/cpu/aarch64/jeandleUtils_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/jeandleUtils_aarch64.cpp
@@ -21,8 +21,24 @@
 #include "jeandle/__llvmHeadersBegin__.hpp"
 #include "llvm/IR/Jeandle/Attributes.h"
 #include "llvm/IR/Jeandle/GCStrategy.h"
+#include "llvm/TargetParser/SubtargetFeature.h"
 
 #include "jeandle/jeandleUtils.hpp"
+
+#include "jeandle/__hotspotHeadersBegin__.hpp"
+#include "runtime/arguments.hpp"
+
+void apply_vm_flag_feature_overrides(llvm::SubtargetFeatures& features) {
+  if (!UseLSE) {
+    features.AddFeature("lse", false);
+  }
+  if (UseSVE < 2) {
+    features.AddFeature("sve2", false);
+  }
+  if (UseSVE < 1) {
+    features.AddFeature("sve", false);
+  }
+}
 
 void JeandleFuncSig::setup_description(llvm::Function* func, bool is_stub) {
   func->setCallingConv(llvm::CallingConv::Hotspot_JIT);
--- a/src/hotspot/cpu/aarch64/jeandleUtils_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/jeandleUtils_aarch64.cpp
@@ -29,6 +29,19 @@
 #include "runtime/arguments.hpp"
 
 void apply_vm_flag_feature_overrides(llvm::SubtargetFeatures& features) {
+  if (!UseAES) {
+    features.AddFeature("aes", false);
+  }
+  if (!UseSHA) {
+    features.AddFeature("sha2", false);
+    features.AddFeature("sha3", false);
+  }
+  if (!UseNeon) {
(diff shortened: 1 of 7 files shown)


minchao1024:
We might need to introduce support for CPU-specific VM flags (e.g., `UseAVX`, `UseSSE`, `UseLSE`).

src/hotspot/share/jeandle/jeandleCompiler.cpp
```diff
@@ -53,6 +53,62 @@ void install_jeandle_llvm_fatal_error_handler() {
   }
 }
 
+void apply_vm_flag_feature_overrides(llvm::SubtargetFeatures& features) {
```
taofengliu: I suggest move it to a target specific source file.

eapenzhan: Agree. I can move the target-specific VM flag to LLVM feature mapping into existing jeandleUtils_<arch>.cpp files, so jeandleCompiler.cpp stays target-agnostic without introducing a brand new compiler source split yet.

src/hotspot/cpu/aarch64/jeandleUtils_aarch64.cpp
```diff
@@ -21,9 +21,25 @@
...
+  if (!UseLSE) {
+    features.AddFeature("lse", false);
+  }
+  if (UseSVE < 2) {
+    features.AddFeature("sve2", false);
+  }
+  if (UseSVE < 1) {
+    features.AddFeature("sve", false);
+  }
+}
```
minchao1024, line 54: There are more vm options here that need to be implemented, like `UseNeon`, `UseCRC32`. You can locate them by searching for `instructions` within `src/hotspot/share/runtime/globals.hpp` and `CPU_HEADER(globals)`.

eapenzhan: OK, let me check.

src/hotspot/share/jeandle/jeandleUtils.cpp
```diff
@@ -73,6 +74,12 @@ std::string JeandleFuncSig::method_name_with_signature(ciMethod* method) {
   return method_name(method) + signature;
 }
 
+#if !defined(AMD64) && !defined(AARCH64)
+void apply_vm_flag_feature_overrides(llvm::SubtargetFeatures& features) {
+  (void)features;
+}
+#endif
```
taofengliu: We can define it for RISCV, using `Unimplemented()`. Then remove the definition in this file.

eapenzhan: I'll remove it. And then I'll try to implement the RISCV feature.

eapenzhan:
Move VM flag feature overrides into target-specific utils

- move apply_vm_flag_feature_overrides out of shared jeandleCompiler.cpp into target-specific jeandleUtils_<arch>.cpp
- extend AArch64 feature masking to honor UseAES, UseSHA, UseNeon, UseCRC32, UseLSE, and UseSVE
- extend x86 feature masking to honor UseAES, UseCLMUL, UseFMA, UseBMI1Instructions, UseBMI2Instructions, UsePopCountInstruction, UseCountLeadingZerosInstruction, UseSSE, UseAVX, and UseSHA
- add RISCV VM flag -> LLVM feature masking for UseRVC, UseRVV, UseZba, UseZbb, UseZbs, UseZic64b, UseZicbom, UseZicbop, UseZicboz, and UseZihintpause
- for RISCV, stop attaching a separate function-level target-features attr and rely on the target machine feature set instead, so there is a single source of truth for effective LLVM features

taofengliu (approved):

taofengliu:
@eapenzhan Please check the CI failure

eapenzhan:
> @eapenzhan Please check the CI failure

Issue [#431](https://github.com/jeandle/jeandle-jdk/issues/431) is the reason of All RISCV build failure. We need to merge the #432 first.

taofengliu:
@mergifyio queue

Issue #370, opened 2026-02-26
Express Java class in Jeandle compiled IR


taofengliu:
**The feature you'd like**
Express Java class in Jeandle compiled IR. This enable Jeandle's optimizer to refer to objects's type to do more
powerful optimization.

PR #430, opened 2026-04-13
feat: Add type information for Jeandle compiled code

taofengliu:
## Related issue(s):

https://github.com/jeandle/jeandle-jdk/issues/370

### DIFF
--- a/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
+++ b/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
@@ -23,6 +23,7 @@
 #include "llvm/IR/Attributes.h"
 #include "llvm/IR/Jeandle/Attributes.h"
 #include "llvm/IR/Jeandle/GCStrategy.h"
+#include "llvm/IR/Jeandle/JavaType.h"
 #include "llvm/IR/Jeandle/Metadata.h"
 
 
@@ -36,6 +37,7 @@
 #include "ci/ciMethodBlocks.hpp"
 #include "ci/ciObjArrayKlass.hpp"
 #include "ci/ciSymbols.hpp"
+#include "oops/objArrayKlass.hpp"
 #include "classfile/javaClasses.hpp"
 #include "interpreter/interpreter.hpp"
 #include "logging/log.hpp"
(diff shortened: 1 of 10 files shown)


src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -2029,6 +2028,23 @@ void JeandleAbstractInterpreter::do_get_xxx(ciField* field, bool is_static) {
...
+  // so a field declared as an interface could hold any Object at runtime.
+  if (field->type()->is_klass()) {
+    ciKlass* field_klass = field->type()->as_klass();
+    if (field_klass->is_loaded() && !field_klass->is_interface()) {
+      Klass* klass_enc = (Klass*)(field_klass->constant_encoding());
+      if (llvm::Instruction* load_inst = llvm::dyn_cast<llvm::Instruction>(value)) {
+        llvm::MDNode* klass_md = llvm::MDNode::get(*_context, {
+            llvm::ConstantAsMetadata::get(_ir_builder.getInt64((intptr_t)klass_enc))
+        });
+        load_inst->setMetadata(llvm::jeandle::Metadata::JavaKlass, klass_md);
```
minchao1024: It appears that `field_klass` and `ret_klass` could also be considered as `Exact` when they are `final`. However, since we can't directly apply attributes to these positions, I was wondering if there might be an alternative way to annotate this information?

src/hotspot/share/jeandle/jeandleUtils.cpp
```diff
@@ -53,6 +55,57 @@ llvm::Function* JeandleFuncSig::create_llvm_func(ciMethod* method, llvm::Module&
...
+      // Skip interface types (see comment above for receiver).
+      if (klass->is_loaded() && !klass->is_interface()) {
+        Klass* klass_enc = (Klass*)(klass->constant_encoding());
+        func->addParamAttr(arg_idx, llvm::Attribute::get(context,
+            llvm::jeandle::Attribute::JavaKlass,
+            std::to_string((uintptr_t)klass_enc)));
+        if (klass->is_instance_klass() && klass->as_instance_klass()->is_final()) {
+          func->addParamAttr(arg_idx, llvm::Attribute::get(context,
+              llvm::jeandle::Attribute::JavaKlassExact));
+        }
```
minchao1024: The current check for `is_instance_klass()` is overly restrictive. `ArrayKlass` should also be considered for type sharpening.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
taofengliu: We can attach metadata.

minchao1024 (approved):

taofengliu:
@mergifyio queue

taofengliu:
@mergifyio queue

Issue #440, opened 2026-04-20
Introduce a unified intrinsic framework for Jeandle Compiler


eapenzhan:
## Summary

Today, intrinsic support is still fairly ad hoc: different intrinsics tend to carry their own lowering logic, semantic assumptions, and runtime glue. That works for a small number of cases, but it does not scale well as we add more intrinsic support.

The goal of this work is to make intrinsic support more structured, easier to extend, and easier to reason about.

## Motivation

We currently lack a single framework that can consistently answer:

- what an intrinsic means semantically
- how its lowering should be selected
- how runtime / deopt / GC-related requirements should be represented
- how JavaOp-based lowering and direct LLVM lowering fit into the same model

Without that, it is easy for intrinsic-specific logic to become fragmented and harder to maintain over time.

## Proposal

Introduce an intrinsic framework with three main pieces:

- **descriptor / registry layer**  
  A central place to describe supported intrinsics and their semantic properties.

- **policy / lowering-plan layer**  
  A separate layer that decides how a given intrinsic should be lowered, e.g. as direct LLVM IR, LLVM builtin, JavaOp, or runtime/stub call.

- **runtime-glue-aware lowering model**  
  Runtime/deopt/GC-related requirements should be treated as first-class parts of intrinsic lowering, rather than handled in an ad hoc way per intrinsic.

The intent is to keep intrinsic semantics separate from current lowering implementation details wherever possible.

## Initial scope

The initial scope is twofold:

### 1. Adapt existing intrinsics to the new framework

As a first step, existing intrinsic support should be moved onto the unified framework instead of continuing to grow as one-off lowering code. This includes, for example:

- existing `Math` / `StrictMath`-related intrinsics
- simple bit-manipulation intrinsics
- primitive type coercion / raw-bits intrinsics

These are good initial candidates because they already exist in some form and can help validate the framework's direct LLVM IR / LLVM builtin / runtime-call paths.

### 2. Add a small number of framework-driven intrinsics

A small initial batch of additional intrinsics can then be used to exercise framework paths that are not covered well by the existing math-style intrinsics, for example:

- `Object.getClass()`
- `Reference.get()`
- `Reference.refersTo0()`
- `PhantomReference.refersTo0()`

These are mainly intended to validate the framework shape, especially around semantic descriptors, JavaOp-based lowering, and GC-sensitive memory semantics.

## Expected benefits

- a clearer and more uniform way to add new intrinsics
[...]

PR #442, opened 2026-04-21
feat: Implement a unified intrinsic framework for Jeandle Compiler

eapenzhan:
## Related issue(s):

issue #440 

## Performance Test

 | test | with this PR (ms) | main branch (ms) | diff |
  |----------|-------------|---------------|---------|
  | chi-squre | 5544 | 6218 | 10.8% |
  | future-genetic | 4254 | 4569 | 6.9% |
  | fj-kmeans | 1875 | 1982 | 5.4% |
  | dec-tree | 3518 | 4206 | 16.4% |


src/hotspot/share/jeandle/jeandleIntrinsicRegistry.cpp
```diff
@@ -0,0 +1,275 @@
...
+constexpr JeandleIntrinsicDescriptor JeandleIntrinsicRegistryTable::_intrinsic_table[];
+
+const JeandleIntrinsicDescriptor* JeandleIntrinsicRegistry::lookup(vmIntrinsics::ID id) {
+  for (const JeandleIntrinsicDescriptor* it = JeandleIntrinsicRegistryTable::table_begin();
+       it != JeandleIntrinsicRegistryTable::table_end();
+       ++it) {
+    if (it->id == id) {
+      return it;
+    }
+  }
```
taofengliu: Will this take a long time? Is there any lookup method that is more efficient?

src/hotspot/share/jeandle/jeandleIntrinsicRegistry.hpp
```diff
@@ -0,0 +1,106 @@
...
+  PureMath,
+  LibmMath,
+  TypeCoercion,   // bit-level reinterpretation between primitive types (bitcast)
+  MemorySemantic,
+  TypeSemantic,
+  AllocationSemantic,
+  BarrierSemantic,
+  MacroSemantic,
+  ArrayScan       // byte-array scan intrinsics (countPositives, etc.)
+};
```
taofengliu: This class is hardly never used. Can we erase it?

src/hotspot/share/jeandle/jeandleIntrinsicRegistry.hpp
```diff
@@ -0,0 +1,106 @@
...
+struct JeandleIntrinsicSemantics {
+  JeandleIntrinsicCategory category;
+  JeandleControlSemantics control;
+  JeandleMemorySemantics memory;
+};
+
+enum class JeandleLoweringKind {
+  PureIRNode,
+  RuntimeLeafCall,
+  GuardedHybrid,
```
taofengliu: Add a comment to describe.

src/hotspot/share/jeandle/jeandleIntrinsicPolicy.cpp
```diff
@@ -0,0 +1,233 @@
...
+    } else {
+      // JeandleIntrinsicImplKind::JavaOperation (and Unsupported fallback)
+      plan.mode = JeandleLoweringMode::JavaOpCall;
+    }
+
+    // exception edge overrides only non-JavaOp paths; JavaOp invoke is handled
+    // by emit_java_op_invoke which is selected in lowering based on needs_exception_edge
+    if (control.needs_exception_edge && plan.mode != JeandleLoweringMode::JavaOpCall) {
+      plan.mode = JeandleLoweringMode::ManagedRuntimeInvoke;
+    }
```
taofengliu: As I understand, PureLLVM doesn't need exception edge. Do we need to some assertions to make sure this?
By the way, if we need other constraints on intrinsic definitions, we can do it altogether in the stage of compiler initialization or somewhere else.

src/hotspot/share/jeandle/jeandleIntrinsicRegistry.cpp
```diff
@@ -0,0 +1,275 @@
...
+  static constexpr const JeandleIntrinsicDescriptor* table_begin() {
+    return &_intrinsic_table[0];
+  }
+
+  static constexpr const JeandleIntrinsicDescriptor* table_end() {
+    return &_intrinsic_table[ARRAY_SIZE(_intrinsic_table)];
+  }
+
+ private:
+  static constexpr JeandleIntrinsicDescriptor _intrinsic_table[] = {
```
taofengliu: The intrinsic declarations are hard to read --- we don't know what each field mean directly from this declaration, especially for the bool fields. I think we need a more readable implementation for this.
If it's hard to do it perfectly, we can add a comment to describe the structure of JeandleIntrinsicDescriptor here.

src/hotspot/share/jeandle/jeandleIntrinsicIRSemantics.hpp
taofengliu, line 1: Move implementations in this file to jeandleIntrinsicPolicy.hpp/cpp

src/hotspot/share/jeandle/jeandleRuntimeEntrypoints.hpp
```diff
@@ -0,0 +1,51 @@
...
+#include "memory/allocation.hpp"
+
+struct JeandleRuntimeEntrypoint {
+  llvm::FunctionCallee callee;
+  llvm::CallingConv::ID calling_conv;
+  bool is_gc_leaf;
+  const char* well_known_name;
+};
+
+class JeandleRuntimeEntrypoints : public AllStatic {
```
taofengliu: It's only used for intrinsic related code. We can give it a more specific name, and move it to jeandleIntrinsicLowering.hpp

taofengliu (commented):
The code seems a bit over-abstracted in some places. Let's find a way to make it simpler.

taofengliu:
The DCO check failed.

src/hotspot/share/jeandle/jeandleIntrinsicPolicy.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+  switch (desc.lowering_kind) {
+    case JeandleLoweringKind::PureIRNode: {
+      // PureIRNode intrinsics are unconditionally supported; the impl_kind is
+      // determined by category alone (no capability query needed).
+      const JeandleIntrinsicImplKind k =
+          (desc.category == JeandleIntrinsicCategory::TypeCoercion ||
+           desc.category == JeandleIntrinsicCategory::BarrierSemantic)
+              ? JeandleIntrinsicImplKind::IRInstruction
+              : JeandleIntrinsicImplKind::LLVMBuiltinCall;
```
taofengliu: Here's a complexity for adding a new intrinsic that is neither an `JeandleIntrinsicCategory::TypeCoercion` or `JeandleIntrinsicCategory::BarrierSemantic` but implemented as `JeandleIntrinsicImplKind::IRInstruction`.
I think the `JeandleIntrinsicPolicy::decide` depends on the intrinsic category.

src/hotspot/share/jeandle/jeandleIntrinsicRegistry.hpp
eapenzhan: It is the first-level lowering dispatch mechanism. lower() switches on category to pick the handler; all 35 intrinsics carry their categories;. Deleting it would flatten the whole dispatch architecture. Addressed instead by adding doc comments on the enum and on lower() so the design is self-evident on next read.

eapenzhan: I have added comments on the items of the enum to make it easier to understand.

src/hotspot/share/jeandle/jeandleIntrinsicRegistry.cpp
eapenzhan: Each field has a // comment; the original raw bool fields have been folded into control_flags / memory_flags / support_flags bitmasks with named accessors, which fixes the "what does this bool mean" problem at the same time

eapenzhan:
> The code seems a bit over-abstracted in some places. Let's find a way to make it simpler.

Thanks for the review. On the overall "over-abstracted" feedback, I did a targeted simplification pass on top of the inline-comment fixes:

1. **Removed declared-but-never-read state.** Four items had zero readers across the whole codebase, only writers: the `JeandleFallbackPolicy` enum + `descriptor.fallback_policy`, `requires_nonnull_receiver`, `Decision::reason` (and all `reason` formal parameters), and the `ImplKind::GuardedHybrid` dead branch. All deleted.

2. **Collapsed redundant intermediates.** `JeandleLoweringMode` was a derived enum that could be reproduced directly from `impl_kind + needs_exception_edge`. Deleted; `JeandleIRSemanticPlan` is now three bools, and the `jeandle.lowering.mode` metadata string is produced on the fly without naming an intermediate. `Decision` is now `{supported, impl_kind, ir_plan}`.

3. **Dropped decorative `AllStatic` wrappers in .cpp files** (`PolicyHelper`, `IRSemanticNames`, `LoweringHelper`) in favor of file-scope `static` functions. (Two header-level `AllStatic` classes — `JeandleIntrinsicSupport` and `JeandleIntrinsicIRSemantics` — are still classes for now.)

src/hotspot/share/jeandle/jeandleIntrinsicRegistry.cpp
eapenzhan: Thanks for flagging this. You're right that the original linear scan was wasteful — it ran on every intrinsic candidate during compilation.

I changed it to an O(1) direct-indexed lookup table. `JeandleIntrinsicRegistry` now owns

```cpp
static const JeandleIntrinsicDescriptor* _lookup[(int)vmIntrinsics::ID_LIMIT];
```

populated once at JIT initialization (called from `JeandleCompiler::initialize` on the `should_perform_init()` path), and the runtime `lookup(vmIntrinsics::ID id)` is now a single array index — no scanning, no string comparison. Memory cost is `414 × 8 bytes ≈ 3.3 KB`, which is negligible.

This matches what C1 and C2 do for the same problem: their `switch (intrinsic_id())` dispatch is compiled into a jump table, also O(1). A dense array is just the data-driven equivalent. The init path also runs a `validate_descriptor` pass in debug builds to catch table mistakes early (duplicate IDs, JavaOperation without a `java_op_name`, etc.).

src/hotspot/share/jeandle/jeandleIntrinsicPolicy.cpp
eapenzhan: Good catch — both points are handled together in the registry's init-time validation.

`PureIRNode` (the renamed-but-equivalent of `PureLLVM` in the current code) requiring no exception edge is now asserted in `validate_descriptor()` in `jeandleIntrinsicRegistry.cpp`:

```cpp
assert(desc.lowering_kind != JeandleLoweringKind::PureIRNode ||
       !desc.needs_exception_edge(),
       "PureIRNode cannot require an exception edge");
```

`validate_descriptor()` is invoked once per entry from `JeandleIntrinsicRegistry::initialize()`, which `JeandleCompiler::initialize` calls on the `should_perform_init()` path. So this is exactly the "do it altogether at compiler initialization" location you suggested — the whole table gets cross-checked once at JIT startup rather than per-invocation.

To answer the second half of your comment: yes, this is the place to add new descriptor-level constraints. A few other invariants live there already:

```cpp
assert(desc.id != vmIntrinsics::_none && vmIntrinsics::is_valid_id(desc.id),
       "invalid Jeandle intrinsic id");
assert(desc.trap_throttle_mask == 0 || desc.may_deopt(),
       "trap throttling requires a deopt-capable intrinsic");
assert(desc.lowering_kind != JeandleLoweringKind::JavaOperation || desc.java_op_name != nullptr,
       "JavaOperation descriptor must have a non-null java_op_name");
assert(desc.java_op_name == nullptr || desc.java_op_name[0] != '\0',
       "empty JavaOp name string");
```

Anything else you'd like enforced here, I'll add to the same function.

src/hotspot/share/jeandle/jeandleIntrinsicPolicy.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+static JeandleIRSemanticPlan make_plan(const JeandleIntrinsicDescriptor& desc,
+                                       JeandleIntrinsicImplKind impl_kind) {
+  const bool is_javaop         = (impl_kind == JeandleIntrinsicImplKind::JavaOperation);
+  const bool is_pure_llvm      = (impl_kind == JeandleIntrinsicImplKind::IRInstruction ||
+                                  impl_kind == JeandleIntrinsicImplKind::LLVMBuiltinCall);
+  const bool is_managed        = is_managed_runtime_call(impl_kind, desc);
+  const bool is_leaf_call      = (impl_kind == JeandleIntrinsicImplKind::HotSpotStub ||
+                                  impl_kind == JeandleIntrinsicImplKind::SharedRuntime) && !is_managed;
+  const bool needs_unwind_edge = desc.needs_exception_edge() && !is_javaop;
```
taofengliu: Why javaop don't need unwind edge?

src/hotspot/share/jeandle/jeandleIntrinsicPolicy.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+  JeandleIRSemanticPlan plan{};
+
+  // attach_deopt_bundle is required when:
+  //   - the intrinsic itself can deopt;
+  //   - the call site is GC-sensitive (RewriteStatepointsForGC needs interpreter state);
+  //   - the call crosses a managed-runtime boundary (call or invoke);
+  //   - the call is a JavaOp (conservatively non-leaf until JavaOp infrastructure can
```
taofengliu: We can always attatch deopt bundle for JavaOp, because if a JavaOp doesn't need deopt bundle, the attatched bundle will be dropped after the JavaOp is inlined. But it may prevent some optimizations

src/hotspot/share/jeandle/jeandleIntrinsicPolicy.cpp
```diff
@@ -0,0 +1,187 @@
...
+  const bool is_managed        = is_managed_runtime_call(impl_kind, desc);
+  const bool is_leaf_call      = (impl_kind == JeandleIntrinsicImplKind::HotSpotStub ||
+                                  impl_kind == JeandleIntrinsicImplKind::SharedRuntime) && !is_managed;
+  const bool needs_unwind_edge = desc.needs_exception_edge() && !is_javaop;
+
+  JeandleIRSemanticPlan plan{};
+
+  // attach_deopt_bundle is required when:
+  //   - the intrinsic itself can deopt;
+  //   - the call site is GC-sensitive (RewriteStatepointsForGC needs interpreter state);
```
taofengliu: RewriteStatepointsForGC doesn't need interpreter state. In fact, any program point that can potentially enter a safepoint is also a potential deoptimization point, so each of them requires a deopt bundle. We need to check this constraint in `validate_descriptor`

src/hotspot/share/jeandle/jeandleIntrinsicPolicy.cpp
```diff
@@ -0,0 +1,187 @@
...
+  //   - the call crosses a managed-runtime boundary (call or invoke);
+  //   - the call is a JavaOp (conservatively non-leaf until JavaOp infrastructure can
+  //     derive precise bundle requirements).
+  plan.attach_deopt_bundle = desc.may_deopt() || desc.needs_gc_state() ||
+                             is_managed || needs_unwind_edge || is_javaop;
```
taofengliu: `is_managed` is redundant.

src/hotspot/share/jeandle/jeandleIntrinsicPolicy.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+  // gc-leaf-function attribute is only correct on truly side-effect-free leaf paths:
+  // pure LLVM IR or a leaf runtime call with no GC/exception/deopt interaction.
+  plan.attach_gc_leaf_attr = !desc.needs_gc_state() &&
+                             !desc.has_memory_effect() &&
```
taofengliu: Why all intrinsics that have memory effect cannot be gc leaf?

src/hotspot/share/jeandle/jeandleUtils.cpp
```diff
@@ -80,6 +82,26 @@ bool is_effectively_final(Klass* klass) {
   return false;
 }
 
+void maybe_attach_java_klass_ret_attr(llvm::CallBase* call,
```
taofengliu:
```suggestion
void attach_java_klass_ret_attr(llvm::CallBase* call,
```

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1878,123 +1886,35 @@ void JeandleAbstractInterpreter::invoke() {
...
-      }
-      break;
-    }
-    default:
-      return false;
+  // Honour HotSpot's standard -XX:DisableIntrinsic=<list> / -XX:ControlIntrinsic
+  // flags. When the id is disabled the caller falls back to a normal invoke,
+  // matching the behaviour of C1/C2 and giving Jeandle a deterministic way to
+  // run baseline comparisons against the intrinsic-lowered path.
+  if (vmIntrinsics::is_disabled_by_flags(desc->id)) {
```
taofengliu: C2 uses `AbstractCompiler::is_intrinsic_available`, which includes more check. Do we need this?

src/hotspot/share/jeandle/jeandleIntrinsicPolicy.cpp
```diff
@@ -0,0 +1,187 @@
...
+      // (e.g. pow(x,2) -> fmul) are chosen at lowering time and call refine().
+      JeandleIntrinsicCapabilities caps = JeandleIntrinsicSupport::query(desc);
+      const bool hotspot_first = caps.hotspot_preferred && caps.any_runtime();
+      if (hotspot_first) {
+        JeandleIntrinsicImplKind k = try_hotspot_path(caps);
+        if (k != JeandleIntrinsicImplKind::Unsupported) return make_decision(desc, k);
+      }
+      if (caps.has_llvm_builtin) return make_decision(desc, JeandleIntrinsicImplKind::LLVMBuiltinCall);
+      return unsupported(desc);
+    }
```
taofengliu: The code here is the same as `JeandleLoweringKind::RuntimeLeafCall`. What's the difference between GuardedHybrid and RuntimeLeafCall?

src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
```diff
@@ -0,0 +1,764 @@
...
+  llvm::SmallVector<llvm::OperandBundleDef, 1> bundles =
+    JeandleIntrinsicIRSemantics::build_operand_bundles(_interp, decision.ir_plan);
+  llvm::InvokeInst* invoke = _interp->call_java_op_ex(desc.java_op_name, args, bundles);
+  annotate_generated_instruction(*invoke, desc, decision);
+  add_string_attr(*invoke, "jeandle.java_op", desc.java_op_name);
+  attach_callee_return_klass_attr(invoke);
+  return invoke;
+}
+
+bool JeandleIntrinsicLowering::lower(const JeandleIntrinsicDescriptor& desc,
```
taofengliu: I think `JeandleIntrinsicCategory` is an unnecessary abstraction. Though we use it to determine the lower function, but we also do an intrinsic id bases switch to do the final work.
Can we directly do an intrinsic id bases switch here? Like `LibraryCallKit::try_to_inline`. Then delete `JeandleIntrinsicCategory`.
What do you think?

taofengliu (commented):
We can add some comments as a guidance about how to implement a new intrinsic to Jeandle.

taofengliu:
The DCO check failed

src/hotspot/share/jeandle/jeandleIntrinsicPolicy.cpp
taofengliu: Can we only attatch deopt bundle for a JavaOp when it may deopt or needs gc state?

src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
eapenzhan: Originally I was thinking it could dispatch the calls and merge some code together. But for now It seems not that usefull. I'll do as the hotspot style.

src/hotspot/share/jeandle/jeandleIntrinsicPolicy.cpp
eapenzhan: You're right that the comment misattributes the motivation — RewriteStatepointsForGC doesn't need interpreter state for its own work (GC root relocation); it just threads the "deopt" bundle into the statepoint. The real reason is the invariant you stated: any point that can safepoint is a potential deopt point, so it must carry interpreter state. I'll fix the comment in policy.cpp (and the matching one in jeandleIntrinsicIRSemantics.hpp).
It is is already structurally guaranteed by make_plan, I think it's not necessary to check in `validate_descriptor`.

src/hotspot/share/jeandle/jeandleIntrinsicPolicy.cpp
eapenzhan: It also needs. It just made a judgment beforehand. I'll delete it to prevent misunderstanding.

src/hotspot/share/jeandle/jeandleIntrinsicPolicy.cpp
```diff
@@ -0,0 +1,182 @@
...
+    if ((mask & 1u) != 0 &&
+        too_many_traps_at(caller, bci,
+                          static_cast<Deoptimization::DeoptReason>(reason_index))) {
+      return true;
+    }
+    reason_index++;
+    mask >>= 1;
+  }
+  return false;
+}
```
taofengliu: #451 has added these interfaces, we can use them.

src/hotspot/share/jeandle/jeandleIntrinsicPolicy.cpp
```diff
@@ -0,0 +1,182 @@
...
+  //   - the call site can safepoint: every safepoint is a potential deopt point,
+  //     so it must carry interpreter state. RewriteStatepointsForGC does not need
+  //     that state for its own work (GC root relocation); it merely threads the
+  //     "deopt" bundle into the gc.statepoint's deopt section for HotSpot to use
+  //     at runtime if the safepoint deopts.
+  //   - the call crosses a managed-runtime boundary (call or invoke);
+  //   - the call is a JavaOp (conservatively non-leaf until JavaOp infrastructure can
+  //     derive precise bundle requirements).
+  plan.attach_deopt_bundle = desc.may_deopt() || desc.needs_gc_state() ||
+                             is_managed || needs_unwind_edge || is_javaop;
```
taofengliu:
```suggestion
                             needs_unwind_edge;
```
If a JavaOp needs deopt bundle, its `desc.needs_gc_state()` or `desc.may_deopt()` returns true.

src/hotspot/share/jeandle/templatemodule/jeandleRuntimeDefinedJavaOps.cpp
```diff
@@ -195,6 +196,173 @@ DEF_JAVA_OP(post_barrier, 1, llvm::Type::getVoidTy(context),
...
+  llvm::GlobalVariable* offset_gv = template_module.getGlobalVariable("java_lang_ref_Reference.referent_offset", /*AllowInternal=*/true);
+  if (!offset_gv) {
+    RuntimeDefinedJavaOps::set_failed("java_lang_ref_Reference.referent_offset global not found in template module");
+    return;
+  }
+  llvm::Value* offset = ir_builder.CreateLoad(ir_builder.getInt32Ty(), offset_gv);
+  llvm::Value* referent_addr = ir_builder.CreateInBoundsGEP(ir_builder.getInt8Ty(), ref_obj, offset);
+  llvm::Type* ref_type = llvm::PointerType::get(context, llvm::jeandle::AddrSpace::JavaHeapAddrSpace);
+  llvm::LoadInst* referent = ir_builder.CreateLoad(ref_type, referent_addr);
+  referent->setAtomic(llvm::AtomicOrdering::Acquire);
```
wazlyqa: jeandle.reference_get JavaOp missing G1 SATB pre-barrier. Now, you can insert the barrier directly in the JavaOp, consistent with do_get_xxx:
```
  referent->setAtomic(llvm::AtomicOrdering::Acquire);
+ if (UseG1GC) {
+   llvm::Function* barrier_func = template_module.getFunction("jeandle.g1_pre_barrier_loaded");
+   assert(barrier_func != nullptr, "jeandle.g1_pre_barrier_loaded not found");
+   llvm::CallInst* call = ir_builder.CreateCall(barrier_func, {referent});
+   call->setCallingConv(llvm::CallingConv::Hotspot_JIT);
+ }
  ir_builder.CreateRet(referent);
```
As a follow-up, the jeandle.semantic.barrier_kind = "weak-referent-load" metadata already provides a hook for late barrier insertion. Similar to [#464](https://github.com/jeandle/jeandle-jdk/issues/464) (late insertion for array GC barriers), a GC-aware LLVM pass could be implemented to consume the barrier_kind metadata and insert the appropriate barrier after optimization.

src/hotspot/share/jeandle/jeandleIntrinsicPolicy.cpp
eapenzhan: Their true difference is in the lowering body: GuardedHybrid embeds fast/slow guards (like lower_pow_hybrid), while RuntimeLeafCall is a straight throughway.
There is no actual difference between the two at the policy level. I'll merge both cases here (via fallthrough) while retaining the two descriptor kinds as semantic labels.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
eapenzhan: I added the directive path without introducing the AbstractCompiler hierarchy. The directive support has been supplemented separately within the policy layer. However, checking LLVM support at this stage is a bit too early.

src/hotspot/share/jeandle/jeandleIntrinsicPolicy.cpp
eapenzhan: Yep, I'll remove the javaop condition and leave it just decided by the field attr.

src/hotspot/share/jeandle/jeandleIntrinsicPolicy.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+  JeandleIRSemanticPlan plan{};
+
+  // attach_deopt_bundle is required when:
+  //   - the intrinsic itself can deopt;
+  //   - the call site can safepoint: every safepoint is a potential deopt point,
+  //     so it must carry interpreter state. RewriteStatepointsForGC does not need
+  //     that state for its own work (GC root relocation); it merely threads the
+  //     "deopt" bundle into the gc.statepoint's deopt section for HotSpot to use
+  //     at runtime if the safepoint deopts.
```
taofengliu:
```suggestion
  //     so it must carry interpreter state.
```

src/hotspot/share/jeandle/jeandleIntrinsicPolicy.cpp
```diff
@@ -0,0 +1,139 @@
...
+  //   - the call crosses a managed-runtime boundary (call or invoke);
+  //   - the call needs an exception edge (the unwind path crosses Java EH).
+  // JavaOps follow the same uniform rule: a JavaOp that promises (via its
+  // descriptor) not to deopt / safepoint / throw gets no bundle, allowing LLVM
+  // attribute-based DCE and aliasing before inlining.
+  plan.attach_deopt_bundle = desc.may_deopt() || desc.needs_gc_state() ||
+                             is_managed || desc.needs_exception_edge();
```
taofengliu: Consider a `is_managed` intrinsic, its `desc.may_deopt() || desc.needs_gc_state()` is true.

src/hotspot/share/jeandle/jeandleIntrinsicPolicy.cpp
```diff
@@ -0,0 +1,139 @@
...
+  return desc.needs_gc_state() || desc.may_deopt();
+}
+
+static JeandleIRSemanticPlan make_plan(const JeandleIntrinsicDescriptor& desc,
+                                       JeandleIntrinsicImplKind impl_kind) {
+  const bool is_pure_llvm = (impl_kind == JeandleIntrinsicImplKind::IRInstruction ||
+                             impl_kind == JeandleIntrinsicImplKind::LLVMBuiltinCall);
+  const bool is_managed   = is_managed_runtime_call(impl_kind, desc);
+  const bool is_leaf_call = (impl_kind == JeandleIntrinsicImplKind::HotSpotStub ||
+                             impl_kind == JeandleIntrinsicImplKind::SharedRuntime) && !is_managed;
```
taofengliu: Simplify this, without using `is_managed`. Then remove `is_managed` related code and `is_managed_runtime_call`.

src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
```diff
@@ -0,0 +1,777 @@
...
+
+bool JeandleIntrinsicLowering::lower_pure_math(const JeandleIntrinsicDescriptor& desc,
+                                               const JeandleIntrinsicDecision& decision) {
+  for (const PureMathSpec& spec : kPureMathTable) {
+    if (spec.vm_id != desc.id) continue;
+    llvm::LLVMContext& ctx = *_interp->_context;
+    llvm::IRBuilder<>& builder = _interp->_ir_builder;
+
+    llvm::SmallVector<llvm::Value*, 2> args;
+    args.push_back(pop_by_type(_interp->_jvm,spec.operand_type));
```
```suggestion
    args.push_back(_interp->_jvm->pop(spec.operand_type));
```
Use this interface directly.
Check other usage also.

src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
```diff
@@ -0,0 +1,777 @@
...
+    llvm::CallInst* call = builder.CreateIntrinsic(
+        JeandleType::java2llvm(spec.operand_type, ctx), spec.llvm_id, args);
+    annotate_generated_instruction(*call, desc, decision);
+
+    llvm::Value* result = call;
+    if (spec.result_type != spec.operand_type) {
+      // e.g. Long.bitCount(long): ctpop gives i64, push as i32.
+      result = builder.CreateTrunc(call, JeandleType::java2llvm(spec.result_type, ctx));
+    }
+    push_by_type(_interp->_jvm,spec.result_type, result);
```
```suggestion
    _interp->_jvm->push(spec.result_type, result);
```

src/hotspot/share/jeandle/jeandleIntrinsicIRSemantics.hpp
```diff
@@ -0,0 +1,79 @@
...
+// IRSemantics evolves on its own axis: new GC-aware metadata, new alias/
+// noalias attributes, new keys synchronised with jeandle-llvm passes — all
+// land here without touching Policy or the descriptor.  It depends on the
+// LLVM headers and on AbstractInterpreter; Policy stays ciMethod-only.
+//
+// Two kinds of facts are attached to the emitted IR here:
+//
+//   1. Behavioural contracts consumed by LLVM passes today.
+//      - The "gc-leaf-function" function attribute is read by
+//        RewriteStatepointsForGC to skip statepoint insertion on leaf calls.
```
taofengliu:
```suggestion
//        RewriteStatepointsForGC to skip statepoint rewriting on leaf calls.
```

src/hotspot/share/jeandle/jeandleIntrinsicIRSemantics.hpp
```diff
@@ -0,0 +1,79 @@
...
+//
+//   2. Observability labels reserved as a forward contract for GC-aware /
+//      barrier-aware LLVM passes that have not yet been written:
+//        jeandle.intrinsic.id, jeandle.lowering.mode,
+//        jeandle.semantic.barrier_kind, jeandle.runtime.entry.
+//      These have no LLVM-side consumer today; they appear in IR dumps and
+//      serve as a stable place for a future pass (e.g. one that reads
+//      barrier_kind to emit collector-specific barriers) to plug in without
+//      re-threading the descriptor through the lowering layer.  Treat them as
+//      part of the IR contract: keep the strings stable, do not relocate.
```
taofengliu: What kinds of LLVM passes will read these metadata?

src/hotspot/share/jeandle/jeandleIntrinsicPolicy.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+enum class JeandleIntrinsicImplKind {
+  Unsupported,
+  // Pure-IR sub-kinds: inline into the current function's IR, no call boundary
+  IRInstruction,    // bare LLVM IR instruction: bitcast, fmul, icmp, etc.
+  LLVMBuiltinCall,  // call to a named llvm.* builtin or platform-specific LLVM target intrinsic
+                    // (e.g. llvm.fabs, llvm.ctpop, llvm.x86.sse2.pause, llvm.aarch64.hint)
+  // Non-pure-IR kinds: cross call boundaries or delegate to the runtime
+  HotSpotStub,
+  SharedRuntime,
```
taofengliu: Add comments to explain HotSpotStub and SharedRuntime

src/hotspot/share/jeandle/jeandleIntrinsicPolicy.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+
+enum class JeandleIntrinsicImplKind {
+  Unsupported,
+  // Pure-IR sub-kinds: inline into the current function's IR, no call boundary
+  IRInstruction,    // bare LLVM IR instruction: bitcast, fmul, icmp, etc.
+  LLVMBuiltinCall,  // call to a named llvm.* builtin or platform-specific LLVM target intrinsic
+                    // (e.g. llvm.fabs, llvm.ctpop, llvm.x86.sse2.pause, llvm.aarch64.hint)
+  // Non-pure-IR kinds: cross call boundaries or delegate to the runtime
+  HotSpotStub,
```
taofengliu:
```suggestion
  HotspotStub,
```

src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
```diff
@@ -0,0 +1,777 @@
...
+  llvm::LLVMContext& ctx = *_interp->_context;
+
+  llvm::Value* exp = _interp->_jvm->dpop();
+  llvm::Value* base = _interp->_jvm->dpop();
+
+  // Constant fast path: pow(x, 2.0) => x * x.
+  // This overrides the policy decision (IRInstruction instead of LLVMBuiltinCall/HotSpotStub);
+  // refine() produces accurate metadata for the emitted instruction.
+  if (is_double_constant(exp, 2.0,
+                                                         _interp->_module.getDataLayout())) {
```
```suggestion
  if (is_double_constant(exp, 2.0, _interp->_module.getDataLayout())) {
```

src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
```diff
@@ -0,0 +1,777 @@
...
+    if (llvm::Instruction* inst = llvm::dyn_cast<llvm::Instruction>(fast)) {
+      annotate_generated_instruction(*inst, desc, refined);
+    }
+    _interp->_jvm->dpush(fast);
+    return true;
+  }
+
+  // Constant fast path: pow(x, 0.5) => llvm.sqrt(x) only for x > 0.0.
+  if (is_double_constant(exp, 0.5,
+                                                         _interp->_module.getDataLayout())) {
```
```suggestion
  if (is_double_constant(exp, 0.5, _interp->_module.getDataLayout())) {
```

src/hotspot/share/jeandle/jeandleIntrinsicEntrypoints.hpp
```diff
@@ -0,0 +1,51 @@
...
+class JeandleIntrinsicEntrypoints : public AllStatic {
+ public:
+  static bool resolve_math(vmIntrinsics::ID id,
+                           JeandleIntrinsicImplKind impl_kind,
+                           llvm::Module& module,
+                           JeandleIntrinsicEntrypoint& out);
+
+  static bool resolve_count_positives(llvm::Module& module,
+                                      JeandleIntrinsicEntrypoint& out);
+};
```
taofengliu: These two names are so confusing. Move `resolve_math` and `resolve_count_positives` to `JeandleIntrinsicEntrypoint`.

src/hotspot/share/jeandle/jeandleIntrinsicRegistry.hpp
```diff
@@ -0,0 +1,139 @@
...
+  RawReferentRead,   // Reference.refersTo0: intentional raw read, bypasses GC barrier
+  CardMarkPost,      // post-write card table mark
+  VolatileLoad,      // acquire semantics
+  VolatileStore,     // release semantics
+};
+
+enum class JeandleLoweringKind {
+  PureIRInstruction, // lower to a bare LLVM IR instruction (bitcast, fence)
+  PureLLVMBuiltin,   // lower to a named llvm.* builtin or LLVM target intrinsic
+  RuntimeLeafCall,   // emit a runtime/stub call that does not safepoint or throw
```
taofengliu: Is this leaf means "gc-leaf"? If not, we can call it "RuntimeCall"

src/hotspot/share/jeandle/jeandleIntrinsicRegistry.hpp
```diff
@@ -0,0 +1,139 @@
...
+};
+
+enum class JeandleMemoryBarrierKind {
+  None,
+  WeakReferentLoad,  // Reference.get: GC-specific load barrier required
+  RawReferentRead,   // Reference.refersTo0: intentional raw read, bypasses GC barrier
+  CardMarkPost,      // post-write card table mark
+  VolatileLoad,      // acquire semantics
+  VolatileStore,     // release semantics
+};
```
taofengliu: Most of intrinsics don't need this. We can remove this, and consider to handle barrier in specific needed intrinsics.

src/hotspot/share/jeandle/jeandleIntrinsicRegistry.hpp
```diff
@@ -0,0 +1,139 @@
...
+// multi-valued enum tied to GC barrier semantics, orthogonal to the data-flow
+// direction encoded here.
+//
+// Direction bits (MEM_READ / MEM_WRITE) describe the LLVM-visible memory
+// effects of the intrinsic's lowered call site, and are translated into LLVM
+// `memory()` call-site attributes by JeandleIntrinsicLowering when the call is
+// a leaf path (no safepoint / deopt / exception edge).  MEM_ORDERING_ONLY
+// marks fences and similar primitives that constrain ordering without
+// transferring data; it is mutually exclusive with MEM_READ / MEM_WRITE.
+enum JeandleMemoryFlag : uint8_t {
```
taofengliu: It seems no code use this.

eapenzhan: This is used to add call attributes with read-only/writeonly/readnone to show llvm some black-box call whether it'll effect memory. Marking a call as readonly allows the [LLVM optimizer](https://llvm.org/docs/Passes.html) to perform several key transformations. Maybe useful or not. It depends.

src/hotspot/share/jeandle/jeandleIntrinsicRegistry.hpp
eapenzhan: I'll add them into JeandleMemoryFlags

src/hotspot/share/jeandle/jeandleIntrinsicRegistry.hpp
```diff
@@ -0,0 +1,123 @@
...
+// a weak referent load — after optimization and before JavaOperationLower(1)
+// inlines the JavaOp body.  How the kind is threaded to LLVM (named metadata on
+// the call site? an operand bundle? a marker intrinsic?) is NOT decided yet, so
+// nothing is emitted from it today (validate_descriptor only checks it for
+// consistency); meanwhile the G1 pre-barrier stays inside the JavaOp body for
+// correctness.  It is lowering-independent (a future inlined load/store could
+// carry the same kind), so it lives on the base descriptor, not in
+// JeandleCallInfo.  See jeandle-docs/intrinsics/pending-barrier-semantic-stability.md.
+// Barriers are mutually exclusive, so a scoped enum models them better than a bitmask.
+enum class JeandleBarrierKind : uint8_t {
```
taofengliu: If this is not used currently, I suggest to remove this.

src/hotspot/share/jeandle/jeandleIntrinsicRegistry.cpp
```diff
@@ -0,0 +1,352 @@
...
+//       `V(name, intrinsic, well_known)`.  NOTE: currently models the single
+//       "double in -> double out" shape only (libm math today); a differently
+//       shaped runtime-stub intrinsic needs its own row/table.
+//   JEANDLE_CALL_JAVAOP_TABLE — Call delegating to a JavaOp.
+//       `V(name, java_op, ctrl, mem, barrier, arg_count, result, arg_types...)`.
+//   JEANDLE_PURE_TABLE — PureLLVM (bare IR / inline-asm / uncommon_trap); no
+//       call site, call_info == nullptr.  `V(name)`.
+//
+// Intrinsics that do not fit a table (Hybrid bodies, trap-throttled PureLLVM) are
+// written out explicitly right after the tables.  After adding a row:
```
taofengliu:
```suggestion
// written out explicitly in the end of _intrinsic_table.
//
// Then add the corresponding implementation in the following locations:
```

src/hotspot/share/jeandle/jeandleIntrinsicCallInfo.hpp
```diff
@@ -0,0 +1,159 @@
...
+  RuntimeStub,
+};
+
+// Property-driven runtime-callee resolver.  One function per runtime routine that
+// materializes the stub / SharedRuntime FunctionCallee in the given module.
+// Storing these as data on the descriptor lets emit_simple_call_intrinsic and the
+// Hybrid bodies resolve a callee generically — never switching on intrinsic id.
+using JeandleRuntimeCalleeFn = llvm::FunctionCallee (*)(llvm::Module&);
+
+struct JeandleCallInfo {
```
taofengliu:
```suggestion
struct JeandleIntrinsicCallInfo {
```
This name conflicts with `CallSiteInfo`.

src/hotspot/share/jeandle/jeandleIntrinsicCallInfo.hpp
```diff
@@ -0,0 +1,159 @@
...
+enum JeandleSupportFlag : uint8_t {
+  SUPPORT_NONE          = 0,
+  // A HotSpot-generated stub or SharedRuntime fallback is an available impl.
+  SUPPORT_HOTSPOT_STUB  = 1u << 0,
+  // LLVM has a builtin or direct IR representation for this intrinsic.
+  SUPPORT_LLVM_INTRIN   = 1u << 1,
+};
+
+// Which kind of callee a Call / Hybrid intrinsic targets.
+enum class JeandleCalleeKind : uint8_t {
```
taofengliu:
```suggestion
enum class JeandleIntrinsicCalleeKind : uint8_t {
```

src/hotspot/share/jeandle/jeandleIntrinsicRegistry.cpp
```diff
@@ -0,0 +1,352 @@
...
+  V(rint,         rint,  T_DOUBLE)
+
+// Call backed by a HotSpot runtime stub / SharedRuntime routine (one double in,
+// one double out — libm math is the only user today).  The stub and SharedRuntime
+// resolvers are derived from vm_name by token paste (StubRoutines_<vm_name>_callee
+// / SharedRuntime_<vm_name>_callee); the llvm builtin is the fallback when no
+// runtime path is available/preferred.  Columns:
+//   vm_name       — vmIntrinsics::_<vm_name>; also drives the resolver names
+//   llvm_builtin  — llvm::Intrinsic::<llvm_builtin> id (builtin fallback)
+#define JEANDLE_CALL_RUNTIME_STUB_TABLE(V) \
```
taofengliu: We can merge JEANDLE_CALL_RUNTIME_STUB_TABLE and JEANDLE_CALL_LLVM_BUILTIN_TABLE.
Don't constraint on the number and type of arguments and return value.

eapenzhan: I think llvm-buildin has different meaning in the macro. It's not a good idea to merge them together even though they have the same format. What do you think?

src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
```diff
@@ -0,0 +1,607 @@
...
+
+    llvm::SmallVector<llvm::Value*, 2> args;
+    args.push_back(_interp->_jvm->pop(spec.operand_type));
+    if (spec.needs_poison_flag) {
+      args.push_back(builder.getInt1(false));
+    }
+
+    llvm::CallInst* call = builder.CreateIntrinsic(
+        JeandleType::java2llvm(spec.operand_type, ctx), spec.llvm_id, args);
+    annotate_generated_instruction(*call, desc);
```
taofengliu: LLVM does not rewrite intrinsic calls to statepoint. So we don't need to call this.

src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
```diff
@@ -0,0 +1,607 @@
...
+  }
+  llvm::APFloat expected_value(expected);
+  return fp_constant->getValueAPF().bitwiseIsEqual(expected_value);
+}
+
+void JeandleIntrinsicLowering::annotate_generated_instruction(llvm::Instruction& inst,
+                                                              const JeandleIntrinsicDescriptor& desc,
+                                                              const JeandleIntrinsicEntrypoint* entry) const {
+  JeandleIntrinsicIRSemantics::annotate_instruction(inst, desc, entry);
+}
```
taofengliu: Don't add this wrapper, we can call `JeandleIntrinsicIRSemantics::annotate_instruction` directly

src/hotspot/share/jeandle/jeandleIntrinsicSupport.hpp
```diff
@@ -0,0 +1,59 @@
...
+  bool any_path()    const { return has_llvm_builtin || any_runtime(); }
+};
+
+class JeandleIntrinsicSupport : public AllStatic {
+ public:
+  // Return the set of available lowering paths for the given descriptor.
+  // All stub-installation, CPU-feature, and flag checks live in this single
+  // method; callers receive a plain capability struct and apply their own
+  // priority rules.
+  static JeandleIntrinsicCapabilities query(const JeandleIntrinsicDescriptor& desc);
```
taofengliu:
```suggestion
  static bool query(vmIntrinsics::ID id);
```

src/hotspot/share/jeandle/jeandleIntrinsicEntrypoints.hpp
```diff
@@ -0,0 +1,50 @@
...
+  llvm::FunctionCallee callee;
+  llvm::CallingConv::ID calling_conv;
+  bool is_gc_leaf;
+};
+
+class JeandleIntrinsicEntrypoints : public AllStatic {
+ public:
+  static bool resolve_count_positives(llvm::Module& module,
+                                      JeandleIntrinsicEntrypoint& out);
+};
```
taofengliu: We don't need to use a separate general class to resolve the entry point of specific intrinsics. We can do it in their lower functions.

src/hotspot/share/jeandle/jeandleIntrinsicIRSemantics.hpp
```diff
@@ -0,0 +1,63 @@
...
+// nothing is stamped (see the TODO in the .cpp).  The observability-only keys
+// (jeandle.intrinsic.id / jeandle.lowering.mode / jeandle.runtime.entry) were
+// dropped in the call-shape refactor as they had no consumer.
+class JeandleIntrinsicIRSemantics : public AllStatic {
+ public:
+  static llvm::SmallVector<llvm::OperandBundleDef, 1> build_operand_bundles(
+      JeandleAbstractInterpreter* interp, bool attach_deopt_bundle);
+  static void annotate_instruction(llvm::Instruction& inst,
+                                   const JeandleIntrinsicDescriptor& desc,
+                                   const JeandleIntrinsicEntrypoint* entry = nullptr);
```
taofengliu:
```suggestion
  static void annotate_call(llvm::CallBase* call,
                                   const JeandleIntrinsicCallInfo call_info);
```

src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
```diff
@@ -0,0 +1,607 @@
...
+  annotate_generated_instruction(*site, desc, entry);
+  attach_callee_return_klass_attr(site);
+  return site;
+}
+
+llvm::CallBase* JeandleIntrinsicLowering::emit_runtime_call(const JeandleIntrinsicDescriptor& desc,
+                                                            const JeandleIntrinsicEntrypoint& entry,
+                                                            llvm::ArrayRef<llvm::Value*> args) {
+  return emit_callsite(desc, entry.callee, entry.calling_conv, args, &entry);
+}
```
taofengliu: Delete this wrapper, we can use emit_callsite directly

src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
```diff
@@ -0,0 +1,607 @@
...
+  attach_java_klass_ret_attr(call,
+                             _target->signature()->return_type(),
+                             *_interp->_context);
+}
+
+llvm::CallBase* JeandleIntrinsicLowering::emit_callsite(const JeandleIntrinsicDescriptor& desc,
+                                                        llvm::FunctionCallee callee,
+                                                        llvm::CallingConv::ID calling_conv,
+                                                        llvm::ArrayRef<llvm::Value*> args,
+                                                        const JeandleIntrinsicEntrypoint* entry) {
```
```suggestion
llvm::CallBase* JeandleIntrinsicLowering::emit_callsite(JeandleIntrinsicCallInfo& desc,
                                                        llvm::CallingConv::ID calling_conv,
                                                        llvm::ArrayRef<llvm::Value*> args) {
```

taofengliu (approved):

PR #446, opened 2026-04-24
fix: Tier2 test case 

lfeng14:
## Related issue(s):

issue #

fix tier2 testcase:
- certificate expired


rcjhd:
Jeandle's baseline is out of date with community. These testcases maybe also failed for C2. 
I think it's not necessary for us to fix these "Jeandle not-related" problems because community has fixed that, such as
https://github.com/openjdk/jdk21u/commit/0da7a731e5dedfb7d248b74283339769080c08e0
https://github.com/openjdk/jdk21u/commit/f3112ef86c0b7601c9c7cd39b843aa5ac2f5d0e6

Issue #453, opened 2026-04-30
Normal flow branches to exception handler cause PHI incoming value missing


lfeng14:
**Problem:**                                                                                                                                                                                                                                 
                                                          
Jeandle abstract interpreter assumes that exception handler blocks are only reached via exception paths (i.e., throw_exception()). However, .jasm can construct bytecode where normal flow instructions (goto/if_icmp) directly branch to a  handler BCI, which breaks this assumption.
```
compile_java_method()
    1. JeandleAbstractInterpreter interpret(...)   ← Generate LLVM IR (PHI lacks incoming at this point, but no error reported)
    2. llvm::verifyModule(...)   ← Fails here! LLVM verifier detects that the number of PHI incoming values != number of predecessors
```
```
// test/hotspot/jtreg/compiler/parsing/TestExceptionBlockWithPredecessors.jasm
Method test:"()V"
	stack 4 locals 2
{
		try t0, t1;
		aload_0;
		invokevirtual	Method dontInlineThrow:"()V";
                ...
		putfield	Field loopCounter:"I";
		goto	L32;
		endtry t0, t1;
	L67:	stack_frame_type same;
		goto	L74;
		catch t0 java/lang/RuntimeException;
		stack_frame_type stack1;
		stack_map class java/lang/RuntimeException;
		astore_1;
		aload_1;
		aload_0;
		getfield	Field loopCounter:"I";
		bipush	10;
		if_icmpge	L70;                               // <--- here !
        aload_0;
        dup;
		getfield	Field loopCounter:"I";
		iconst_1;
		iadd;
		putfield	Field loopCounter:"I";
		goto L70;                                         // <--- here !
		catch t1 java/lang/Throwable;
	L70:    stack_frame_type stack1;
		stack_map class java/lang/Throwable;
		astore_1;
	L74:	stack_frame_type same;
		return;
}
```
**Root Cause:**

In interpret_block(), the successor loop skips merging VMState for handler blocks:
```
for (JeandleBasicBlock* suc : block->successors()) {
    // Don't update handlers' VM state here. They are updated by exception throwers.
    if (!suc->is_exception_handler() && !suc->merge_VM_state_from(...)) { ... }
}
```
When a normal flow branch (goto/if_icmp) targets a handler block, the handler is listed as a successor but its VMState merge is skipped due to is_exception_handler(). This leaves the handler's PHI nodes missing incoming values from normal
flow predecessors, causing LLVM IR verification failure / crash.

[...]

PR #454, opened 2026-04-30
fix: normal flow branches to exception handler cause PHI incoming value missing

lfeng14:
## Related issue(s):

issue #453

### DIFF
--- a/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
+++ b/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
@@ -1310,9 +1310,7 @@ void JeandleAbstractInterpreter::merge_into_exception_handler(JeandleBasicBlock*
   // Exception handlers expect a stack with one exception oop. Normal flow branches
   // (goto/if/switch) have an empty stack at the branch point. Push a null placeholder
   // to match the stack depth when merging from normal flow.
-  JeandleVMState* adjusted_state = _jvm->copy(true /* clear_stack */);
-  adjusted_state->apush(llvm::ConstantPointerNull::get(
-      llvm::cast<llvm::PointerType>(JeandleType::java2llvm(BasicType::T_OBJECT, *_context))));
+  JeandleVMState* adjusted_state = _jvm->copy(false /* clear_stack */);
   if (!handler_block->merge_VM_state_from(adjusted_state, _ir_builder.GetInsertBlock(), _method)) {
     JEANDLE_ERROR_ASSERT_AND_RET_VOID_ON_FAIL(false, "failed to merge VM state into exception handler block from normal flow");
   }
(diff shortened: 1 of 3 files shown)


src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1298,11 +1306,27 @@ void JeandleAbstractInterpreter::fcmp(BasicType type, bool true_if_unordered) {
...
 }
 
+void JeandleAbstractInterpreter::merge_into_exception_handler(JeandleBasicBlock* handler_block) {
+  // The bytecode verifier guarantees that all paths reaching the same BCI have
+  // consistent stack depth and types. So a normal flow branch (goto/if) targeting
+  // a handler block has the same stack layout as the exception path. We just need
+  // to merge the current VMState into the handler, bypassing the is_exception_handler()
+  // skip in the successor loop of interpret_block().
+  JeandleVMState* adjusted_state = _jvm->copy();
+  if (!handler_block->merge_VM_state_from(adjusted_state, _ir_builder.GetInsertBlock(), _method)) {
```
rcjhd, line 1316: Is this correct? I see jeandle has a function **copy_for_exception_handler(exception_oop)** to push a exception_oop, do you need to do so here? Another question, GetInsertBlock() should have a SetInsertPoint firstly, I didn't see that.

lfeng14: Thanks for your review!

For the first question:
We don’t need to call copy_for_exception_handler(exception_oop) here, because the manually emitted bytecode has already pushed the exception class reference onto the stack. Previously, pushing a null pointer manually also worked. From my understanding, as long as the stack is not empty, it is sufficient.
https://github.com/jeandle/jeandle-jdk/blob/7af20fcb39afbe8d8c89a097fe29733347a24022/test/hotspot/jtreg/compiler/parsing/TestExceptionBlockWithPredecessors.jasm#L73-L83

For the second question:
My understanding is as follows:
1. JandleAbstractInterpreter first splits the bytecode into Jandle basic blocks based on control-flow instructions like goto and ifcmp. Then it traverses each Jandle block in reverse post-order and parses them one by one in interpret_block.
2. Merging vmstate is an action between different jeandle blocks. Only goto, ifcmp, and createinvoke instructions may require merging vm state.
   - For invoke-related instructions (such as check_cast), which generate LLVM basic block branches, we first merge the vm state, then create the LLVM basic block and setInsertPoint.

https://github.com/jeandle/jeandle-jdk/blob/7af20fcb39afbe8d8c89a097fe29733347a24022/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp#L1560-L1569

In summary: we are already inside the predecessor Jandle block and its corresponding LLVM basic block of the exception handler. So there is no need to call SetInsertPoint again here.

taofengliu (commented):
_tableswitch and _lookupswitch may also need `merge_into_exception_handler`

lfeng14:
> _tableswitch and _lookupswitch may also need `merge_into_exception_handler`

Thanks for the suggestion, updated the code

taofengliu (approved):

taofengliu:
@mergifyio queue

taofengliu:
@mergifyio queue

PR #456, opened 2026-05-06
feat: Implement inline-tree based method inlining

minchao1024:
## Related issue(s):

issue #428 #528

Implement Jeandle inline support around an explicit JVM-side inline tree, and
wire the LLVM inliner callbacks to JVM inline policy, replay data, and debug
scope reconstruction.

- Add JeandleInlineTree and JeandleParseContext to track root/inlinee method
  scopes, successful inline history, and per-scope inline policy decisions.
- Wire LLVM VM callbacks for fetching inline callee IR, checking whether a
  call is OK to inline, recording successful inlines, and dumping inline replay
  IR after inlining completes.
- Align Jeandle inline policy with the main C2 InlineTree checks, including
  force/dont inline commands, ciReplay inline decisions, size/depth limits,
  accessor handling, recursive inline limits, and a Jeandle node-count cutoff.
- Extend deopt bundle and stackmap parsing to support inlined Java scopes,
  including method markers, duplicated BCI markers, chained scope values, and
  nested DebugInformationRecorder scopes.
- Move oop constant handling to JeandleCompilation so root and inlinee parsers
  share module-level oop globals, and attach Java method/class metadata needed
  by LLVM passes.
- Support ciReplay inline data for Jeandle inline decisions, and dump Jeandle
  inline trees so replay can preserve successful inline scopes.
- Add Jeandle replay dump support for Jeandle inline data and inline callee IR modules,
  plus JeandlePrintInlineTree for debugging compiled inline trees.
- Update Jeandle template/runtime helper attributes for leaf/nounwind behavior
  and Java method/accessor marking across supported architectures.

TODO:
- Jeandle currently only tries to inline monomorphic call sites. Broader virtual
  call inlining depends on future devirtualization analysis.
- BasicLock slots are still function-local and cannot be reused across caller
  and inlinee methods yet. This needs a later LLVM pass to hoist or merge lock
  slots after inlining.
- Recursive inlining of the root method is not supported. Recursive calls among
  non-root inlinee methods are supported and limited by the inline policy.


minchao1024:
Here is a small case:
```
// java -Xcomp -XX:-TieredCompilation -XX:CompileCommand=compileonly,TestInline::testInline -XX:CompileCommand=printassembly,TestInline::testInline -XX:CompileCommand=inline,TestInline::print TestInline
class TestInline {
    public static Object testInline(Object a) {
      print(a);
      return a;
    }
    public static Object print(Object b) {
      System.out.println(b);
      return b;
    }
  public static void main(String[] args) {
    testInline("Hello, World!");
  }
}
```
```
pc-bytecode offsets:
PcDesc(pc=0x000076cf5c86d79f offset=ffffffff bits=0):
PcDesc(pc=0x000076cf5c86d7c3 offset=23 bits=1):
   TestInline::print@0 (line 7)  reexecute=true
   Locals
    - l0: stack[0],oop
   TestInline::testInline@1 (line 3)
   Locals
    - l0: stack[0],oop
PcDesc(pc=0x000076cf5c86d7e5 offset=45 bits=0):
   TestInline::testInline@-1 (line 3)
PcDesc(pc=0x000076cf5c86d809 offset=69 bits=0):
```

src/hotspot/cpu/aarch64/templatemodule/template_aarch64.ll
```diff
@@ -26,13 +26,13 @@
 ; ==============================================================================
 
 ; Get the stack pointer
-define hotspotcc i64 @jeandle.get_stack_pointer() "lower-phase"="0" {
+define hotspotcc i64 @jeandle.get_stack_pointer() "lower-phase"="0" #0 {
```
taofengliu, line 29: These functions are inlined in `lower-phase=0`, so the `RewriteStatepointForGC` pass will not see them. Will the `gc-leaf-function` attribute work?

src/hotspot/share/jeandle/jeandleVMCallback.cpp
```diff
@@ -87,6 +91,47 @@ bool jeandle_is_effectively_final(uintptr_t klass_ptr) {
   return false;
 }
 
+bool jeandle_should_inline(uintptr_t caller_name, uintptr_t callee_name) {
```
taofengliu: caller_name is not used

src/hotspot/share/jeandle/jeandleCompilation.hpp
```diff
@@ -94,13 +107,16 @@ class JeandleCompilation : public StackObj {
...
   ciMethod* _method;
+  ciMethod* _inlinee;
   const std::string _name;
   int _entry_bci;
   std::unique_ptr<llvm::LLVMContext> _context;
   std::unique_ptr<llvm::Module> _llvm_module;
   std::string _comp_start_time;
   uint _trap_hist[MethodData::_trap_hist_limit];
 
+  llvm::StringMap<ciMethod*> _inline_candidates;
```
taofengliu: Maybe we don't need this, all Java method can be inline candidates.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -2936,7 +2979,29 @@ void JeandleAbstractInterpreter::dispatch_exception_to_handler(llvm::Value* exce
...
+      // to signal that an exception occurred here. During inlining, this resume
+      // will be replaced with a branch to the caller's landing pad. Since all
+      // exception data is stored in and loaded from TLS (exception_oop), the
+      // dummy resume value is never consumed and causes no issues.
+      llvm::Type* landingpad_result_type = llvm::Type::getInt64Ty(*_context);
+      llvm::Value* dummy = llvm::ConstantInt::get(landingpad_result_type, 0);
+      _ir_builder.CreateResume(dummy);
+    } else {
+      _ir_builder.CreateResume(landingpad);
+    }
```
taofengliu, line 3171: According to your comment, can we always use `_ir_builder.CreateResume(dummy)`?
```suggestion
    // For the unwind of athrow bytecode, resume passes a dummy value solely
    // to signal that an exception occurred here. During inlining, this resume
    // will be replaced with a branch to the caller's landing pad. Since all
    // exception data is stored in and loaded from TLS (exception_oop), the
    // dummy resume value is never consumed and causes no issues.
    llvm::Type* landingpad_result_type = llvm::Type::getInt64Ty(*_context);
    llvm::Value* dummy = llvm::ConstantInt::get(landingpad_result_type, 0);
    _ir_builder.CreateResume(dummy);
```

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -222,9 +222,28 @@ void JeandleVMState::store(BasicType type, int index, llvm::Value* value) {
...
+#ifdef ASSERT
+  if (log_is_enabled(Trace, jeandle)) {
+    tty->print_cr("Build deopt bundle at bci %d :", bci);
+  }
+#endif
+
   llvm::SmallVector<llvm::Value*> args;
-  // |--- bci ---|--- locals ---|--- stack ---|--- monitor ---|--- orig_pc ---|
+  // Inliner:                |--- bci ---|--- locals ---|--- stack ---|--- monitor ---|--- orig_pc ---|
+  // Inlinee: |--- method ---|--- bci ---|--- locals ---|--- stack ---|--- monitor ---|--- orig_pc ---|
```
taofengliu: Add comment to describe how deopt_args will looks like after inline

src/hotspot/share/jeandle/jeandleCompiledCode.cpp
```diff
@@ -323,7 +334,13 @@ void JeandleCompiledCode::resolve_reloc_info(JeandleAssembler& assembler) {
...
       }
       if (call_info) {
-        relocs.push_back(new JeandleCallReloc(inst_end_offset, _env, _method, parse_stackmap(stackmaps, record, call_info), call_info));
+        auto location = record->location_begin();
+        int num_deopts = parse_stackmap_prologue(record, location);
+        JeandleCallReloc* reloc = new JeandleCallReloc(inst_end_offset, _env, _method, call_info);
+        do {
+          reloc->add_stack_map(parse_stackmap(stackmaps, record, location, call_info, num_deopts));
+        } while (location != record->location_end());
+        relocs.push_back(reloc);
```
taofengliu, line 408: Add comment to describe the format of stack map and how we handle it.

src/hotspot/share/jeandle/jeandleCompiledCode.cpp
```diff
@@ -529,27 +547,49 @@ JeandleStackMap* JeandleCompiledCode::parse_stackmap(StackMapParser& stackmaps,
...
+}
+
+JeandleStackMap* JeandleCompiledCode::parse_stackmap(StackMapParser& stackmaps,
+                                                     StackMapParser::record_iterator& record,
+                                                     StackMapParser::RecordAccessor::location_iterator& location,
+                                                     CallSiteInfo* call_info,
+                                                     int& num_deopts) {
   bool reexecute = false;
+  int bci = -1;
+  ciMethod* current_method =nullptr;
```
```suggestion
  ciMethod* current_method = nullptr;
```

src/hotspot/share/jeandle/jeandleVMCallback.cpp
```diff
@@ -87,6 +91,47 @@ bool jeandle_is_effectively_final(uintptr_t klass_ptr) {
...
+  if (callee_func != nullptr && !callee_func->isDeclaration()) {
+    return true;
+  }
+  ciMethod* callee = comp->get_inline_candidate((const char*)callee_name);
+  if (callee == nullptr) {
+    return false;
+  }
+  {
+    SetInlinee inlinee_guard(callee);
+    JeandleAbstractInterpreter interpret(callee, -1, *M, *comp->compiled_code(), comp->trap_hist());
```
taofengliu: Should we share JeandleAbstractInterpreter::_oops between different interpreters?

taofengliu (commented):
When compiling a function, we perform some stack space allocations in the function's entry block, which are folded into the prologue code as StaticAllocas. However, once this function gets inlined, is it possible that these stack space allocations can no longer be folded?

taofengliu (commented):
During compilation, we reuse the stack lock slot of monitors. In the case of inlining, is this optimization still applicable?

src/hotspot/share/jeandle/jeandleVMCallback.cpp
```diff
@@ -87,6 +90,58 @@ bool jeandle_is_effectively_final(uintptr_t klass_ptr) {
...
+    return false;
+  }
+
+  // LLVM calls this callback only after the inline transformation succeeds.
+  // Keep Jeandle's inline tree in sync at that point so later policy checks
+  // account for the bytecodes that were actually inlined.
+  return comp->build_inline_tree_for_callee(scope_id, bci, callee) != nullptr;
+}
+
+bool jeandle_get_inline_callee_ir(uintptr_t callee_name) {
```
taofengliu: This callback will modify IR. So it cannot be replayed through callback log. Can we make it able to replay?

src/hotspot/cpu/aarch64/templatemodule/template_aarch64.ll
minchao1024: This attribute is not required for correctness; adding it can make the semantics more complete.

src/hotspot/share/jeandle/jeandleVMCallback.cpp
minchao1024: Fixed

src/hotspot/share/jeandle/jeandleCompilation.hpp
minchao1024: Removed already.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
minchao1024: Both can run using `dummy landingpad`, but for the `resume` of `invoke`, using the previous landingpad is better. Firstly, the semantics are more complete, and secondly, even when using `dummy landingpad` here, the previous `landingpad` cannot be omitted.

src/hotspot/share/jeandle/jeandleCompiledCode.cpp
minchao1024: Done

src/hotspot/share/jeandle/jeandleVMCallback.cpp
minchao1024: Yes, Fixed.

src/hotspot/share/jeandle/jeandleVMCallback.cpp
minchao1024: Already supported.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.hpp
```diff
@@ -324,25 +327,24 @@ class JeandleAbstractInterpreter : public StackObj {
...
-  llvm::SmallVector<llvm::Value*> _allocated_basic_lock;
+  // Reuse BasicLock stack slots within the current LLVM function: each monitor
+  // nesting level maps to a fixed slot. Inline currently cannot reuse BasicLock
+  // slots across caller/callee functions because the slot is a function-local
+  // alloca. A later LLVM pass should hoist or merge these slots after inlining.
+  llvm::SmallVector<llvm::Value*> _basic_lock_slots;
 
-  bool need_alloc_for(int monitor_nest_level) {
-    return (int)_allocated_basic_lock.size() <= monitor_nest_level;
+  bool needs_basic_lock_slot(int monitor_nest_level) {
```
taofengliu:
```suggestion
  bool needs_new_basic_lock_slot(int monitor_nest_level) {
```

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -224,10 +225,39 @@ void JeandleVMState::store(BasicType type, int index, llvm::Value* value) {
...
+  if (log_is_enabled(Trace, jeandle)) {
+    tty->print_cr("Build deopt bundle at bci %d :", bci);
+  }
+#endif
+
   llvm::SmallVector<llvm::Value*> args;
-  // |--- bci ---|--- locals ---|--- stack ---|--- monitor ---|--- orig_pc ---|
+  // Total   deopt bundle: [Root deopt bundle] + [Inlined deopt bundle] + [Inlined deopt bundle] + ...
+  // Root    deopt bundle:                |--- bci ---|--- bci ---|--- locals ---|--- stack ---|--- monitor ---|--- orig_pc ---|
+  // Inlined deopt bundle: |--- method ---|--- bci ---|--- bci ---|--- locals ---|--- stack ---|--- monitor ---|--- orig_pc ---|
```
taofengliu, line 246:
```suggestion
  // Root    deopt bundle:                |--- bci ---|--- locals ---|--- stack ---|--- monitor ---|--- orig_pc ---|
  // Inlined deopt bundle: |--- method ---|--- bci ---|--- locals ---|--- stack ---|--- monitor ---|--- orig_pc ---|
```
We can give bci a specific `DeoptValueEncoding`, then we don't need to repeat it. And we can move `DeoptValueEncoding` related code to LLVM side to help passes analyze deopt bundles, #500 does this, too.
This will make the format of deopt argument more uniform.

minchao1024: Our current deoptimization bundle design necessitates traversal starting from the beginning; while we typically only care about the deoptimization bundle for the current method, inlining places that bundle at the very end, requiring us to iterate through the entire sequence. Consequently, instead of using a marker, I use repeated BCI values. This specific design not only allows for the rapid retrieval of the current BCI via post-order traversal but also makes it intuitive to locate the BCI within the IR.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.hpp
minchao1024: Fixed

src/hotspot/share/jeandle/jeandleReloc.cpp
```diff
@@ -90,30 +90,45 @@ int JeandleCallReloc::inst_end_offset() {
...
+    recorder->end_safepoint(inst_end_offset());
+    return;
+  }
+
+  if (_call->is_method_handle_invoke()) {
+    // nmethod requires a DeoptMH handler exactly when at least one emitted
+    // PcDesc is marked as a method-handle invoke. Set the flag here, where the
+    // final post-LLVM-inlining debug info is actually recorded.
+    JeandleCompilation::current()->compiled_code()->set_has_method_handle_invoke(true);
+  }
```
taofengliu, line 111: Maybe we should only set it when the inlined final call instruction is method handle invoke.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.hpp
taofengliu: Is it unused? We can delete it, and related functions.

src/hotspot/share/jeandle/jeandleCompiledCode.cpp
```diff
@@ -578,25 +628,45 @@ JeandleStackMap* JeandleCompiledCode::parse_stackmap(StackMapParser& stackmaps,
...
-    assert(this->_method != nullptr, "must be method compilation");
+    assert(current_method != nullptr, "must be method compilation");
 
     // bci goes first in deopt operands
-    int bci = (location++)->getSmallConstant();
-    num_deopts--;
+    bci = (location++)->getSmallConstant();
+    guarantee(bci == (int)((location++)->getSmallConstant()), "duplicated bci must match");
+    num_deopts -= 2;
     call_info->set_bci(bci);
```
taofengliu, line 653: Since we already have bci in deopt bundle, we don't need bci in CallSiteInfo anymore, we can delete it.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -2225,23 +2327,7 @@ llvm::InvokeInst* JeandleAbstractInterpreter::call_java_op_ex(llvm::StringRef ja
 
 llvm::OperandBundleDef JeandleAbstractInterpreter::create_current_deopt_bundle() {
   ensure_orig_pc_slot();
```
taofengliu, line 2486: Maybe ersure_orig_pc_slot() should only make effect in root compilations. We can make ersure_orig_pc_slot as a noop in non-root compilation

src/hotspot/share/jeandle/jeandleReloc.cpp
minchao1024: The `_call` here already represents the final emitted callsite after inlining, not the inline callsite that was eliminated

src/hotspot/share/jeandle/jeandleAbstractInterpreter.hpp
minchao1024: Done

src/hotspot/share/ci/ciEnv.cpp
```diff
@@ -1619,22 +1622,30 @@ void ciEnv::find_dynamic_call_sites() {
...
-    }
-#endif
     Method* method = task->method();
     int entry_bci = task->osr_bci();
     int comp_level = task->comp_level();
+    if (ReplayReduce && compiler_data() != nullptr) {
+      if (task->compiler()->is_jeandle()) {
+#ifdef JEANDLE
+        // Dump Jeandle "reduced" inlining data.
+        ((JeandleCompilation*)compiler_data())->dump_inline_data_reduced(out);
```
taofengliu, line 1632: Do we need to clear compiler_data in the destruction of JeandleCompilation to avoid dereferencing a destructed JeandleCompilation?

taofengliu (approved):

Issue #463, opened 2026-05-12
TLAB fast-path enhancements — Style 1 prefetch & array allocation


eapenzhan:
**Goals**
Extend the existing TLAB instance fast path (landed in [#269](https://github.com/eapenzhan/jeandle-jdk/issues/269)) with two additions:

1. **Style 1 allocation prefetch (`AllocatePrefetchStyle == 1`)** — emit `AllocatePrefetchLines / AllocateInstancePrefetchLines` write prefetches at `tlab_top + AllocatePrefetchDistance + i * AllocatePrefetchStepSize` after a successful bump-pointer, before object initialization. Both `jeandle.new_instance` and `jeandle.newarray` benefit.
2. **Array TLAB fast path** — `jeandle.newarray` upgraded from always-slow-path to inline TLAB bump-pointer with header/length stores, payload zeroing, prefetch, and release publication. Falls back to `JeandleRuntimeRoutine::new_array` on TLAB miss, negative length, or length exceeding the per-element-type cap.
Style 2 (watermark-based prefetch), G1 support, compressed class pointers, and multidimensional array fast path are explicitly out of scope and tracked as follow-ups.

**Motivation**

`new T()` already hits the inline TLAB fast path after [#269](https://github.com/eapenzhan/jeandle-jdk/issues/269). Two gaps remain in compiled allocation throughput:

- **Arrays (`new int[N], new T[N]`)** still go to the runtime on every call, even when the TLAB has capacity. Array allocation is a hot path in workloads that build collections, strings, or buffers, and the unconditional `JeandleRuntimeRoutine::new_array` call costs us a stub transition plus a JRT block on every allocation.
- **Allocation prefetch has no Jeandle equivalent yet**. C2 emits prefetches at allocation sites (`PhaseMacroExpand → PrefetchAllocation`) so that the next bump-pointer worth of TLAB memory is warm in L1 by the time the next allocation reaches it. Without this, every fast-path allocation pays a cache miss for the next object header on cold lines.
Adding both moves Jeandle's compiled allocation cost closer to C2 on Serial GC without touching any runtime allocation policy.

**Description**

**Compiler / runtime boundary (unchanged from [#269](https://github.com/eapenzhan/jeandle-jdk/issues/269))**
[...]

PR #269, opened 2025-12-31
feat: Implement the fast path of object allocation

kuaiwei:
## Related issue(s):

issue #263 

## What this PR does / why we need it:
fast path of object allocation.

PR #465, opened 2026-05-12
feat: Add TLAB fast path for array allocation and Style 1 prefetch

eapenzhan:
Extend the TLAB instance fast path from #269 with Array TLAB fast path & Style 1 allocation prefetch.

## Related issue(s):

issue #463 

### Compiler / runtime boundary (unchanged from #269)

The compiler emits only the existing-TLAB bump-pointer path. The runtime owns TLAB refill, allocation outside TLAB (humongous, slow allocator), class initialization fallback, and all exception construction (`NegativeArraySizeException`, `OutOfMemoryError`). The slow path remains the semantic source of truth.

### Array fast path

`jeandle.newarray` is invoked with `(klass, length, size_in_bytes, base_offset, max_length)`. The frontend (`do_unified_newarray` in `jeandleAbstractInterpreter.cpp`) decodes the array klass at IR build time using `Klass::layout_helper()`:

- `log2_element_size = Klass::layout_helper_log2_element_size(lh)`
- `base_offset = arrayOopDesc::base_offset_in_bytes(element_type)`
- `max_length = arrayOopDesc::max_array_length(element_type)`
- `size_in_bytes = align((length << log2_element_size) + base_offset, MinObjAlignmentInBytes)` — computed in IR. When `length` is a compile-time constant (the common case), LLVM folds shift + add + and into the constant size at the call site, so the bump-pointer is branch-free in the hot path.

The JavaOp body then performs:

1. **Pre-checks**: branch to slow path if `length < 0` or `(unsigned) length > max_length`. A single unsigned compare covers both negative-as-large-unsigned and the per-element-type overflow cap. We never throw from IR — the runtime is responsible for constructing the right exception with the right stack.
2. **Bump-pointer**: load `JavaThread.tlab_top` / `tlab_end`, compute `new_top`, compare against `tlab_end`, branch to slow path on miss, store `new_top` back to `tlab_top`.
3. **Prefetch** (Style 1, see next section).
4. **Header / length / body**:
   - store mark word (`markWord::prototype()`)
   - store full 64-bit klass pointer
   - store `length`
   - `memset` payload (skipped under `ZeroTLAB`)
5. **Publish** with `fence release`.

The store ordering matches HotSpot C1 (`c1_MacroAssembler_aarch64.cpp:181-199`). `length` must be stored before body zeroing, because the `memset` length is derived from it.

Under Serial GC the single `fence release` is sufficient — GC is STW, so partial-init objects are never externally visible.

### Style 1 prefetch

Both `jeandle.new_instance` and `jeandle.newarray` gain a prefetch block immediately after the bump-pointer store. The block is gated at the IR level on `@VMOptions.AllocatePrefetchStyle == 1`. When enabled, it emits `N = AllocatePrefetchLines` (or `AllocateInstancePrefetchLines` for instance allocation) write prefetches at addresses `tlab_top + AllocatePrefetchDistance + i * AllocatePrefetchStepSize`.

Prefetch addresses are allowed to overrun `tlab_end` or unmapped memory. AArch64 `prfm` and x86 `prefetcht0` are hardware hints and do not fault — this is the same property C2 relies on in `macro.cpp:1841-1862`.

The Jeandle-LLVM optimization pipeline at `OptimizationLevel::O3` (set in `Jeandle.h`) includes `LoopFullUnrollPass` via `buildPerModuleDefaultPipeline → buildModuleSimplificationPipeline → buildFunctionSimplificationPipeline` (`PassBuilderPipelines.cpp:528`, unconditional outside SamplePGO). Combined with `IPSCCP` folding loads of constant globals, the prefetch loop with a constant trip count fully unrolls into N straight `@llvm.prefetch` calls before phase=1 inlining.

### Template globals added

All marked `setConstant(true)` so IPSCCP folds loads into immediates:

- `@VMOptions.AllocatePrefetchStyle`
- `@VMOptions.AllocatePrefetchDistance`
- `@VMOptions.AllocatePrefetchStepSize`
- `@VMOptions.AllocatePrefetchLines` (array path)
- `@VMOptions.AllocateInstancePrefetchLines` (instance path)
- `@arrayOopDesc.length_offset`

Existing globals from #269 (`@VMOptions.UseTLAB`, `@VMOptions.ZeroTLAB`, `@JavaThread.tlab_top_offset`, `@JavaThread.tlab_end_offset`, mark/klass offsets) are reused.

### Out of scope(tracked as follow-ups)

- Style 2 watermark prefetch (needs tlab_pf_top state),
- G1 support (needs SATB pre-barrier at publish),
- UseCompressedClassPointers (cross-cutting Jeandle change),
- multianewarray fast path.

### DIFF
--- a/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
+++ b/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
@@ -38,6 +38,7 @@
 #include "ci/ciObjArrayKlass.hpp"
 #include "ci/ciSymbols.hpp"
 #include "ci/ciTypeFlow.hpp"
+#include "oops/arrayOop.hpp"
 #include "oops/objArrayKlass.hpp"
 #include "classfile/javaClasses.hpp"
 #include "interpreter/interpreter.hpp"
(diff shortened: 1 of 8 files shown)


eapenzhan:
This PR will be seperate by 2 PRs:
main ──── ●           
                             ╲       
                                 ●  feat/array-fast-path  ←── PR A base = main
                                     ╲     
                                         ●  feat/style1-prefetch ←── PR B base = feat/array-fast-path

PR #470, opened 2026-05-15
feat: Add TLAB fast path for array allocation

eapenzhan:
Extend the TLAB instance fast path from #269 with Array TLAB fast path

## Related issue(s):

issue #463 

The compiler emits only the existing-TLAB bump-pointer path. The runtime owns TLAB refill, allocation outside TLAB (humongous, slow allocator), class initialization fallback, and all exception construction (`NegativeArraySizeException`, `OutOfMemoryError`). The slow path remains the semantic source of truth.

### Array fast path

`jeandle.newarray` is invoked with `(klass, length, size_in_bytes, base_offset, max_length)`. The frontend (`do_unified_newarray` in `jeandleAbstractInterpreter.cpp`) decodes the array klass at IR build time using `Klass::layout_helper()`:

- `log2_element_size = Klass::layout_helper_log2_element_size(lh)`
- `base_offset = arrayOopDesc::base_offset_in_bytes(element_type)`
- `max_length = arrayOopDesc::max_array_length(element_type)`
- `size_in_bytes = align((length << log2_element_size) + base_offset, MinObjAlignmentInBytes)` — computed in IR. When `length` is a compile-time constant (the common case), LLVM folds shift + add + and into the constant size at the call site, so the bump-pointer is branch-free in the hot path.

The JavaOp body then performs:

1. **Pre-checks**: branch to slow path if `length < 0` or `(unsigned) length > max_length`. A single unsigned compare covers both negative-as-large-unsigned and the per-element-type overflow cap. We never throw from IR — the runtime is responsible for constructing the right exception with the right stack.
2. **Bump-pointer**: load `JavaThread.tlab_top` / `tlab_end`, compute `new_top`, compare against `tlab_end`, branch to slow path on miss, store `new_top` back to `tlab_top`.
3. **Prefetch** ：Not Implement
4. **Header / length / body**:
   - store mark word (`markWord::prototype()`)
   - store full 64-bit klass pointer
   - store `length`
   - `memset` payload (skipped under `ZeroTLAB`)
5. **Publish** with `fence release`.

The store ordering matches HotSpot C1 (`c1_MacroAssembler_aarch64.cpp:181-199`). `length` must be stored before body zeroing, because the `memset` length is derived from it.

### DIFF
--- a/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
+++ b/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
@@ -38,6 +38,7 @@
 #include "ci/ciObjArrayKlass.hpp"
 #include "ci/ciSymbols.hpp"
 #include "ci/ciTypeFlow.hpp"
+#include "oops/arrayOop.hpp"
 #include "oops/objArrayKlass.hpp"
 #include "classfile/javaClasses.hpp"
 #include "gc/shared/gc_globals.hpp"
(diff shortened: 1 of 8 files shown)


eapenzhan:
Measured with the JDK microbenchmark suite (`test/micro/org/openjdk/bench/vm/gc/Alloc.java`)
  on AArch64 (fastdebug, `-XX:+UseJeandleCompiler -XX:-TieredCompilation`, JMH `-f 2 -wi 3 -i 5`). The array fast path is gated by `UseTLAB`: with `+UseTLAB` (default) the compiled
  `jeandle.new_array` takes the inline bump-pointer fast path; `-XX:-UseTLAB` forces the runtime slow path. Same build for both columns. Lower is better.

  | Benchmark                | Fast path (ns/op) | Slow path (ns/op) | Speedup |
  | ------------------------ | ----------------: | ----------------: | ------: |
  | `testSmallConstArray`    |     5,060 ±   386 |   184,864 ± 2,227 |  36.5×  |
  | `testSmallVariableArray` |     5,707 ±   225 |   185,403 ± 2,793 |  32.5×  |
  | `testSmallObject`        |    13,119 ±   130 |   216,001 ± 3,764 |  16.5×  |
  | `testLargeConstArray`    |  73,835 ± 15,615 |   235,747 ± 4,778 |   3.2×  |
  | `testLargeVariableArray` |    60,360 ± 1,875 |   240,244 ± 7,585 |   4.0×  |

  Small array allocations — the fast path's target case — are ~30–36× faster.

  **Methodology note:** `-XX:-UseTLAB` disables TLAB globally, so it also slows the runtime slow path and alters GC behavior; these numbers therefore reflect the compiled fast path *together with* TLAB-assisted allocation, not the fast path in isolation. The large-array cases exceed `FastAllocateSizeLimit` and take the slow path under both configurations, so their delta is mostly the TLAB effect on the runtime path rather than the new fast path.

Issue #462, opened 2026-05-11
JVM crash in stackmap parser when LLVM merges multiple uncommon_trap calls into a single statepoint


lfeng14:
## Reproduction

```bash
./build-test.sh ./test/jdk/java/net/httpclient/ShortResponseBodyPost.java fastdebug
```

## Crash Details

**Exit code**: 134 (SIGABRT)

**Assertion**:
```
assert(basic_type >= 0 && basic_type <= BasicType::T_ILLEGAL) failed
```

**Location**: `src/hotspot/share/jeandle/jeandleCompiledCode.hpp:81` in `DeoptValueEncoding::decode()`

**Crashing method example**: `java.lang.StringLatin1.charAt`

The stackmap parser in `JeandleCompiledCode::parse_stackmap()` calculates `num_deopts` based on the **total** number of locations in the record, which includes all merged groups. It reads the first group correctly:

1. Reads `bci = 2` ✓
2. Reads `DeoptValueEncoding` pairs correctly ✓
3. Finishes the first group, but `num_deopts > 0` because the total count includes the second group

It then continues reading into the second group, interpreting the next value (`807`, a `bci`) as a `DeoptValueEncoding`:

```
encode = 807
basic_type = 807 & 0xffff = 807    ← far exceeds T_ILLEGAL (~15)
value_type = (807 >> 16) & 0xffff = 0
```

The assertion `basic_type >= 0 && basic_type <= BasicType::T_ILLEGAL` in `DeoptValueEncoding::decode()` fails, causing JVM abort.

## Root Cause
- In short, LLVM’s `InlineFunction` concatenates `deopt` operand bundles when inlining: the caller’s deopts are prepended to each callee `deopt` bundle, breaking `DeoptValueEncoding::decode()` and causing `parse_stackmap()` to crash. 
  - generate the LLVM IR for the ShortResponseBodyPost.java test case without running any optimization passes
    ```
    BOOT_JDK=/usr/ JT_HOME=/usr/share/jtreg JDK_IMAGE_DIR="${BASE_DIR}/images/jdk/" TEST_IMAGE_DIR="${BASE_DIR}/images/test/" make test-prebuilt TEST="$TEST_NAME" \
[...]

PR #472, opened 2026-05-18
fix: stackmap parser crash caused by merged deopt bundles in inlined Java uncommon traps

lfeng14:
## Related issue(s):

issue #462 

plz refer #462

### DIFF
--- a/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
+++ b/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
@@ -1860,6 +1860,27 @@ void JeandleAbstractInterpreter::invoke() {
   llvm::InvokeInst* invoke = _ir_builder.CreateInvoke(callee, dispatched._normal_dest, dispatched._unwind_dest, args,
                                                       {create_current_deopt_bundle()});
 
+  // If the callee contains any calls with "deopt" operand bundles, inlining
+  // would cause the parent's deopt bundle to be prepended to the child's,
+  // breaking the deopt state encoding.  Prevent inlining in that case.
+  if (auto *callee_func = llvm::dyn_cast<llvm::Function>(callee.getCallee())) {
+    if (!callee_func->isDeclaration()) {
+      bool has_deopt = false;
+      for (auto &BB : *callee_func) {
+        for (auto &I : BB) {
+          if (auto *CB = llvm::dyn_cast<llvm::CallBase>(&I)) {
+            if (CB->getOperandBundle(llvm::LLVMContext::OB_deopt)) {
+              has_deopt = true;
+              break;
+            }
+          }
+        }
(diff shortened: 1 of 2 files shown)


src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1860,6 +1565,27 @@ void JeandleAbstractInterpreter::invoke() {
...
+  // would cause the parent's deopt bundle to be prepended to the child's,
+  // breaking the deopt state encoding.  Prevent inlining in that case.
+  if (auto *callee_func = llvm::dyn_cast<llvm::Function>(callee.getCallee())) {
+    if (!callee_func->isDeclaration()) {
+      bool has_deopt = false;
+      for (auto &BB : *callee_func) {
+        for (auto &I : BB) {
+          if (auto *CB = llvm::dyn_cast<llvm::CallBase>(&I)) {
+            if (CB->getOperandBundle(llvm::LLVMContext::OB_deopt)) {
+              has_deopt = true;
```
rcjhd: I understand that with this approach, we avoid inlining whenever the callee may deoptimize. Does this lead to a reduction in inlining opportunities?

lfeng14: The noinline restriction applies only to call sites whose callee has an LLVM IR definition and whose function body itself contains call bases with deopt bundles. This situation arises only in recursive Java‑method scenarios.
- `invoke()` Java methods  
  Because the Jeandle JIT compiler processes one Java method at a time and can recursively re‑enter itself, non‑empty deopt bundles on these calls cause problems.

- `call_java_op()` → `JavaOp`  
  - In `template.ll` (phase 0), there are no deopt bundles.  
  - In `DEF_JAVA_OP` (phase 1), some operations include bundled calls: `call_java_op("jeandle.safepoint_poll", ..., {bundle})`, `call_java_op_ex("jeandle.new_instance", ..., {bundle})`, `call_java_op_ex("jeandle.newarray", ..., {bundle})`, and `create_call_ex(multi_dim_array, ..., {bundle})`.  
  - Concatenating calls with empty bundles is harmless.

- Indirect routines  
  C++ functions that are only declared and have no IR body cannot be inlined.

- Direct routines  
  C function pointers or declarations without an IR body cannot be inlined.

- Assembly routines  
  Handwritten assembly with no IR body cannot be inlined.

rcjhd: I know that only `invoke()` is affected. However, it's just what our inline to do at most cases.

lfeng14: I believe the current solution does not cause performance degradation. The proper approach should be to implement handling for bundles when inlining Java method.

PR #486, opened 2026-05-23
fix: guard missing oop handle in reloc info

### LINKED ISSUE #481: StringMap::operator[] silently inserts null jobject when resolving relocInfo
Description: 

If `target_name` is not in `_oop_handles`, `operator[]` inserts a default `nullptr` jobject. This null handle is passed to `OopRecorder::find_index`, causing crashes during GC.

 Fix: 

Use `_oop_handles.find()` and check for existence before access.

qorexdevs:
﻿Fixes #481.

This replaces `_oop_handles[target_name]` with a `find()` guard in both relocInfo paths (`is_oop_reloc` and `is_oop_addr_reloc`).
When the handle is missing, Jeandle now reports a compilation error and returns early instead of inserting a null `jobject` and passing it to relocation recording.

I did not add a regression test here because this path is hard to trigger in a small standalone jtreg case without broader Jeandle runtime setup, and test addition was marked optional in the issue thread.


taofengliu (approved):
Good!

taofengliu:
@mergifyio queue

PR #491, opened 2026-05-24
enhance: keep the first compilation error msg during error reporting

### LINKED ISSUE #489: Error message overwritten on multiple errors during error reporting.
Description:
In jeandleCompilation.hpp, `report_error()` unconditionally overwrites `_error_msg`. Only the last error is preserved. But we should keep the first error we encounter, as it's more likely the root cause of the compilation failure.

Fix: 
Only set `_error_msg` if it is still `nullptr`.

Noai-oss:
## Related issue(s):

issue #489

Fixes #489

add `_error_msg == nullptr` to make sure the `_error_msg` is first error msg


taofengliu (approved):
Thanks for contribution!

taofengliu:
@mergifyio queue

PR #492, opened 2026-05-25
enhance: make `verifyModule` only for debug build and also check after ir optimization

### LINKED ISSUE #488: IR verification should only for debug builds, and should do post-optimization IR verification
Description:
In jeandleCompilation.cpp:304, `verifyModule()` is called without macro guard for debug build only. It is also only called before optimization but not after. Custom Jeandle passes could corrupt IR without detection.
 
Fix: 
Use macro guard to make IR verification for debug build only. Add `DEBUG_ONLY(verifyModule(*_llvm_module, &errs());)` after `llvm::jeandle::optimize()`.

Noai-oss:
## Related issue(s):

issue #488

fixes #488 

Make verifyModule and error handing into a code block under `DEBUG_ONLY` macro guard, and add the same check after ir optimization. Changes coverage two functions `JeandleCompilation` and `compile_java_method`.


taofengliu (approved):
Looks good! Thanks!

taofengliu:
@mergifyio queue

PR #71, opened 2025-09-09
feat: Add a test tool for IR checking

taofengliu:
## Related issue(s):

issue #70 

## What this PR does / why we need it:
Add a test tool like LLVM's FileCheck to help check content in dumped IR.
There is an example in test/hotspot/jtreg/compiler/jeandle/fileCheck/TestFileCheck.java

PR #497, opened 2026-05-26
feat: PGO from interpreter/C1 MDO

### LINKED ISSUE #496: Profile-guided optimization (PGO) from interpreter/C1 MDO
PGO for the Jeandle JIT, reading the per-method MDO that the interpreter and C1 already populate.

1. **Function entry count** -- interpreter invocation counter attached as LLVM `function_entry_count` metadata.
2. **Branch / switch weights** -- `BranchData` / `MultiBranchData` lowered to `!prof` `branch_weights` on conditional branches and switches.
3. **Unstable-if pruning** -- strict-zero one-sided branches replaced with a deopt trap (`Reason_unstable_if`, `Action_reinterpret`). Gated on a mature MDO + `too_many_traps` history (C2-style).
4. **Liveness-pruned deopt bundles** -- dead single-word locals at deopt points encoded as `T_ILLEGAL` placeholders.

A read-only `JeandleProfile` facade over `ciMethodData` shields the consumers from MDO layout. Single product flag `JeandleUseProfile` (default off).

Three IR-level concerns and their fix:

- **Same-target `CondBr` + `!prof`** (degenerate `if (cond) {}`) trips a `verifyModule` null-deref on LLVM 22 aarch64. Same-target CondBr emitted without weights and without M2 prune; SimplifyCFG folds it later.
- **Reassociate flattens loop-carried recurrences** like `s += a*7 + (a>>>2)`. A `llvm.fake.use` on the inner add gives it a second SSA use so `isReassociableOp()`'s `hasOneUse` check bails; dropped before instruction selection, no runtime cost.
- **`call @uncommon_trap; unreachable` lets LLVM reverse-propagate** "trap edge unreachable → operand is constant on the kept edge" into catch-path phi inputs and adjacent invokes. The trap is emitted via `llvm.experimental.deoptimize.<ret_type>` instead; LLVM treats the intrinsic as an opaque deopt barrier. `RewriteStatepointsForGC` lowers it to a statepoint call targeting `__llvm_deoptimize`, which `ALL_JEANDLE_DIRECT_ROUTINES` registers as an alias of `SharedRuntime::uncommon_trap_blob()->entry_point()`. Runtime path unchanged.

Measured impact (median, last-iter, aarch64):

| Bench (Renaissance) | NoPGO | PGO   | Δ      |
| ------------------- | ----- | ----- | ------ |
| dotty               | 1748  | 1732  | -0.9%  |
| scrabble            |  328  |  316  | -3.9%  |
| philosophers        | 11419 | 11041 | -3.3%  |
| naive-bayes         |  1727 |  1664 | -3.7%  |
| page-rank           |  7760 |  7669 | -1.2%  |
| fj-kmeans           |  2530 |  2495 | -1.4%  |
| akka-uct            | 22646 | 23407 | +3.4%  |
| finagle-http        |  7311 |  7365 | +0.7%  |
| log-regression      |  1591 |  1580 | -0.7%  |
| mnemonics           |  6758 |  6725 | -0.5%  |

StandaloneNTG.work microbench (median ns/iter): C2 ~500, Jeandle -PGO ~1040, Jeandle +PGO ~880 (-16% vs Jeandle baseline).

Not in this PR (parked for follow-up):

- guarded devirt of monomorphic / bimorphic virtual calls from `ReceiverTypeData`
- type and null speculation on `checkcast` / `instanceof` / explicit null checks
- call-site invocation-count `!prof` from `CounterData` / `VirtualCallData` for the inliner
- array store check elision via `speculative_type` on element klass
- strict-zero arm pruning for `tableswitch` / `lookupswitch`

yadongw:
Resolves #496.

PGO for Jeandle, consuming the per-method MDO from the interpreter and C1. Gated by `JeandleUseProfile` (default off):

- `function_entry_count` from the interpreter invocation counter
- `!prof` `branch_weights` on conditional branches and switches
- strict-zero unstable-if branches replaced with a deopt trap (`Reason_unstable_if`, `Action_reinterpret`), gated on a mature MDO + `too_many_traps`
- dead locals at deopt points encoded as `T_ILLEGAL` (long/double take two slots)

Read-only `JeandleProfile` facade over `ciMethodData`.

> Depends on jeandle/jeandle-llvm#71 (Reassociate fix). CI builds LLVM from that branch via the `CO_REPO`/`CO_BRANCH` lines in the head commit; drop them once #71 lands in `main`.

### Notes

- **Same-target `CondBr`** (degenerate `if (cond) {}`): emitted without weights or the prune; SimplifyCFG folds it. Avoids an LLVM 22 aarch64 verifier null-deref on same-target `!prof` CondBrs.
- **Unstable-if trap**: lowered to `llvm.experimental.deoptimize` (an opaque deopt barrier), which `RewriteStatepointsForGC` turns into a statepoint to `__llvm_deoptimize` — aliased to `uncommon_trap_blob`. Runtime path unchanged. This lowering is unconditional (not gated on `JeandleUseProfile`), so a profile-off build isn't byte-identical, but runs the same blob.
- **Saturated MDO counters**: when a counter hits `max_juint`, weights and the prune are skipped (ratio untrustworthy). Counters saturate upward, never wrap to 0, so a never-taken side still reads strict-zero.
- **Loop-carried recurrences**: the prune drops the cold path and its merge phi, which previously stopped Reassociate from sinking the accumulator into the inner add. Fixed in Reassociate (#71), not worked around here; the old `llvm.fake.use` barrier is removed.

### Performance

Scalar ALU-bound reduction microbench, median ns/iter: C2 ~500, Jeandle off ~1040, Jeandle on ~880 (**-16%**). Renaissance (10 benches): within ±4% of the no-PGO baseline (noise).

### Tests

jtreg `test/hotspot/jtreg/compiler/jeandle/pgo/`: `TestBranchWeights`, `TestUnstableIf`, `TestUnstableIfTypedCatch`, `TestSameTargetCondBr`, `TestReassociateBarrier`, `TestEntryCount`, `TestDeoptLiveness`, `TestUnstableIfPruneBoundaries`. JMH: `test/micro/org/openjdk/bench/jeandle/JeandlePGO.java`.

### Validation

8/8 pgo jtreg on aarch64 and x64; `hotspot tier1` with `+UseJeandleCompiler +JeandleUseProfile` (minus ProblemLists): 2082 passed, 0 failures.

### Follow-up (not in this PR)

Guarded devirt of mono/bimorphic calls; type/null speculation on `checkcast`/`instanceof`; call-site `!prof` for the inliner; array-store-check elision; switch arm pruning.


src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1010,6 +1027,101 @@ void JeandleAbstractInterpreter::interpret() {
...
+
+// Workaround for an LLVM Reassociate heuristic that flattens
+// `(phi + (a + b))` into a 3-operand chain and sinks the loop-header phi onto
+// the recurrence critical path. isReassociableOp() bails when the inner add
+// has more than one use, so we give it a second SSA use via llvm.fake.use --
+// which has no runtime cost (codegen drops it before instruction selection).
+//
+// Innermost loops only: in nested loops the standard rank ordering is what
+// lets LICM group outer-loop operands together for hoisting.
+void JeandleAbstractInterpreter::insert_loop_carried_reassociate_barriers() {
```
taofengliu: As I understand, this makes a loop's recurrence critical path shorter. But does this have any unintended consequences? For example, could the off-critical path become a bottleneck due to resource contention? And fake.use may also interfere with other optimizations.

If this is a general-purpose optimization that yields benefits in the vast majority of scenarios, should we consider improving the Reassociate pass? Moreover, this could be an opportunity to contribute upstream to LLVM.

yadongw: You are right. It's a general bug in the Reassociate pass, merely covered up by loop vectorize, maybe. Let me check.

taofengliu:
Is this PR ready for review? It's still in draft state.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -221,13 +225,20 @@ void JeandleVMState::store(BasicType type, int index, llvm::Value* value) {
...
   for (size_t i = 0; i < _locals.size(); i++) {
-    if (!_locals[i].is_null()) {
+    bool is_double_word = !_locals[i].is_null() &&
+                          is_double_word_type(_locals[i].computational_type());
+    // Locals dead at this bci are encoded as T_ILLEGAL so they don't pin
+    // values live across the deopt point. Restricted to single-word locals
+    // to keep the long/double two-slot layout untouched.
+    bool dead_single_word = !_locals[i].is_null() && !is_double_word &&
+                            liveness.is_valid() && !liveness.at(i);
+    if (!_locals[i].is_null() && !dead_single_word) {
```
taofengliu: We can extend this to long/double. Just push two T_ILLEGAL.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1480,35 +1540,170 @@ void JeandleAbstractInterpreter::increment() {
...
+  }
+  llvm::MDBuilder md_builder(*_context);
+  switch_inst->setMetadata(llvm::LLVMContext::MD_prof, md_builder.createBranchWeights(weights));
+}
+
+// Mirrors C2's path_is_suitable_for_uncommon_trap: gate the strict-zero
+// unstable-if prune on a mature profile, a meaningful sample count, and a
+// trap history that hasn't already de-speculated this bci.
+bool JeandleAbstractInterpreter::path_is_suitable_for_unstable_if_prune(
+    int bci, JeandleProfile::BranchCounts counts) {
```
taofengliu, line 1596:
```suggestion
    int bci, JeandleProfile::BranchCounts counts) {
    if (!UseInterpreter) {
      return false;
    }
```
Like C2.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1480,35 +1540,170 @@ void JeandleAbstractInterpreter::increment() {
...
+  // OSR sees a partially-warmed profile; the unobserved side may just be a
+  // path the resumed frame hasn't reached yet.
+  if (is_osr()) {
```
taofengliu, line 1605: C2's path_is_suitable_for_uncommon_trap has no OSR check. When an OSR-compiled method has a mature MDO with a strict-zero branch side, C2 still prunes it.
Would the C2 approach be better?

taofengliu (commented):
Firstly, for profile datas like branch counters, we need to add overflow detection. They are stored in `uint`. C2 does overflow detection, too.

C2 has two additional uncommon-trap insertion scenarios that this PR does not currently cover. I'd like to discuss whether they're worth adding (now or as follow-up).

  ## Scenario 1: PROB_UNKNOWN → Reason_unreached trap

  In C2's do_if / do_ifnull (parse2.cpp:1475), when branch_prediction() returns PROB_UNKNOWN (MDO not mature, or no profile data at this BCI), C2 stops compiling both successors entirely and emits:

  uncommon_trap(Deoptimization::Reason_unreached,
                Deoptimization::Action_reinterpret, nullptr, "cold");

  The rationale: "I don't have enough information to compile this well; deopt back to the interpreter, accumulate profile, and recompile later with better data."

  In Jeandle, this case currently falls through to emitting a branch without !prof metadata — LLVM applies its own static heuristics for block layout. The generated code is correct but unguided.

  ## Scenario 2: OSR backward-branch with zero count → PROB_UNKNOWN

  In C2's branch_prediction static-heuristic fallback (parse2.cpp:1323), there is an OSR-specific path:

  if (target_bci < bci() && is_osr_parse()) {
      ciProfileData* data = methodData->bci_to_data(bci());
      if (data == nullptr ||
          (data->as_BranchData()->taken() + data->as_BranchData()->not_taken() == 0)) {
          return PROB_UNKNOWN;  // triggers Reason_unreached trap
      }
  }

  When an OSR compilation encounters a backward branch (potential loop back-edge) whose MDO counters are completely zero, C2 interprets this as "the OSR frame has never explored this loop path" and forces a deopt so the interpreter can discover it.

  This PR already returns false from path_is_suitable_for_unstable_if_prune for OSR (preventing erroneous M2 pruning), but does not insert a proactive trap for the zero-count back-edge case.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1480,35 +1540,170 @@ void JeandleAbstractInterpreter::increment() {
...
+  int bci = _bytecodes.cur_bci();
+  JeandleBasicBlock* taken_jbb = bci2block()[_bytecodes.get_dest()];
+  JeandleBasicBlock* fallthrough_jbb = bci2block()[_bytecodes.next_bci()];
+  llvm::BasicBlock* taken_block = taken_jbb->header_llvm_block();
+  llvm::BasicBlock* fallthrough_block = fallthrough_jbb->header_llvm_block();
+
+  // Degenerate `if (cond) {}` where taken == fallthrough. BasicBlockBuilder
+  // pushes the shared target into _successors twice, so the post-loop merge
+  // must see two LLVM edges -- emit a CondBr (not an unconditional br) but
+  // skip M1 weights and M2 prune; !prof on a same-target CondBr trips a
```
taofengliu: What are m0 and m1? They appear elsewhere in the code too, and they don't seem like common JVM terms.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
yadongw: Done. A dead two-word local now emits two T_ILLEGAL at distinct consecutive slots (test: `TestDeoptLiveness.deadLongLocalsAtTrap`, asserts slots [0,1,2,3,4,7]).

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
yadongw: Adopted.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
yadongw: Deliberately more conservative: an OSR compile sees a partially-warmed profile, so a strict-zero side may just be a path the resumed frame hasn't reached yet — pruning it would be a false prune. C2 prunes OSR, but I'd rather not speculate on a half-warmed profile here. Happy to align with C2 if you prefer.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
yadongw: Removed — they were internal milestone labels, not JVM terms. Renamed to describe the feature (branch weights, unstable-if prune).

src/hotspot/share/jeandle/jeandleProfile.cpp
```diff
@@ -0,0 +1,94 @@
...
+    return result;
+  }
+  BranchData* branch = data->as_BranchData();
+  result.taken     = branch->taken();
+  result.not_taken = branch->not_taken();
+  result.valid     = true;
+  // BranchData counters saturate at max_juint (inc_taken clamps, never wrapping
+  // to 0). A saturated side means the taken/not_taken ratio is no longer
+  // trustworthy; flag it so callers can be conservative.
+  result.overflow  = (result.taken == max_juint || result.not_taken == max_juint);
```
taofengliu: I took a closer look at how the profile handles overflow, and I noticed that the `inc_taken` function is never actually called.

The interpreter (`InterpreterMacroAssembler::profile_taken_branch`) uses `addptr` and `sbbptr` to manipulate the branch taken data, treating it as 64-bit and falling back to uint64_max on overflow. 

C1, on the other hand, doesn't handle overflow at all; it simply wraps around and continues accumulating from scratch once the data overflows. 

As for C2's overflow check, it reads the profile data as int and rejects negative values. 

It seems the JVM inherently lacks a unified approach to handling profile data overflow.

For Jeandle, I think we can just follow C2's approach. Although it's not perfectly rigorous, the compilation thresholds ensure that profile overflow almost never occurs in practice.

Or do you have a better alternative?

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1498,35 +1546,186 @@ void JeandleAbstractInterpreter::increment() {
...
+  // pruning; reaching here with a 0 count means immature profile, where 0 is
+  // "rare" rather than "impossible".
+  uint32_t taken_weight     = counts.taken     == 0 ? 1u : (uint32_t) counts.taken;
+  uint32_t not_taken_weight = counts.not_taken == 0 ? 1u : (uint32_t) counts.not_taken;
+  llvm::MDBuilder md_builder(*_context);
+  br->setMetadata(llvm::LLVMContext::MD_prof,
+                  md_builder.createBranchWeights(taken_weight, not_taken_weight));
+}
+
+void JeandleAbstractInterpreter::attach_switch_weights(llvm::SwitchInst* switch_inst, int bci) {
```
taofengliu, line 1556: Do we need to do the mature check and reject the immature profile?

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1498,35 +1546,186 @@ void JeandleAbstractInterpreter::increment() {
   _jvm->istore(_bytecodes.get_index(), result);
 }
 
+void JeandleAbstractInterpreter::attach_branch_weights(llvm::BranchInst* br, int bci) {
```
taofengliu, line 1537: Do we need to do the mature check and reject the immature profile?

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1498,35 +1546,186 @@ void JeandleAbstractInterpreter::increment() {
...
+          cond_true_is_hot ? _ir_builder.CreateCondBr(cond, hot_block, trap_block)
+                           : _ir_builder.CreateCondBr(cond, trap_block, hot_block);
+      llvm::MDBuilder md_builder(*_context);
+      uint32_t hot_weight  = 0x7FFFFFFFu;
+      uint32_t cold_weight = 1u;
+      br->setMetadata(llvm::LLVMContext::MD_prof,
+                      md_builder.createBranchWeights(
+                          cond_true_is_hot ? hot_weight : cold_weight,
+                          cond_true_is_hot ? cold_weight : hot_weight));
+      uncommon_trap(Deoptimization::Reason_unstable_if,
```
taofengliu, line 1677: Here maybe we can use the liveness of the pruned branch's bci, then the liveness info will be more accurate. Just update the `_bytecodes`'s bci to the pruned bci and recover it after `uncommon_trap`.

test/hotspot/jtreg/compiler/jeandle/pgo/TestReassociateBarrier.java
taofengliu, line 1: If we delete this test, we can merge this PR separately from the LLVM's reassociate PR  (or add it through another PR).

src/hotspot/share/jeandle/jeandleRuntimeRoutine.hpp
```diff
@@ -150,120 +143,143 @@
...
+      llvm::Type::getDoubleTy(context))                              \
+                                                                     \
+  def(uncommon_trap,                                                 \
+      SharedRuntime::uncommon_trap_blob()->entry_point(),            \
+      true,                                                          \
+      false,                                                         \
+      llvm::Type::getVoidTy(context),                                \
+      llvm::Type::getInt32Ty(context))                               \
+                                                                     \
+  def(__llvm_deoptimize,                                             \
```
rcjhd: this is a duplicate definition of `uncommon_trap` above, we can delete the old one to keep consistency in future use.

src/hotspot/share/jeandle/jeandleProfile.cpp
yadongw: `==max_juint` 确实漏了 `[INT_MAX+1, max_juint)`。改成 `(jint)<0` 判负,branch/switch 一致。不查 sum:每边是独立 u32 权重,无需限制两者之和。`inc_taken` 注释删了。

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
yadongw: 已加 `!is_mature()` 拒绝,和 prune 的门一致。

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
yadongw: 已加 `!is_mature()` 拒绝,和 prune 的门一致。

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
yadongw: if 操作数有意留在栈上,在 if 处 reinterpret(≈ C2 `repush_if_args`)。只改记 pruned bci、栈不动会让帧不一致;要用 pruned liveness 得整体换成 cold-target reinterpret,是另一套 deopt 策略。这个 PR 先保持 at-if,cold-target 版可后续做。

test/hotspot/jtreg/compiler/jeandle/pgo/TestReassociateBarrier.java
yadongw: 已删,这样能独立于 reassociate PR 合入,后续再加回。

src/hotspot/share/jeandle/jeandleRuntimeRoutine.hpp
yadongw: 已删,保留 `__llvm_deoptimize`。deopt 现在统一走 `llvm.experimental.deoptimize` → `__llvm_deoptimize`,没人再用 `uncommon_trap` 这条,两者指向同一 blob。aarch64/x64 重建过,PGO 测试都过。

taofengliu:
@yadongw Please fix the CI and resolve conflicts

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -930,6 +949,10 @@ void JeandleAbstractInterpreter::interpret() {
     assert(current->is_set(JeandleBasicBlock::is_loop_header), "sanity");
   }
 
+  if (uint entry_count = _profile.entry_count()) {
+    _llvm_func->setEntryCount((uint64_t) entry_count);
+  }
```
taofengliu: We can remove entry count profile since it makes JITLink crash wile creating link graph, and it's not used by any optimization in Jeandle.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -930,6 +949,10 @@ void JeandleAbstractInterpreter::interpret() {
...
+
```
```suggestion
```

taofengliu (dismissed):

taofengliu (approved):

Issue #468, opened 2026-05-13
CHA analysis not implemented.


stevenJOJO129:
**jtreg tier2 CHA test failed**
All three testcase of CHA failed.
- compiler/cha/AbstractRootMethod.java
- compiler/cha/DefaultRootMethod.java
- compiler/cha/StrengthReduceInterfaceCall.java

By analysing, it seemed that **CHA is not implemented**, so there will be no deoptimization when class hierarchy relationship changed, and that cause the testcase failed.

**The feature you'd like**
CHA(Class Hierarchy Analysis) is an important feature in c2 compiler. It allow virtual method call and interface call back to static call. So that enables further inlining optimization, which may improve performance significantly in some situation.

**Implementation you've considered**
In jeandleAbstractInterpreter, add CHA check when translate invokevirtual and invokeinterface bytecode, if satisfy CHA devirtualization condition, change the virtual invoke to static ones. Guard code is also needed, so when the class hierarchy chain changed, we can find it and deoptimize. 

**Additional context**

PR #500, opened 2026-05-27
feat: Support invokevirtual and invokeinterface CHA optimization.

stevenJOJO129:
## Related issue(s):

issue #468

This PR implements CHA optimization and deopt for normal invoke (invokevirtual and invokeinterface bytecode). invokehandle support need constant propagation, will be support later.

## Test
| testcase | passed |
| -- | -- |
| compiler/cha/StrengthReduceInterfaceCall.java | ✅ | 
| compiler/cha/AbstractRootMethod.java | passed normal invoke part, **failed at MethodHandle.invokeExact** |
| compiler/cha/DefaultRootMethod.java | passed normal invoke part, **failed at MethodHandle.invokeExact** |

### DIFF
--- a/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
+++ b/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
@@ -26,13 +26,14 @@
 #include "llvm/IR/Intrinsics.h"
 #include "llvm/IR/MDBuilder.h"
 #include "llvm/IR/Jeandle/Attributes.h"
+#include "llvm/IR/Jeandle/Deoptimization.h"
 #include "llvm/IR/Jeandle/GCStrategy.h"
 #include "llvm/IR/Jeandle/JavaType.h"
 #include "llvm/IR/Jeandle/Metadata.h"
 
-
 #include "jeandle/jeandleAbstractInterpreter.hpp"
 #include "jeandle/jeandleCompiledCall.hpp"
+#include "jeandle/jeandleCompiledCode.hpp"
 #include "jeandle/jeandleIntrinsicLowering.hpp"
 #include "jeandle/jeandleRuntimeRoutine.hpp"
 #include "jeandle/jeandleType.hpp"
@@ -56,6 +57,9 @@
 #include "runtime/sharedRuntime.hpp"
 #include "runtime/stubRoutines.hpp"
 #include "utilities/ostream.hpp"
+
+using llvm::jeandle::DeoptValueEncoding;
+using llvm::jeandle::HotspotBasicType;
 
 JeandleVMState::JeandleVMState(int max_stack, int max_locals, llvm::LLVMContext *context) :
                                _stack(), _locals(max_locals), _locks(), _context(context) {
@@ -256,10 +260,10 @@ llvm::SmallVector<llvm::Value*> JeandleVMState::deopt_args(llvm::IRBuilder<>& bu
   /* TODO: scalar */
 
   if (parse_context.is_inlinee()) {
-    uint64_t encode = DeoptValueEncoding(0, DeoptValueEncoding::MethodType, T_METADATA).encode();
+    uint64_t encode = DeoptValueEncoding(0, DeoptValueEncoding::MethodType, llvm::jeandle::T_METADATA).encode();
 #ifdef ASSERT
     if (log_is_enabled(Trace, jeandle)) {
-      DeoptValueEncoding::decode(encode).print();
+      print_deopt_value(DeoptValueEncoding::decode(encode));
     }
 #endif
     args.push_back(builder.getInt64(encode));
@@ -287,10 +291,11 @@ llvm::SmallVector<llvm::Value*> JeandleVMState::deopt_args(llvm::IRBuilder<>& bu
     // slots (one per word) so the two-slot layout of later locals stays aligned.
     bool dead = !_locals[i].is_null() && liveness.is_valid() && !liveness.at(i);
     if (!_locals[i].is_null() && !dead) {
-      uint64_t encode = DeoptValueEncoding(i, DeoptValueEncoding::LocalType, _locals[i].computational_type()).encode();
+      uint64_t encode = DeoptValueEncoding(i, DeoptValueEncoding::LocalType, 
+          static_cast<HotspotBasicType>(_locals[i].computational_type())).encode();
 #ifdef ASSERT
       if (log_is_enabled(Trace, jeandle)) {
-        DeoptValueEncoding::decode(encode).print();
+        print_deopt_value(DeoptValueEncoding::decode(encode));
       }
 #endif
       args.push_back(builder.getInt64(encode));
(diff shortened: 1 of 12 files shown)


taofengliu (commented):
Maybe it's better to optimize virtual calls in LLVM side. In LLVM side, we can query the type of receiver with full context, it helps us to do the virtual call optimization. How do you think?

stevenJOJO129:
> Maybe it's better to optimize virtual calls in LLVM side. In LLVM side, we can query the type of receiver with full context, it helps us to do the virtual call optimization. How do you think?

Change it to LLVM side. We implement a callback in jdk side to query and change the structure. The receiver's java type is get in llvm side.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1879,8 +1883,24 @@ void JeandleAbstractInterpreter::invoke() {
...
+                                                 std::to_string(reinterpret_cast<uintptr_t>(ciEnv::get_instance_klass_for_declared_method_holder(holder))));
+  llvm::Attribute call_site_attr = llvm::Attribute::get(*_context,
+                                                 llvm::jeandle::Attribute::CallSite,
+                                                 std::to_string(reinterpret_cast<uintptr_t>(call_site)));
   invoke->addFnAttr(id_attr);
   invoke->addFnAttr(patch_bytes_attr);
+  invoke->addFnAttr(bc_attr);
+  invoke->addFnAttr(resolved_method_attr);
+  invoke->addFnAttr(declared_holder_attr);
+  invoke->addFnAttr(call_site_attr);
```
taofengliu: In fact we can query CallSiteInfo according to statepoint_id, then we can get everything else from CallSiteInfo.
So maybe we don't need to store so many information to the call inst.

src/hotspot/share/jeandle/jeandleVMCallback.cpp
```diff
@@ -87,6 +107,323 @@ bool jeandle_is_effectively_final(uintptr_t klass_ptr) {
...
+    ciMethod* exact_method = callee->resolve_invoke(caller->holder(), actual_receiver);
+    if (exact_method != nullptr) {
+      return exact_method->holder();
+    }
+  }
+
+  return nullptr;
+}
+
+bool jeandle_optimize_callbase(uintptr_t cb_ptr, uintptr_t dtu_ptr, uintptr_t caller_ptr,
```
taofengliu: I suggest we don't update IR directly in VM callbacks. VM callbacks are used to supply runtime information. All IR update should be done in LLVM passes.
If we update IR in VM callbacks, they can not be replayed through VM callback log.

src/hotspot/share/jeandle/jeandleVMCallback.cpp
```diff
@@ -87,6 +107,323 @@ bool jeandle_is_effectively_final(uintptr_t klass_ptr) {
...
+
+bool getUInt64FnAttr(const llvm::CallBase *CB, llvm::StringRef Name,
+                            uint64_t &Out) {
+  llvm::Attribute A = CB->getFnAttr(Name);
+  if (!A.isValid() || !A.isStringAttribute())
+    return false;
+  return !A.getValueAsString().getAsInteger(10, Out);
+}
+
+llvm::StringRef getStringFnAttr(const llvm::CallBase *CB, llvm::StringRef Name) {
```
```suggestion
llvm::StringRef getStringFnAttr(const llvm::CallBase* CB, llvm::StringRef Name) {
```
This is the Hotspot coding style

taofengliu (commented):
Please organize the code into clear sections and add more comments.

Cruise20:
Please set `CO_REPO` and `CO_BRANCH` in your commit comment, as described here: https://github.com/jeandle/jeandle-jdk/blob/main/jeandle-docs/development-guide.md#co-dependent-changes

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
stevenJOJO129: Delete CallSite Attribute. Use id to get callsite ptr.

src/hotspot/share/jeandle/jeandleVMCallback.cpp
stevenJOJO129: Now the callback did't update llvm IR. However, we need to update callsite, it's saved in jdk side, so there is no side-effect on llvm IR.

src/hotspot/share/jeandle/jeandleVMCallback.cpp
stevenJOJO129: Changed.

src/hotspot/share/jeandle/jeandleCompilation.cpp
```diff
@@ -1070,6 +1074,13 @@ void JeandleCompilation::setup_llvm_module(llvm::MemoryBuffer* template_buffer)
 
   llvm::NamedMDNode* metadata_node = _llvm_module->getOrInsertNamedMetadata(llvm::jeandle::Metadata::JavaMethodCompilation);
   assert(metadata_node != nullptr, "invalid metadata node");
+  llvm::NamedMDNode* patch_node = _llvm_module->getOrInsertNamedMetadata(llvm::jeandle::Metadata::StaticCallPatchSize);
+  assert(patch_node != nullptr, "invalid patch node");
+  llvm::Metadata* patch_size_md =
+    llvm::ConstantAsMetadata::get(
+      llvm::ConstantInt::get(llvm::Type::getInt32Ty(*_context),
+                                JeandleCompiledCall::call_site_patch_size(JeandleCompiledCall::STATIC_CALL)));
+  patch_node->addOperand(llvm::MDNode::get(*_context, patch_size_md));
```
taofengliu: Move to `RuntimeDefinedJavaOps::define_metadata`

src/hotspot/share/jeandle/jeandleVMCallback.cpp
```diff
@@ -308,6 +332,142 @@ bool jeandle_record_inlining_complete() {
...
+
+llvm::jeandle::CHAOptInfo optimize_virtual_call(ciMethod* caller,
+                            ciMethod* callee, ciInstanceKlass* holder,
+                            Klass* receiver_klass, bool is_exact) {
+  ciEnv* env = ciEnv::current();
+
+  // Same fast paths as C2's Compile::optimize_inlining.
+  if (callee->can_be_statically_bound()) {
+    return {};
+  }
```
taofengliu: `JeandleAbstractInterpreter` has handled this, we can comment it out.

taofengliu (commented):
I think we should add some jtreg tests for CHA devirtualization

taofengliu (commented):
https://github.com/jeandle/jeandle-jdk/commit/bdc33b2b3b16f0da61b58ef8fb7c9ce07ab8404c

This commit changed the format of deopt bundle, we need to adapt.

stevenJOJO129:
> I think we should add some jtreg tests for CHA devirtualization

There are already CHA test cases in jtreg, under `compiler/cha`. After this PR is merged, `StrengthReduceInterfaceCall.java` will pass.

taofengliu (approved):

PR #505, opened 2026-05-28
refactor: type usage in frontend

taofengliu:
-  We don't need to check class loaded for null check. We can delay this to the usage of the oop. C2 does this check because C2 may has sharper type from compile-time optimizations, the sharper type maybe unloaded.
-  Add a TODO: we can attach array element type in LLVM side, then we can use context-sensitive type information of array.


minchao1024 (approved):
Looks good

taofengliu:
@mergifyio queue

PR #510, opened 2026-06-12
feat: constant folding

### LINKED ISSUE #509: Const oops' types are not handle by JavaType interface.
**Goals**
Make JavaType interface aware of const oops' types.

**Description**
JavaType interface is used to query a value's type for LLVM passes, but it only read types from attributes and metadatas. Const oops are declared as global variables, and load instructions for them are not attached with metadata

taofengliu:
Implement constant folding for static final fields.
Also fix #509 

TODO:
- Support stable fields
- Support const array elements folding


src/hotspot/share/jeandle/jeandleVMCallback.cpp
```diff
@@ -87,6 +94,127 @@ bool jeandle_is_effectively_final(uintptr_t klass_ptr) {
...
+  ciConstant con;
+  if (!jeandle_constant_field(oop_id, offset, &field, &con))
+    return -1;
+  return field->layout_type();
+}
+
+const char* jeandle_get_oop_handle_name(int oop_id) {
+  JeandleCompilation* compilation = JeandleCompilation::current();
+  assert(compilation != nullptr, "no active compilation");
+  return compilation->compiled_code()->oop_handle_name_cstr(oop_id);
```
Cruise20: Will the pointer returned here become invalid due to the expansion of smallvector? Maybe a copy of `name` string can be returned here.

src/hotspot/share/jeandle/jeandleVMCallback.cpp
```diff
@@ -87,6 +94,127 @@ bool jeandle_is_effectively_final(uintptr_t klass_ptr) {
...
+  *field = found;
+  *con = value;
+  return true;
+}
+
+int64_t jeandle_get_constant_field_value(int oop_id, int offset) {
+  ciField* field = nullptr;
+  ciConstant con;
+  if (!jeandle_constant_field(oop_id, offset, &field, &con)) {
+    return 0;
```
Cruise20: How to distinguish between constant 0 and non-constant?

Cruise20 (approved):

taofengliu:
@mergifyio queue

taofengliu:
@mergifyio queue

Issue #514, opened 2026-06-16
java/lang/instrument/NativeMethodPrefixAgent.java crash


DXwangg:
**Describe the bug**
java/lang/instrument/NativeMethodPrefixAgent.java crash

**Minimal reproduce steps**
Run java/lang/instrument/NativeMethodPrefixAgent.java using Jeandle

DXwangg:
Jeandle exposed this dynamic Java call to LLVM as a direct call to the resolved Java method symbol. LLVM could then optimize it as if the callee were statically known, and incorrectly fold ClassVisitor::visitMethod into returning null.

This breaks the ASM visitor chain used during class transformation in NativeMethodPrefixAgent, eventually leading to JVM crash in later class loading / initialization paths.

DXwangg:
`test/jdk/java/lang/instrument/NativeMethodPrefixAgent.java` fails with a JVM crash when running with Jeandle.

The failure was narrowed down to JDK internal ASM. Excluding ASM from Jeandle compilation makes the test pass:
```
 -XX:CompileCommand=exclude,jdk.internal.org.objectweb.asm.*::*
```

Further reduction shows the minimal problematic method is: 
``` jdk.internal.org.objectweb.asm.ClassVisitor::visitMethod ```

This method contains a non-statically-bound invokevirtual call: 
``` return cv.visitMethod(access, name, descriptor, signature, exceptions); ```

Bytecode inspection confirms that this is an invokevirtual, so the actual target must be resolved by Java dynamic dispatch at runtime.

However, Jeandle exposed this dynamic call to LLVM as a direct call to the resolved Java method symbol. LLVM therefore misinterpreted ClassVisitor::visitMethod as a direct self-recursive call, while it should actually be modeled as a Java dynamic dispatch.

Before LLVM optimization, the IR still preserved the expected behavior:
```
%ret = invoke hotspotcc ptr addrspace(1) @"...ClassVisitor_visitMethod..."(...)
        to label %normal_dest unwind label %unwind_dest

normal_dest:
  ret ptr addrspace(1) %ret
```
After native LLVM optimization passes, especially SCCPPass / IPSCCPPass, the method was incorrectly folded to:
```
ret ptr addrspace(1) null
```
This breaks the ASM visitor chain used by NativeMethodPrefixAgent, eventually causing JVM crashes during later class transformation / class loading paths.

PR #515, opened 2026-06-16
fix: hide dynamic invoke target from LLVM

DXwangg:
## Related issue(s):

issue #514 

This PR fixes Jeandle IR generation for invokevirtual/invokeinterface calls that aren't statically bound.

Before, Jeandle exposed these dynamic Java calls to LLVM as direct calls to the resolved Java method, which could make LLVM treat them like normal direct calls and wrongly fold return values. This caused test/jdk/java/lang/instrument/NativeMethodPrefixAgent.java and test/hotspot/jtreg/compiler/whitebox/DeoptimizeMultipleOSRTest.java to crash in Jeandle.

Now, this PR uses a __jeandle_dynamic_call. placeholder as the LLVM callee for dynamic virtual/interface calls, while still recording the HotSpot call site as DYNAMIC_CALL. This keeps Java's dynamic dispatch behavior and avoids bad LLVM optimizations.


taofengliu (approved):

taofengliu:
@mergifyio queue

PR #516, opened 2026-06-16
feat: support receiver-profile devirtualization

lfeng14:
## Related issue(s):

issue #

This PR adds receiver-profile based devirtualization support in Jeandle.

It covers monomorphic, bimorphic, and major-receiver virtual/interface call sites. For monomorphic and bimorphic profiles, Jeandle emits receiver class guards and converts matched paths to direct static calls, with uncommon traps on unexpected receivers. For major-receiver profiles, Jeandle emits a guarded direct call for the dominant receiver and keeps a normal dynamic-call fallback for minority receivers.

Note: this PR includes changes based on another contributor's commit.


src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1712,49 +1988,44 @@ void JeandleAbstractInterpreter::invoke() {
...
-    ciKlass* receiver_constraint = nullptr;
-    if (bc == Bytecodes::_invokespecial && !target->is_object_initializer()) {
-      ciInstanceKlass* sender_klass = _method->holder();
-      if (sender_klass->is_interface()) {
-        receiver_constraint = sender_klass;
-      }
-    } else if (bc == Bytecodes::_invokeinterface && target->is_private()) {
-      assert(holder->is_interface(), "How did we get a non-interface method here!");
-      receiver_constraint = holder;
+  ciKlass* receiver_constraint = nullptr;
```
rcjhd: This part of the code doesn't seem to have any substantial changes compared to the previous version. Can we revert it back?

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1786,34 +2057,373 @@ void JeandleAbstractInterpreter::invoke() {
...
-  // Construct arguments.
-  const int arg_size = method_signature->count() + receiver;
-  llvm::SmallVector<llvm::Value*> args(arg_size);
-  llvm::SmallVector<llvm::Type*> args_type(arg_size);
-  for (int i = method_signature->count() - 1; i >= 0; --i) {
-    BasicType type = method_signature->type_at(i)->basic_type();
-    args[i + receiver] = _jvm->pop(type);
-    args_type[i + receiver] = JeandleType::java2llvm(type, *_context);
+  // Devirtualize based on receiver profile (monomorphic or bimorphic).
+  bool monomorphic_devirt = false;
```
rcjhd: I'd suggest factoring out the judgment logic for monomorphic, bimorphic, and major types into dedicated helper functions, rather than keeping it all inside the `invoke` method.

src/hotspot/share/jeandle/jeandleRuntimeRoutine.hpp
```diff
@@ -150,120 +150,143 @@
...
-      llvm::Type::getDoubleTy(context),                                             \
-      llvm::Type::getDoubleTy(context))                                             \
-                                                                                    \
-  def(install_exceptional_return_for_call_vm,                                       \
-      JeandleRuntimeRoutine::install_exceptional_return_for_call_vm,                \
-      false,                                                                        \
-      true,                                                                         \
-      llvm::Type::getVoidTy(context))                                               \
-                                                                                    \
+#define ALL_JEANDLE_DIRECT_ROUTINES(def)                             \
```
rcjhd, line 153: I don't see why we need to change this just for backslash alignment.

src/hotspot/share/jeandle/jeandleProfile.cpp
```diff
@@ -0,0 +1,257 @@
...
+}
+
+JeandleProfile::ReceiverProfile JeandleProfile::monomorphic_receiver_at(int bci) const {
+  ReceiverProfile result = {nullptr, 0, 0, false};
+  if (_method == nullptr || !is_mature()) {
+    return result;
+  }
+
+  ciCallProfile profile = _method->call_profile_at_bci(bci);
+  if (profile.morphism() != 1 || !profile.has_receiver(0) || profile.receiver_count(0) <= 0) {
```
rcjhd: The three `receiver_at` methods contain similar if logic. We could extract that into a helper function with a clear name that indicates its intent.

src/hotspot/share/jeandle/jeandleProfile.cpp
```diff
@@ -0,0 +1,257 @@
...
+  ciKlass* recv0 = profile.receiver(0);
+  ciKlass* recv1 = profile.receiver(1);
+  if (recv0 == nullptr || !recv0->is_loaded() ||
+      recv1 == nullptr || !recv1->is_loaded()) {
+    return result;
+  }
+
+  uint count0 = (uint) profile.receiver_count(0);
+  uint count1 = (uint) profile.receiver_count(1);
+  if (count0 <= 0 || count1 <= 0) {
```
rcjhd: This is strange for `uint <= 0` , you should refactor like this:
```
int count0 = profile.receiver_count(0);
int count1 = profile.receiver_count(1);
if (count0 <= 0 || count1 <= 0) return result;

result.count0 = (uint) count0;
result.count1 = (uint) count1;
```

src/hotspot/share/jeandle/jeandleProfile.cpp
```diff
@@ -0,0 +1,257 @@
...
+                                             Deoptimization::DeoptReason reason,
+                                             int taken,
+                                             int not_taken) const {
+  return should_use_branch_profile(taken, not_taken) &&
+         !has_trap_at(bci, reason) &&
+         !has_too_many_traps(reason) &&
+         !has_too_many_recompiles(bci, reason);
+}
+bool JeandleProfile::should_speculate_receiver(int bci, Deoptimization::DeoptReason reason) const {
+  return is_mature() &&
```
rcjhd, line 122: Are these conditions too weak? Please cross-check the admission criteria against C2. 
Also, structurally, you might consider picking some precondition checks from functions like `monomorphic_receiver_at` below into this layer as well, to separate the 'if can do' logic from the 'get profile' logic — similar to how the branch function handles it.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1786,34 +2057,373 @@ void JeandleAbstractInterpreter::invoke() {
...
+      phi->addIncoming(invoke_miss, normal_dest_miss);
+      _jvm->push(return_type, phi);
+    }
+
+    return;
+  }
+
+  // Interface CHA devirtualization: guard the receiver against the unique
+  // implementor constraint, then call the unique concrete target directly.
+  if (interface_cha_devirt) {
```
rcjhd: You don't need to do CHA — someone else has already taken care of it. Remove all codes about that.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1786,34 +2057,373 @@ void JeandleAbstractInterpreter::invoke() {
...
+  uint major_receiver_site_count = 0;
+
+  bool cha_devirt = false;
+  ciMethod* cha_resolved = nullptr;
+
+  bool interface_cha_devirt = false;
+  ciMethod* interface_cha_resolved = nullptr;
+  ciInstanceKlass* interface_cha_constraint = nullptr;
+
+  if ((bc == Bytecodes::_invokevirtual || bc == Bytecodes::_invokeinterface) &&
```
rcjhd: I think you could move this logic into the switch cases below. The current approach with early if returns feels a bit messy.

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1786,34 +2057,373 @@ void JeandleAbstractInterpreter::invoke() {
...
+      !target->can_be_statically_bound() &&
+      receiver_value != nullptr) {
+    int bci = _bytecodes.cur_bci();
+
+    // Try monomorphic first.
```
rcjhd: I'd suggest we look into the order used in C2. We could open an issue to record it and revisit together later. Ideally, we should keep the order consistent with how C2 does it — the current order looks a bit naive to me.

PR #523, opened 2026-06-22
feat: add numberOf{Leading,Trailing}Zeros intrinsics for int and long

eapenzhan:

Map Integer/Long.numberOf{Leading,Trailing}Zeros to llvm.ctlz/cttz, the same LLVM-builtin family as bitCount. Fix the trailing is_zero_poison flag to false so numberOf*Zeros(0) returns the bit width (32/64) as the Java spec requires; the long variants truncate the i64 result to the int that Java returns.

Unlike popcount, ctlz/cttz need no CPU gating: LLVM always lowers them to native sequences on x86-64 (bsr/bsf fallback) and aarch64 (CLZ, RBIT+CLZ), never a libcall, matching C2 which always intrinsifies these.


taofengliu (approved):

taofengliu:
@mergifyio queue

PR #525, opened 2026-06-25
fix: implement Math.addExact intrinsic for int and long (closes #524)

### LINKED ISSUE #524: Implement intrinsic _addExactI and _addExactL
**The feature you'd like**
Implement intrinsic _addExactI and _addExactL

xugeeeee:
Add Jeandle intrinsic support for Math.addExact(int,int) and Math.addExact(long,long) (vmIntrinsicID::_addExactI / _addExactL).

Both variants use llvm::Intrinsic::sadd_with_overflow to compute the sum and detect signed overflow in a single instruction.  On overflow the JVM-state is re-executed via uncommon_trap so the interpreter can throw ArithmeticException; on the non-overflow path the result is pushed normally.

Stack peek (raw_peek) rather than pop is used before the branch so the safepoint that guards uncommon_trap sees the original arguments on the operand stack, allowing correct deoptimization re-execution.

A jtreg test TestAddExact.java verifies:
 - correct results for in-range additions (int and long)
 - ArithmeticException on positive and negative overflow
 - LLVM IR contains sadd.with.overflow and the expected block labels

## Related issue(s):

issue 524#

Jeandle currently falls through to the Java implementation for Math.addExact, missing the intrinsic optimization. This PR adds native intrinsic support so Jeandle compiles addExact calls directly to a sadd.with.overflow instruction instead of an invokestatic.

### DIFF
--- a/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
+++ b/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
@@ -274,6 +274,7 @@ llvm::SmallVector<llvm::Value*> JeandleVMState::deopt_args(llvm::IRBuilder<>& bu
   // Duplicate the BCI as a BCI marker for the LLVM backend.
   // Keep TestScopeValues.java in sync with this duplicated-BCI convention.
   args.push_back(builder.getInt32(bci));
+  args.push_back(builder.getInt32(bci));
   for (size_t i = 0; i < _locals.size(); i++) {
     bool is_double_word = !_locals[i].is_null() &&
                           is_double_word_type(_locals[i].computational_type());
--- a/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
+++ b/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
@@ -233,7 +233,8 @@ void JeandleVMState::store(BasicType type, int index, llvm::Value* value) {
 llvm::SmallVector<llvm::Value*> JeandleVMState::deopt_args(llvm::IRBuilder<>& builder,
                                                            MethodLivenessResult liveness,
                                                            const JeandleParseContext& parse_context,
-                                                           int bci) {
+                                                           int bci,
+                                                           bool should_reexecute) {
 #ifdef ASSERT
   if (log_is_enabled(Trace, jeandle)) {
     tty->print_cr("Build deopt bundle at bci %d :", bci);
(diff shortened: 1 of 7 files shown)


src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
```diff
@@ -669,3 +681,53 @@ bool JeandleIntrinsicLowering::lower_new_array() {
...
+  bool is_long = (id == vmIntrinsics::_addExactL);
+
+  llvm::Type* ty = JeandleType::java2llvm(
+      is_long ? BasicType::T_LONG : BasicType::T_INT, ctx);
+
+  // Stack layout for int (1-slot each): depth 0 = arg2 (top), depth 1 = arg1.
+  // Stack layout for long (double-word): depth 0 = null_hi2, depth 1 = arg2,
+  //                                      depth 2 = null_hi1, depth 3 = arg1.
+  llvm::Value* arg2 = _interp->_jvm->raw_peek(is_long ? 1 : 0).value();
+  llvm::Value* arg1 = _interp->_jvm->raw_peek(is_long ? 3 : 1).value();
```
taofengliu: Use `peek_value`

src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
```diff
@@ -669,3 +681,53 @@ bool JeandleIntrinsicLowering::lower_new_array() {
...
+      llvm::Intrinsic::sadd_with_overflow, {ty}, {arg1, arg2});
+  llvm::Value* result   = builder.CreateExtractValue(res, 0);
+  llvm::Value* overflow = builder.CreateExtractValue(res, 1);
+
+  const std::string pfx = "bci_" + std::to_string(cur_bci) +
+                           (is_long ? "_addExactL" : "_addExactI");
+  llvm::BasicBlock* ok_bb = llvm::BasicBlock::Create(ctx, pfx + "_ok",       _interp->_llvm_func);
+  llvm::BasicBlock* ov_bb = llvm::BasicBlock::Create(ctx, pfx + "_overflow", _interp->_llvm_func);
+
+  builder.CreateCondBr(overflow, ov_bb, ok_bb);
```
taofengliu: C2 explicitly marks the overflow branch as very unlikely. We should do the same using branch weight hint metadatas.

taofengliu (commented):
In jeandle-jdk/src/hotspot/share/opto/library_call.cpp:1915, C2 sets the bci as "should_reexecute", maybe we should do the same. This state should be propagated to `JeandleCallReloc::process_stack_map`

taofengliu (dismissed):

taofengliu:
@mergifyio queue

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -265,6 +266,11 @@ llvm::SmallVector<llvm::Value*> JeandleVMState::deopt_args(llvm::IRBuilder<>& bu
     args.push_back(builder.getInt64(uint64_t(parse_context.method())));
   }
 
+  // should_reexecute is explicitly set by intrinsic lowering (e.g. addExact's
+  // overflow trap) to force reexecution of the current bci on deopt, matching
+  // C2's should_reexecute semantics for intrinsic-emitted uncommon traps.
+  args.push_back(builder.getInt32(should_reexecute ? 1 : 0));
```
```suggestion
  args.push_back(builder.getInt64(should_reexecute ? 1 : 0));
```

taofengliu (approved):

taofengliu (approved):

PR #527, opened 2026-06-26
feat: polish PGO code

taofengliu:


src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1472,35 +1534,187 @@ void JeandleAbstractInterpreter::increment() {
   _jvm->istore(_bytecodes.get_index(), result);
 }
 
+void JeandleAbstractInterpreter::attach_branch_weights(llvm::BranchInst* br, int bci) {
+  JeandleProfile::BranchCounts counts = _profile.branch_at(bci);
+
+  if (!_profile.is_mature() || !counts.valid || counts.overflow) {
```
rcjhd: Do we need a special case to exclude the scenario where all counts are zero? Otherwise, the following logic would default to setting it to **weight (1,1)**, creating the illusion that all branches have equal hotness. I think this might differ from having no profile at all, and could potentially render static optimization analysis ineffective.
Also the later attach_switch_weights().

src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
```diff
@@ -1410,12 +1436,48 @@ void JeandleAbstractInterpreter::uncommon_trap(Deoptimization::DeoptReason reaso
...
-  // mark unreachable
-  _ir_builder.CreateUnreachable();
+  // Pre-declare __llvm_deoptimize with the exact signature
+  // RewriteStatepointsForGC will request when it lowers the intrinsic
+  // (`void(i32)` for the trap request operand), pinned to Hotspot_JIT so
+  // the lowered statepoint's call agrees with the uncommon_trap blob's
+  // entry. RS4GC otherwise synthesises a default-CC declaration via
+  // getOrInsertFunction, leaving the lowered call as CallingConv::C with
+  // a latent CC mismatch against the runtime entry.
+  if (_module.getFunction("__llvm_deoptimize") == nullptr) {
```
wazlyqa: I think `__llvm_deoptimize` should be declared in template.ll instead of being synthesized from JeandleAbstractInterpreter.

This symbol is an IR-level runtime target produced by RewriteStatepointsForGC when it lowers `llvm.experimental.deoptimize`, so keeping the declaration in the template module is more direct and ensures every Java method compilation module starts with the correct declaration:

```
  declare hotspotcc void @__llvm_deoptimize(i32)
```

taofengliu: Fixed

PR #535, opened 2026-07-07
enhance: Track Jeandle inline attempts with reasons

minchao1024:
## Related issue(s):

issue #428

Track Jeandle inline attempts with explicit success and failure reasons.

- Preallocate inline tree nodes before LLVM mutates IR, and commit them only after inline success.
- Handle LLVM RecordInlineResult so failed inline attempts are recorded for diagnostics.
- Print inline trees with reason annotations and mark failed call sites as `X@bci`.
- Update TestInline to distinguish successful inline nodes from failed attempts.

Example inline tree:
```
Jeandle inline tree:
java/lang/invoke/InvokerBytecodeGenerator::addMethod
|- @1 java/lang/invoke/InvokerBytecodeGenerator::methodPrologue  [inline (hot)]
|  `- @4 java/lang/invoke/MethodType::toMethodDescriptorString  [inline (hot)]
|     `- X@17 sun/invoke/util/BytecodeDescriptor::unparseMethod  [failed: too big]
|- @194 java/lang/invoke/InvokerBytecodeGenerator::emitStoreResult  [inline (hot)]
|  `- @23 java/lang/invoke/InvokerBytecodeGenerator::emitStoreInsn  [inline (hot)]
|     `- X@2 java/lang/invoke/InvokerBytecodeGenerator::storeInsnOpcode  [failed: too big]
`- @721 java/lang/invoke/InvokerBytecodeGenerator::methodEpilogue  [inline (hot)]
```

### DIFF
--- a/src/hotspot/share/jeandle/jeandleCompilation.cpp
+++ b/src/hotspot/share/jeandle/jeandleCompilation.cpp
@@ -682,7 +682,7 @@ bool JeandleInlineTree::should_inline(JeandleCompilation* comp,
     return false;
   }
   reason = JeandleInlineReason::InlineHot;
-  return size <= max_inline_size;
+  return true;
 }
 
 bool JeandleInlineTree::should_not_inline(JeandleCompilation* comp,
--- a/src/hotspot/share/jeandle/jeandleCompilation.cpp
+++ b/src/hotspot/share/jeandle/jeandleCompilation.cpp
@@ -124,6 +124,91 @@ static void print_inline_tree_method(outputStream* out, ciMethod* method) {
   }
 }
 
+const char* jeandle_inline_reason_name(JeandleInlineReason reason) {
+  switch (reason) {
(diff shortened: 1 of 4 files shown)


src/hotspot/share/jeandle/jeandleCompilation.cpp
```diff
@@ -579,35 +672,49 @@ bool JeandleInlineTree::should_inline(JeandleCompilation* comp,
...
     return false;
   }
 
+  if (size > max_inline_size) {
+    reason = max_inline_size == FreqInlineSize ? JeandleInlineReason::HotMethodTooBig :
+                                                 JeandleInlineReason::TooBig;
+    return false;
+  }
+  reason = JeandleInlineReason::InlineHot;
   return size <= max_inline_size;
```
taofengliu:
```suggestion
  return true;
```

taofengliu (approved):

Issue #536, opened 2026-07-08
Intrinsics are being implemented


eapenzhan:
Some **intrinsics** are being implemented by my colleagues. I'll list it here, in case of work conflicts.
- [x] revertBytes
- [ ] String(String)
- [ ] String.indexOf
- [ ] String.compareTo
- [ ] String.equals
- [ ] ArrayCopy
- [ ] _floatToIntBits _doubleToLongBits
- [ ] _dcopySign _fcopySign
- [ ] _floatIsInfinite _doubleIsInfinite _floatIsFinite _doubleIsFinite
- [ ] _vectorizedMismatch
- [ ] _compressStringC _compressStringB
- [ ] _inflateStringC _inflateStringB
- [ ] valueOf系列
- [ ] Unsafe CAS系列
- [ ] Unsafe Get Acquire系列
- [ ] StringCoding.countPositives
- [ ] Unsafe weak CAS 系列

eapenzhan:
@taofengliu

Cruise20:
@eapenzhan 
I'm developing the following unsafe intrinsics, to avoid work conflicts.

| Type | Summary | Done | TODO |
|------|------|----------|----------|
| Fence / Barrier | 6 | 3 | 3 |
| Cache Writeback | 3 | 1 | 2 |
| Plain Get/Put | 18 | 0 | 18 |
| Volatile Get/Put | 18 | 0 | 18 |
| Opaque Get/Put | 18 | 0 | 18 |
| Acquire/Release Get/Put | 18 | 0 | 18 |
| Unaligned Get/Put | 8 | 0 | 8 |
| compareAndSet  | 5 | 0 | 5 |
| compareAndExchange | 15 | 0 | 15 |
| weakCompareAndSet | 20 | 0 | 20 |
| getAndAdd | 4 | 0 | 4 |
| getAndSet | 5 | 0 | 5 |
| unsafe ld/store | 1 | 0 | 1 |
| allocation | 2 | 0 | 2 |
| park/unpark | 2 | 0 | 2 |
| **Total** | **141** | **4** | **137** |

PR #538, opened 2026-07-09
feat: support reverseBytes in Jeandle

eapenzhan:
## Related issue(s):

issue #536 

Jeandle did not lower `reverseBytes` directly, so these methods depended on the normal call/inlining path instead of emitting the expected LLVM IR at intrinsic lowering time.

This PR adds direct Jeandle intrinsic lowering for:

- `Integer.reverseBytes(int)` and `Long.reverseBytes(long)` via `llvm.bswap`
- `Character.reverseBytes(char)` with explicit zero-extension semantics
- `Short.reverseBytes(short)` with explicit sign-extension semantics

The narrow `char`/`short` cases need special handling because the JVM stack value is an `int`, but only the low 16 bits are part of the Java value. After byte swapping, `char` must be zero-extended while `short` must be sign-extended.

## Approach

- Register `_reverseBytes_i`, `_reverseBytes_l`, `_reverseBytes_c`, and `_reverseBytes_s` as supported Jeandle intrinsics.
- Lower full-width `int`/`long` cases directly to `llvm.bswap`.
- Add portable narrow lowering using `bswap.i16` followed by `zext`/`sext`.
- Add jtreg coverage for IR shape and `char`/`short` extension semantics.

## Performance

| bench(ns/op) | Jeandle ON | Jeandle Disable c/s | Jeandle Disable all | C2 default |
|---|---:|---:|---:|---:|
| `char` | 1.066 | 1.066 | 1.071 | 1.069 |
| `short` | 1.043 | 1.549 | 1.549 | 1.042 |
| `int` | 0.694 | 0.694 | 0.694 | 0.694 |
| `long` | 0.766 | 0.766 | 0.765 | 0.767 |

### DIFF
--- a/src/hotspot/cpu/aarch64/jeandleIntrinsicLowering_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/jeandleIntrinsicLowering_aarch64.cpp
@@ -20,7 +20,6 @@
 
 #include "jeandle/__llvmHeadersBegin__.hpp"
 #include "llvm/IR/IRBuilder.h"
-#include "llvm/IR/Intrinsics.h"
 #include "llvm/IR/IntrinsicsAArch64.h"
 
 #include "jeandle/jeandleAbstractInterpreter.hpp"
@@ -56,24 +55,6 @@ bool JeandleIntrinsicLowering::cpu_supports_spin_wait() {
 // =============================================================================
 // Arch-specific intrinsic lowering (AArch64)
 // =============================================================================
-
-bool JeandleIntrinsicLowering::lower_reverse_bytes_narrow_arch(vmIntrinsics::ID id) {
-  if (id != vmIntrinsics::_reverseBytes_c) {
-    return false;
-  }
-
-  llvm::IRBuilder<>& builder = _interp->_ir_builder;
-  llvm::Value* arg = _interp->_jvm->ipop();
-  llvm::Value* swapped =
-      builder.CreateIntrinsic(builder.getInt32Ty(), llvm::Intrinsic::bswap, {arg});
-  // Character arguments are canonical zero-extended i32 values here; rotating
-  // the i32 bswap lets AArch64 select REV16 while preserving char semantics.
-  llvm::Value* result =
-      builder.CreateIntrinsic(builder.getInt32Ty(), llvm::Intrinsic::fshr,
-                              {swapped, swapped, builder.getInt32(16)});
-  _interp->_jvm->ipush(result);
-  return true;
-}
 
 bool JeandleIntrinsicLowering::lower_spin_wait_hint() {
   llvm::IRBuilder<>& builder = _interp->_ir_builder;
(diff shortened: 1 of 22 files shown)


src/hotspot/cpu/x86/jeandleIntrinsicLowering_x86.cpp
```diff
@@ -56,6 +57,23 @@ bool JeandleIntrinsicLowering::cpu_supports_spin_wait() {
 // Arch-specific intrinsic lowering (x86)
 // =============================================================================
 
+bool JeandleIntrinsicLowering::lower_reverse_bytes_narrow_arch(vmIntrinsics::ID id) {
+  if (id != vmIntrinsics::_reverseBytes_c && id != vmIntrinsics::_reverseBytes_s) {
+    return false;
+  }
```
Cruise20: `assert` is better

src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
```diff
@@ -626,6 +643,33 @@ bool JeandleIntrinsicLowering::lower_count_zeros(vmIntrinsics::ID id,
...
+// ---- lower_reverse_bytes_narrow ----
+// Character.reverseBytes(char) / Short.reverseBytes(short). The value sits on
+// the operand stack as a computational int, but only the low 16 bits are
+// meaningful. Prefer target-specific C2-like IR when it can preserve those
+// semantics; otherwise use the portable i16 swap plus zero/sign extension.
+bool JeandleIntrinsicLowering::lower_reverse_bytes_narrow(vmIntrinsics::ID id) {
+  if (lower_reverse_bytes_narrow_arch(id)) {
+    return true;
+  }
+  return lower_reverse_bytes_narrow_portable(id);
```
Cruise20: Is this fallback necessary ?

src/hotspot/cpu/aarch64/jeandleIntrinsicLowering_aarch64.cpp
```diff
@@ -56,6 +57,24 @@ bool JeandleIntrinsicLowering::cpu_supports_spin_wait() {
 // Arch-specific intrinsic lowering (AArch64)
 // =============================================================================
 
+bool JeandleIntrinsicLowering::lower_reverse_bytes_narrow_arch(vmIntrinsics::ID id) {
+  if (id != vmIntrinsics::_reverseBytes_c) {
+    return false;
+  }
```
Cruise20: Why `_reverseBytes_s` is not handled here and fall back to `lower_reverse_bytes_narrow_portable`?

src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
eapenzhan: The logic has been modified. C2-like instructions are proved unnecessary.

taofengliu (approved):

Issue #483, opened 2026-05-23
last_Java_sp not cleared on exception path in call_VM stub


taofengliu:
Description: 
In jeandleCallVM.cpp, the `forward_exception_block` (taken when a pending exception is detected) does NOT clear `last_Java_sp`. Only the `no_exception_block` clears it. Stale `last_Java_sp` can cause invalid stack walks during GC.
 
Fix: 
Move `last_Java_sp`/`last_Java_pc` clearing before the exception check.

Refer to: C2's `GraphKit::gen_stub`

wryyyds7:
I think I can deal with it——Following C2's GraphKit::gen_stub, I will movedthe last_Java_sp/last_Java_pc clearing before the exception check so that both the exception and no-exception paths see a clean anchor, preventing stale values from causing invalid stack walks during GC.

PR #545, opened 2026-07-16
fix: clear last_Java_sp/last_Java_pc before exception check in call_VM stub (#483)

wryyyds7:
The forward_exception_block path in generate_call_VM did not clear
last_Java_sp, only the no_exception_block did. Stale last_Java_sp
can cause invalid stack walks during GC.

Move last_Java_sp/last_Java_pc clearing before the exception check
so both paths see a clean anchor, matching C2's GraphKit::gen_stub.


wryyyds7:
#483

derrickCcc:
#483

wenzeyu8888-rgb:
#483

Issue #479, opened 2026-05-23
Missing IllegalMonitorStateException in shared_unlock


taofengliu:
**Describe the bug**
The slow-path unlock call uses a regular `CallInst` not `InvokeInst`, so exceptions from `complete_monitor_unlocking_C` are not caught by Java exception handlers.

**Fix**
Wrap the slow-path call with `create_call_ex()` or verify the runtime call cannot throw to Java code.

wryyyds7:
I can try it.

wryyyds7:
The `langtools/tier1` failure also occurs on the `main` branch (run [29645266685](https://github.com/jeandle/jeandle-jdk/actions/runs/29645266685)), so it appears to be a pre-existing issue unrelated to this PR.

PR #546, opened 2026-07-16
fix: use InvokeInst for shared_unlock slow path to enable exception handling

wryyyds7:
### Problem
The slow-path unlock call in `shared_unlock` used `CallInst` instead of `InvokeInst`, so `IllegalMonitorStateException` from `complete_monitor_unlocking_C` could not be caught by Java exception handlers.

### Fix
Replace `_ir_builder.CreateCall` with `create_call_ex()` in the `shared_unlock` slow path. This generates an `InvokeInst` with a proper unwind destination (landing pad), allowing exceptions to be dispatched to Java exception handlers via `dispatch_exception_for_invoke`.

### Details
- `create_call_ex` internally calls `dispatch_exception_for_invoke()` to create both `normal_dest` and `unwind_dest` blocks
- The `unwind_dest` block contains a `LandingPadInst` that reads the pending exception and dispatches it to the correct Java catch handler based on the current BCI's exception handler table
- The `normal_dest` block continues execution flow to `monitor_exited` as before
- Calling convention (`llvm::CallingConv::C`) is set by `create_call_ex` internally

### DIFF
--- a/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
+++ b/src/hotspot/share/jeandle/jeandleAbstractInterpreter.cpp
@@ -3623,8 +3623,9 @@ void JeandleAbstractInterpreter::shared_unlock(LockValue lock) {
   _ir_builder.SetInsertPoint(monitorexit_slow_path);
   llvm::FunctionCallee monitorexit_callee = JeandleRuntimeRoutine::SharedRuntime_complete_monitor_unlocking_C_callee(_module);
   llvm::CallInst* current_thread = call_java_op("jeandle.current_thread", {});
-  llvm::CallInst* call_monitorexit = _ir_builder.CreateCall(monitorexit_callee, {lock.object().value(), lock.lock(), current_thread});
-  call_monitorexit->setCallingConv(llvm::CallingConv::C);
+  // Use create_call_ex (InvokeInst) so that IllegalMonitorStateException from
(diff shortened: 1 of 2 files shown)


wryyyds7:
#479

wryyyds7:
After investigation, the approach of wrapping the slow-path call with `create_call_ex()` (InvokeInst) is not feasible because it causes a compile-time crash: `dispatch_exception_for_invoke` → `dispatch_exception_to_handler` → rethrow path → `shared_unlock` corrupts interpreter state (`_block` and `_jvm`), since `dispatch_exception_for_invoke` only saves/restores the IRBuilder position but not `_block`/`_jvm`.

Instead, I verified that `complete_monitor_unlocking_C` is declared as `JRT_LEAF` and uses `ExceptionMark` internally, so it cannot produce Java-visible exceptional control flow. Using `CallInst` is therefore safe — this addresses the second option in the issue ("verify the runtime call cannot throw to Java code").

The final change is a documentation comment explaining why `CallInst` is safe. No code logic is changed, so CI should pass cleanly.

wryyyds7:
The `langtools/tier1` failure also occurs on the `main` branch (run [29645266685](https://github.com/jeandle/jeandle-jdk/actions/runs/29645266685)), so it appears to be a pre-existing issue unrelated to this PR.

derrickCcc:
#479

wenzeyu8888-rgb:
#479

Issue #548, opened 2026-07-17
feat: support vectorizedMismatch intrinsic


DXwangg:
`jdk.internal.util.ArraysSupport.vectorizedMismatch` is the Java entry point of
the HotSpot `_vectorizedMismatch` intrinsic. It compares two array or raw-memory
ranges and returns the index of the first mismatching element, or `-1` when the
entire range is equal.

This change uses a hybrid design that combines inline LLVM IR with a HotSpot
platform stub. The path is selected by the number of bytes to compare:

```text
ArraysSupport.vectorizedMismatch
  -> Jeandle recognizes _vectorizedMismatch
  -> compute raw addresses from object + offset
  -> byte_length = zext(length) << zext(scale)
  -> byte_length < 16
       -> inline width-tiered 8/4/2/1-byte LLVM IR
  -> 16 <= byte_length < 64 and the target supports fixed-width SIMD
       -> inline <16 x i8> LLVM vector IR
  -> byte_length >= 64
       -> C leaf call
       -> StubRoutines::vectorizedMismatch
       -> AArch64 NEON stub
```

The short path is inlined to avoid stub-call overhead and to allow LLVM to
optimize it together with the surrounding call site. Medium-sized inputs use
fixed 128-bit vector IR, which the AArch64 LLVM backend lowers to NEON
instructions. Larger inputs continue to use the HotSpot stub, preserving its
64-byte and 256-byte chunked loops and stable throughput for large ranges.

PR #549, opened 2026-07-17
feat: support vectorizedMismatch intrinsic

DXwangg:
## Related issue(s):

issue #548 

Implement _vectorizedMismatch intrinsic support in Jeandle with a size-tiered
LLVM IR and platform-stub design. Inputs below 16 bytes use width-tiered
8/4/2/1-byte integer IR,
16-63-byte inputs use fixed 128-bit vector IR on supported targets, and larger
inputs use StubRoutines::vectorizedMismatch. The AArch64 stub compares
256-byte blocks with NEON and precisely locates the first differing element.

This avoids stub overhead and byte-wise loops for small inputs, enables NEON lowering for medium
inputs, and preserves the optimized stub loop for large arrays. The intrinsic
is enabled by default only when Jeandle is enabled on AArch64 and remains
user-controllable through UseVectorizedMismatchIntrinsic.

The change includes Jeandle-specific semantic and IR coverage, boundary tests
for all three tiers, and a direct JMH benchmark for
ArraysSupport.vectorizedMismatch.

### DIFF
--- a/src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
+++ b/src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
@@ -22,6 +22,7 @@
 #include "jeandle/__llvmHeadersBegin__.hpp"
 #include "llvm/Analysis/ConstantFolding.h"
 #include "llvm/IR/Constants.h"
+#include "llvm/IR/DerivedTypes.h"
 #include "llvm/IR/MDBuilder.h"
 
 #include "jeandle/jeandleAbstractInterpreter.hpp"
@@ -39,6 +40,7 @@
 #include "oops/klass.hpp"
 #include "runtime/deoptimization.hpp"
 #include "runtime/globals.hpp"
+#include "utilities/globalDefinitions.hpp"
 
 // =============================================================================
 // Call-site IR annotation helpers (migrated from JeandleIntrinsicIRSemantics)
@@ -93,6 +95,9 @@ bool JeandleIntrinsicLowering::is_supported(vmIntrinsics::ID id) {
 
     case vmIntrinsics::_onSpinWait:
       return cpu_supports_spin_wait();
+
+    case vmIntrinsics::_vectorizedMismatch:
+      return UseVectorizedMismatchIntrinsic;
 
     default: break;
   }
@@ -370,6 +375,9 @@ bool JeandleIntrinsicLowering::lower(vmIntrinsics::ID id, const ciMethod* target
     case vmIntrinsics::_PhantomReference_refersTo0:
       return lower_java_op("jeandle.reference_refers_to",
                            {CTRL_NONE, MEM_READ | MEM_NEEDS_GC_STATE});
+
+    case vmIntrinsics::_vectorizedMismatch:
+      return lower_vectorized_mismatch();
 
     // newArray
     case vmIntrinsics::_newArray:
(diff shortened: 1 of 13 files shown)


taofengliu (commented):
Some code is too dense; it needs better spacing to improve readability, along with proper comments.

DXwangg:
> Some code is too dense; it needs better spacing to improve readability, along with proper comments.

Updated

src/hotspot/share/jeandle/jeandleIntrinsicLowering.hpp
```diff
@@ -133,6 +133,7 @@ class JeandleIntrinsicLowering : public StackObj {
   static bool cpu_supports_rounding();          // floor/ceil/rint
   static bool cpu_supports_popcount();          // bitCount_i/bitCount_l
   static bool cpu_supports_spin_wait();         // onSpinWait
+  static bool cpu_supports_string_simd();       // >=128-bit integer SIMD
```
taofengliu: The name is a bit misleading. It only controls whether we inline the medium path for this intrinsic.

src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
```diff
@@ -802,6 +810,261 @@ bool JeandleIntrinsicLowering::lower_new_array() {
...
+  llvm::Value* a_addr = b.CreateGEP(i8, a_obj, a_offset, "mismatch_a_addr");
+  llvm::Value* b_addr = b.CreateGEP(i8, b_obj, b_offset, "mismatch_b_addr");
+
+  llvm::LLVMContext& ctx = *_interp->_context;
+  llvm::Type* i32 = b.getInt32Ty();
+  llvm::Type* i64 = b.getInt64Ty();
+  llvm::Function* f = _interp->_llvm_func;
+
+  static constexpr unsigned small_path_limit = 16;
+  static constexpr unsigned medium_path_limit = 64;
```
taofengliu: Can we use ArrayOperationPartialInlineSize like C2 to make it more flexible?

src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
```diff
@@ -802,6 +810,261 @@ bool JeandleIntrinsicLowering::lower_new_array() {
...
+  b.SetInsertPoint(dispatch_bb);
+  if (cpu_supports_string_simd()) {
+    b.CreateCondBr(is_medium, medium_bb, stub_bb);
+  } else {
+    b.CreateBr(stub_bb);
+  }
+
+  // Tier 2: inline 128-bit vector IR for byte lengths in [16, 64).
+  b.SetInsertPoint(medium_bb);
+  llvm::Value* medium_result = emit_vectorized_mismatch_medium(a_addr, b_addr, byte_length, scale64);
```
taofengliu: Don't emit IR when we don't need it (for x86 and RISC-V)

src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
```diff
@@ -802,6 +810,261 @@ bool JeandleIntrinsicLowering::lower_new_array() {
...
+  b.SetInsertPoint(small_bb);
+  llvm::Value* small_result = emit_vectorized_mismatch_small(a_addr, b_addr, byte_length, scale64);
+  llvm::BasicBlock* small_done_bb = b.GetInsertBlock();
+  b.CreateBr(done_bb);
+
+  // Tier 2 is available only when the target can lower the fixed-width vector
+  // IR efficiently. Unsupported targets skip directly to the platform stub.
+  b.SetInsertPoint(dispatch_bb);
+  if (cpu_supports_string_simd()) {
+    b.CreateCondBr(is_medium, medium_bb, stub_bb);
```
taofengliu: Why are we inlining the medium path only on AArch64? Is it solely to reduce stub calls? But we also need to consider the IR size.

test/hotspot/jtreg/compiler/intrinsics/VectorizedMismatchTest.java
```diff
@@ -25,6 +25,7 @@
 
 /*
  * @test
+ * @requires vm.opt.final.UseJeandleCompiler == false
```
taofengliu: Maybe don't need this. Just run it.

test/jtreg-ext/requires/VMProps.java
```diff
@@ -382,6 +382,7 @@ protected void vmOptFinalFlags(SafeMap map) {
         vmOptFinalFlag(map, "EliminateAllocations");
         vmOptFinalFlag(map, "UnlockExperimentalVMOptions");
         vmOptFinalFlag(map, "UseCompressedOops");
+        vmOptFinalFlag(map, "UseJeandleCompiler");
```
taofengliu: Maybe don't need this.

taofengliu (commented):
I generally prefer implementing complex IR in template.ll, but the implementation in this PR is already fairly mature. For future PRs, we can consider generating complex IR in template.ll.

src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
DXwangg: ArrayOperationPartialInlineSize is not directly equivalent here. C2 uses it as the size of a single masked-vector partial-inline operation, and AArch64 sets it to 0 in NEON-only mode. Jeandle's medium path is a loop of fixed 128-bit loads and is intentionally enabled on baseline NEON. Using the C2 flag directly would disable this path on NEON and would also change the current 64-byte boundary semantics. I suggest keeping a separate threshold here.

src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
DXwangg: We keep the medium path on AArch64 because NEON is part of the architecture baseline, and benchmarks show that inlining significantly reduces the stub-call overhead for medium-sized full scans. On x86 and RISC-V, the medium-path IR is not emitted and execution falls back directly to the platform stub.

test/hotspot/jtreg/compiler/intrinsics/VectorizedMismatchTest.java
DXwangg: Agreed. The restriction was masking the x86-specific UseAVX options on AArch64. I removed it and added -XX:+IgnoreUnrecognizedVMOptions to both runs, so the test can run with Jeandle on both AArch64 and x86. I also verified it passes with Jeandle enabled by default on AArch64.

test/jtreg-ext/requires/VMProps.java
DXwangg: Agreed.  This property was added for the @requires condition. Since that condition is no longer needed, I’ll remove this entry as well, together with the remaining corresponding @requires in the Jeandle-specific test.

src/hotspot/share/jeandle/jeandleIntrinsicLowering.hpp
DXwangg: updated

src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
DXwangg: updated

taofengliu (approved):

PR #558, opened 2026-07-23
fix: clear last_Java_sp/last_Java_pc on exception path in call_VM stub

### LINKED ISSUE #483: last_Java_sp not cleared on exception path in call_VM stub

aminuoshi-378:
## Related issue
Fixes #483

## Summary
Move the clearing of `last_Java_sp` and `last_Java_pc` from `no_exception_block` to before the exception check in `JeandleCallVM::generate_call_VM`, so both the exception and no-exception paths clear these values. This prevents stale `last_Java_sp` from causing invalid stack walks during GC.

## Changes
- Move `last_Java_sp`/`last_Java_pc` clearing before the exception check (refer to C2's `GraphKit::gen_stub`).
- No functional change to the no-exception path; the exception path now also clears these values.

## Testing
Build verified on macOS aarch64. Full jeandle testing requires Linux environment with jeandle feature enabled.

### DIFF
--- a/src/hotspot/share/jeandle/jeandleCallVM.cpp
+++ b/src/hotspot/share/jeandle/jeandleCallVM.cpp
@@ -83,6 +83,15 @@ void JeandleCallVM::generate_call_VM(const char* name, address routine_address, 
   call_routine_address->addFnAttr(id_attr);
   call_routine_address->addFnAttr(patch_bytes_attr);
 
+  // Clear the last_Java_sp and last_Java_pc before checking exceptions.
+  // This must be done on both the exception and no-exception paths to avoid
+  // stale values causing invalid stack walks during GC. See C2's GraphKit::gen_stub.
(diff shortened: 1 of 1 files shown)


aminuoshi-378:
Fix: Move the clearing of last_Java_sp and last_Java_pc to before the exception check, so both paths execute the clearing. This follows the same pattern as C2's GraphKit::gen_stub (see generateOptoStub.cpp:223-226), where the clearing is done unconditionally before the exception branch.

PR #566, opened 2026-07-24
feat: implement remaining Float/Double bit-conversion intrinsics (closes #554)

### LINKED ISSUE #554: Implement some type conversion intrinsics.
**The feature you'd like**
vmIntrinsics::_floatToRawIntBits
vmIntrinsics::_floatToIntBits
vmIntrinsics::_intBitsToFloat
vmIntrinsics::_doubleToRawLongBits
vmIntrinsics::_doubleToLongBits
vmIntrinsics::_longBitsToDouble
vmIntrinsics::_floatToFloat16
vmIntrinsics::_float16ToFloat

xugeeeee:
## Related issue(s)

issue #554

Issue #554 asked for 8 intrinsics: floatToRawIntBits, floatToIntBits,
intBitsToFloat, doubleToRawLongBits, doubleToLongBits, longBitsToDouble,
floatToFloat16, and float16ToFloat. Checking current jeandle main showed
4 of these (the Raw bitcast variants plus intBitsToFloat/longBitsToDouble)
were already implemented via lower_llvm_bitcast(). This PR adds the 4 that
were actually missing:

- Float.floatToIntBits(float) / Double.doubleToLongBits(double): like the
  raw variants, but must canonicalize every NaN input to the single NaN
  bit pattern the javadoc specifies, instead of preserving whatever NaN
  payload/sign the input happened to carry.
- Float.floatToFloat16(float) / Float.float16ToFloat(short): float16
  values are carried as a raw bit pattern in a Java short, so this is a
  real narrowing/widening fp convert through LLVM's `half` type, not a
  plain bitcast like the other 6.

## Approach

- floatToIntBits/doubleToLongBits (lower_fp_to_bits_canonical): arg != arg
  (unordered compare) is true only for NaN; select between the canonical
  NaN constant (0x7fc00000 / 0x7ff8000000000000L, matching the Java spec)
  and the plain bitcast result. Mirrors C2's
  LibraryCallKit::inline_fp_conversions exactly.
- floatToFloat16/float16ToFloat (lower_float16_convert): fptrunc/fpext
  through llvm's `half` type plus a bitcast to move between `half` and the
  integer bit pattern. No NaN special-casing -- C2's
  ConvF2HFNode/ConvHF2FNode don't canonicalize NaN for this conversion
  either. Java short is a computational-int JVM stack type (like the
  reverseBytes_s/_c narrow variants already in this file), so the i16 bits
  are sign-extended to i32 on push and truncated back from the popped i32
  on the way in.
- No CPU feature gating for any of the 4: disabled_by_jvm_flags() only
  checks the generic InlineMathNatives/InlineIntrinsics flags for these
  IDs, no hardware feature check (unlike fmaD/fmaF's UseFMA check).
  fptrunc/fpext to/from half always lower validly on every target Jeandle
  supports -- hardware conversion when available, otherwise LLVM inserts
  a __truncsfhf2/__extendhfsf2 libcall automatically.
- All 4 use hand-built IR (IRBuilder select/fcmp/fptrunc/fpext/bitcast)
  rather than emit_llvm_builtin(), since none of them is a single llvm.*
  intrinsic call.

## Test

Extended TestFloatDoubleBitConversion.java with floatToIntBits/
doubleToLongBits coverage: NaN canonicalization (including explicitly
constructed signaling-NaN and all-ones-payload bit patterns), non-NaN
bit-identity with the raw variants, and a 1000-iteration random-value
fuzz against the interpreter-run Float/Double reference methods.

Added TestFloat16.java: special values (signed zero, infinities, NaN),
exact round trip for exactly-representable values, overflow to infinity
past the binary16 max finite magnitude (65504), underflow to zero, and a
5000+5000 random fuzz across the full float32 bit space checked against
the interpreter-run Float.floatToFloat16/float16ToFloat (TestWrapper::main
is deliberately left out of the compileonly list so those interpreter
calls are a true independent oracle, not silently re-intrinsified by
compiling main() itself).

Verified: both new/extended tests pass, and the full
compiler/jeandle/intrinsic/ suite (34 tests) shows no regressions.

Signed-off-by: Hongxu Wang <wanghongxu2001@163.com>


xugeeeee:
hi,@taofeng.
Both failing checks (langtools/tier1, hs/tier1 compiler part 3) look unrelated to this PR — reproduced locally and traced to IntrinsicAvailableTest (CPU CRC32 support) and TestSignumVector (vector ISA), neither touching Jeandle code. hs/tier1 compiler jeandle and the full local compiler/jeandle/intrinsic/ suite (35/35) pass. Could you re-run / take a look? Thanks!

taofengliu (approved):

PR #572, opened 2026-07-30
feat: return constant field data as tuple

xiejf2020:

Adapt the HotSpot side to the structured VM callback interface.

`GetConstantFieldValue` and `GetConstantFieldInfo` are consolidated into `GetConstantField`, returning the shared `llvm::jeandle::ConstantFieldResult` tuple of `(BasicType, rawValue)`. This makes the two pieces of constant-field-folding information one replayable callback result.

The JDK callback declaration and definition both use the LLVM-owned result type, keeping the HotSpot and LLVM callback signatures in sync and providing a concrete example for future tuple-return callbacks.

### DIFF
--- a/src/hotspot/share/jeandle/jeandleVMCallback.hpp
+++ b/src/hotspot/share/jeandle/jeandleVMCallback.hpp
@@ -21,6 +21,9 @@
 #ifndef SHARE_JEANDLE_VM_CALLBACK_HPP
 #define SHARE_JEANDLE_VM_CALLBACK_HPP
 
+#include "jeandle/__llvmHeadersBegin__.hpp"
+#include "llvm/IR/Jeandle/VMCallback.h"
+
 #include "memory/allocation.hpp"
 #include <string>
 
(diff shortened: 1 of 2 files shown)


src/hotspot/share/jeandle/jeandleVMCallback.hpp
```diff
@@ -21,6 +21,9 @@
 #ifndef SHARE_JEANDLE_VM_CALLBACK_HPP
 #define SHARE_JEANDLE_VM_CALLBACK_HPP
 
+#include "jeandle/__llvmHeadersBegin__.hpp"
+#include "llvm/IR/Jeandle/VMCallback.h"
+
 #include "memory/allocation.hpp"
 #include <string>
```
taofengliu, line 27:
```suggestion
#include "jeandle/__llvmHeadersBegin__.hpp"
#include "llvm/IR/Jeandle/VMCallback.h"

#include <string>

#include "jeandle/jeandleAbstractInterpreter.hpp"
#include "memory/allocation.hpp"
```

taofengliu (commented):
Can we update `JeandleVMCallback::get_cha_opt_info` to use tuple too? Then remove the wrong comment above `optimize_invokeinterface` in jeandleVMCallback.cpp

src/hotspot/share/jeandle/jeandleVMCallback.hpp
xiejf2020: Perhaps we don't have to include jeandleAbstractInterpreter.hpp?

```c++
#include "jeandle/__llvmHeadersBegin__.hpp"
#include "llvm/IR/Jeandle/VMCallback.h"

#include <string>

#include "jeandle/__hotspotHeadersBegin__.hpp"
#include "memory/allocation.hpp"
```

taofengliu (approved):

PR #580, opened 2026-08-10
feat: implement Math exact-arithmetic and multiplyHigh intrinsics

xugeeeee:

Jeandle already had Math.addExact (_addExactI/_addExactL). This adds the
rest of the same overflow-checked-arithmetic family, plus the unrelated
but similarly-shaped wide-multiply family:

- Math.subtractExact(int,int)/(long,long)
- Math.multiplyExact(int,int)/(long,long)
- Math.incrementExact(int)/(long)
- Math.decrementExact(int)/(long)
- Math.negateExact(int)/(long)
- Math.multiplyHigh(long,long)
- Math.unsignedMultiplyHigh(long,long)

None of these were previously intrinsified, so they ran through the normal
call/inlining path instead of emitting the expected LLVM IR at intrinsic
lowering time.

## Approach

- subtractExact/multiplyExact/incrementExact/decrementExact/negateExact
  (lower_exact_arith): same overflow-trap shape as the existing
  lower_add_exact -- llvm.s{sub,mul}.with.overflow, branch on the overflow
  bit to an uncommon_trap (Reason_intrinsic/Action_none) so the interpreter
  re-executes and throws the real ArithmeticException. Args are peeked (not
  popped) before the branch, same as addExact, so the deopt bundle captures
  the correct pre-call stack for re-execution.
  - increment/decrement/negate are unary in Java but all three reduce to
    the with-overflow intrinsic on a synthesized second operand: a+1, a-1,
    and 0-a respectively. Negation overflows exactly when a == MIN_VALUE,
    which is exactly when 0-a overflows the signed range, so
    ssub.with.overflow(0, a) detects it correctly with no separate check.
  - One shared lower_exact_arith(id, overflow_id) handles all 10 IDs rather
    than duplicating the overflow-trap control flow 5 times.
  - No new CPU/flag gating needed: UseMathExactIntrinsics/InlineMathNatives
    are already enforced by the shared vmIntrinsics::disabled_by_jvm_flags(),
    which try_lower_intrinsic() calls unconditionally after is_supported()
    -- the same reasoning already documented for addExact/fmaD/fmaF.
- multiplyHigh/unsignedMultiplyHigh (lower_multiply_high): sign-extend
  (resp. zero-extend) both i64 operands to i128, multiply, and shift right
  64. This is the canonical wide-multiply idiom LLVM's instruction
  selection recognizes and lowers to a single hardware mul-high instruction
  (e.g. imulq/mulq on x86-64) rather than a real 128-bit multiply routine.
  No overflow/exception semantics here (these never throw), and no flag
  gating either -- disabled_by_jvm_flags() doesn't gate these two IDs at
  all, unlike e.g. CRC32/AES/FMA.

## Test

Added TestSubtractExact.java, TestMultiplyExact.java, TestIncrementExact.java,
TestDecrementExact.java, TestNegateExact.java (boundary values + overflow
ArithmeticException cases for both int/long, plus FileCheck verification of
the llvm.s{sub,mul}.with.overflow IR and block labels, following the existing
TestAddExact.java pattern), and TestMultiplyHigh.java (boundary values plus a
5000-iteration random fuzz across the full long range, checked against the
interpreter-run Math.multiplyHigh/unsignedMultiplyHigh -- TestWrapper::main is
deliberately left out of the compileonly list so those interpreter calls are
a true independent oracle, not silently re-intrinsified by compiling main()
itself).

Verified: all 6 new tests pass, and the full compiler/jeandle/intrinsic/
suite (41 tests) shows no regressions.

Signed-off-by: Hongxu Wang <wanghongxu2001@163.com>


src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
```diff
@@ -415,6 +449,28 @@ bool JeandleIntrinsicLowering::lower(vmIntrinsics::ID id, const ciMethod* target
...
 
+    // subtractExact/decrementExact/negateExact all reduce to
+    // llvm.ssub.with.overflow (see lower_exact_arith).
+    case vmIntrinsics::_subtractExactI:
+    case vmIntrinsics::_subtractExactL:
+    case vmIntrinsics::_decrementExactI:
+    case vmIntrinsics::_decrementExactL:
+    case vmIntrinsics::_negateExactI:
+    case vmIntrinsics::_negateExactL:
+      return lower_exact_arith(id, llvm::Intrinsic::ssub_with_overflow);
```
taofengliu, line 461: Can we use `lower_exact_arith` to handle `_addExact`?

src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
```diff
@@ -197,6 +197,30 @@ bool JeandleIntrinsicLowering::is_supported(vmIntrinsics::ID id) {
     // addExact
     case vmIntrinsics::_addExactI:
     case vmIntrinsics::_addExactL:
```
taofengliu, line 207: Merge to the Math.{subtract,multiply,increment,decrement,negate}Exact below

taofengliu (commented):
Add some fuzz tests

src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
```diff
@@ -1147,31 +1203,62 @@ llvm::Value* JeandleIntrinsicLowering::emit_vectorized_mismatch_medium(
...
+// llvm.s{sub,mul}.with.overflow, branch on the overflow bit to an
+// uncommon_trap (Reason_intrinsic/Action_none) so the interpreter
+// re-executes. Args are peeked (not popped) before the branch for the same
+// deopt re-execution reason.
+//
+// increment/decrement/negate are unary in Java but all three reduce to the
+// with-overflow intrinsic on a synthesized second operand: a+1, a-1, and 0-a
+// respectively -- negation overflows exactly when a == MIN_VALUE, which is
+// exactly when 0-a overflows the signed range, so ssub.with.overflow(0, a)
+// detects it correctly without a separate check.
```
taofengliu, line 1217:
```suggestion
// ---- lower_exact_arith ----
// Math.addExact/subtractExact/multiplyExact/incrementExact/decrementExact/negateExact
// llvm.s{sub,mul}.with.overflow, branch on the overflow bit to an
// uncommon_trap (Reason_intrinsic/Action_none) so the interpreter
// re-executes. Args are peeked (not popped) before the branch for the same
// deopt re-execution reason.
//
// increment/decrement/negate are unary in Java but all three reduce to the
// with-overflow intrinsic on a synthesized second operand: a+1, a-1, and 0-a
// respectively -- negation overflows exactly when a == MIN_VALUE, which is
// exactly when 0-a overflows the signed range, so ssub.with.overflow(0, a)
// detects it correctly without a separate check.
```

src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
```diff
@@ -197,9 +197,32 @@ bool JeandleIntrinsicLowering::is_supported(vmIntrinsics::ID id) {
     case vmIntrinsics::_reverseBytes_l:
     case vmIntrinsics::_reverseBytes_s:
     case vmIntrinsics::_reverseBytes_c:
-    // addExact
+
+    // Math.{add,subtract,multiply,increment,decrement,negate}Exact:
+    // UseMathExactIntrinsics and InlineMathNatives are enforced by
+    // vmIntrinsics::disabled_by_jvm_flags(), which the caller
```
```suggestion
    // vmIntrinsics::is_disabled_by_flags(), which the caller
```

taofengliu (approved):

Issue #518, opened 2026-06-18
Implement intrinsics


Cruise20:
As listed here: https://github.com/jeandle/jeandle-jdk/issues/536#issuecomment-4922103626

PR #582, opened 2026-08-13
feat: Implement Unsafe.allocateInstance intrinsic

minchao1024:
## Related issue(s):

issue #518

Implement Jeandle lowering for Unsafe.allocateInstance.

- Use a uniform mirror-to-Klass load for constant and dynamic Class mirrors
- Handle primitive mirrors through re-execution traps
- Preserve class-initialization and layout-helper slow-path checks
- Route allocation through the Jeandle new_instance JavaOp
- Extend new_instance with an initial slow-path test
- Allow the allocation runtime routine to accept any Klass


src/hotspot/share/jeandle/jeandleRuntimeRoutine.hpp
```diff
@@ -404,7 +404,7 @@ class JeandleRuntimeRoutine : public AllStatic {
   static address search_landingpad(JavaThread* current);
 
   // Array allocation routine
```
taofengliu:
```suggestion
  // Allocation routine
```

taofengliu (commented):
Add some tests

src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
```diff
@@ -879,6 +889,96 @@ bool JeandleIntrinsicLowering::lower_new_array() {
   return true;
 }
 
+bool JeandleIntrinsicLowering::lower_unsafe_allocate_instance() {
+  llvm::LLVMContext& ctx = *_interp->_context;
+  llvm::IRBuilder<>& builder = _interp->_ir_builder;
+  llvm::Module& module = _interp->_module;
```
```suggestion
  llvm::Module& m = _interp->_module;
```
module is a C++ keyword, so avoid using it as a variable name.

taofengliu (dismissed):

taofengliu (approved):

PR #586, opened 2026-08-20
feat: support profile-guided virtual call devirtualization

lfeng14:
## Related issue(s):

issue #

Adds profile-guided devirtualization for Jeandle `invokevirtual` and `invokeinterface` calls. HotSpot reads receiver profiles and resolves concrete targets, while LLVM emits receiver guards and rewrites the selected paths to direct calls.

```text
virtual call
  |
  v
Can CHA determine the target?
  | yes
  +------> CHA guard + direct call
  |
  no
  v
Read receiver profile
  |
  +------> monomorphic
  |           receiver0 -> direct call
  |           miss      -> uncommon trap or virtual call
  |
  +------> bimorphic
  |           receiver0 -> direct call target0
  |           receiver1 -> direct call target1
  |           miss      -> uncommon trap or virtual call
  |
  +------> major receiver
  |           receiver0 -> direct call
  |           other     -> virtual call
  |
  +------> no reliable prediction
              keep the virtual call
```

The miss path uses an uncommon trap until the C2-style trap or recompilation limit is reached, after which it falls back to the original virtual call.

## Main changes
- Select and validate monomorphic, bimorphic, and major-receiver targets from mature profiles.
- Recover the call-site context from LLVM IR and deopt scopes, following CHA; CallSiteInfo remains responsible only for runtime metadata and statepoint tracking.
- Preserve receiver type information after rewriting guarded calls so later type-based optimizations can still use it.
- Preserve class-unloading and accessor information, validate profile counts, and update JDK call-site metadata before rewriting LLVM IR.
- Add JeandleUseProfiledVirtualCallDevirtualization to control the optimization.

## Correctness
- Include `ciMethod` identity in LLVM symbols to distinguish same-named methods from different class loaders.

## Performance
Compared profile devirtualization ON and OFF on 25 Renaissance benchmarks using the release build. Each configuration ran five times on the same NUMA node, with the last three runs averaged. In the reported run, 21 benchmarks improved and four regressed; the geometric-mean improvement was 10.86% and total runtime decreased by 6.16%. The individual regressions were not stable across follow-up runs, and the earlier 8.42% regression reported for `philosophers` could not be reproduced.

### DIFF
--- a/src/hotspot/share/jeandle/jeandleProfile.cpp
+++ b/src/hotspot/share/jeandle/jeandleProfile.cpp
@@ -77,7 +77,7 @@ bool JeandleProfile::has_too_many_recompiles(
   uint method_cutoff = static_cast<uint>(PerMethodRecompilationCutoff) / 2 + 1;
   Deoptimization::DeoptReason per_bc_reason =
       Deoptimization::reason_recorded_per_bytecode_if_any(reason);
-  ciMethod *trap_method =
+  ciMethod* trap_method =
       Deoptimization::reason_is_speculate(reason) ? _method : nullptr;
 
   if ((per_bc_reason == Deoptimization::Reason_none ||
@@ -92,10 +92,10 @@ bool JeandleProfile::has_too_many_recompiles(
          compilation->decompile_count() >= method_cutoff;
 }
 
-static ciMethod *resolve_profile_virtual_target(ciMethod *caller,
-                                                ciMethod *callee,
-                                                ciInstanceKlass *holder,
-                                                ciKlass *receiver) {
+static ciMethod* resolve_profile_virtual_target(ciMethod* caller,
+                                                ciMethod* callee,
+                                                ciInstanceKlass* holder,
+                                                ciKlass* receiver) {
   if (receiver == nullptr) {
     return nullptr;
   }
@@ -112,7 +112,7 @@ static ciMethod *resolve_profile_virtual_target(ciMethod *caller,
     return nullptr;
   }
 
-  ciInstanceKlass *receiver_klass = receiver->as_instance_klass();
+  ciInstanceKlass* receiver_klass = receiver->as_instance_klass();
   if (!receiver_klass->is_loaded() || !receiver_klass->is_initialized() ||
       receiver_klass->is_interface() ||
       !receiver_klass->is_subtype_of(holder)) {
@@ -122,7 +122,7 @@ static ciMethod *resolve_profile_virtual_target(ciMethod *caller,
 }
 
 JeandleProfile::DevirtualizationInfo
-JeandleProfile::devirtualization_at(ciMethod *callee, ciInstanceKlass *holder,
+JeandleProfile::devirtualization_at(ciMethod* callee, ciInstanceKlass* holder,
                                     int bci) const {
   if (_method == nullptr || callee == nullptr || holder == nullptr ||
       callee->can_be_statically_bound() || !is_mature() ||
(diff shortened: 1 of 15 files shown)


src/hotspot/share/jeandle/jeandleProfile.cpp
```diff
@@ -38,6 +48,192 @@ bool JeandleProfile::is_mature() const {
...
+  }
+
+  ciKlass* receiver2 = nullptr;
+  ciMethod* target2 = nullptr;
+  int64_t receiver_count2 = 0;
+  if (bimorphic_candidate) {
+    if (!call_profile.has_receiver(1) || call_profile.receiver_count(1) <= 0) {
+      if (!has_major_receiver) {
+        return {};
+      }
```
taofengliu: Why we return null here? 
Seems C2 does not have similar path.

src/hotspot/share/jeandle/jeandleProfile.cpp
```diff
@@ -38,6 +48,192 @@ bool JeandleProfile::is_mature() const {
...
+    } else {
+      receiver2 = call_profile.receiver(1);
+      target2 = resolve_target(receiver2);
+      if (target2 == nullptr) {
+        if (!has_major_receiver) {
+          return {};
+        }
```
taofengliu: Why we return null here?
Seems C2 does not have similar path.

src/hotspot/share/jeandle/jeandleProfile.cpp
```diff
@@ -38,6 +48,192 @@ bool JeandleProfile::is_mature() const {
...
+  if (_mdo == nullptr || _mdo->is_empty()) {
+    return false;
+  }
+
+  uint bc_cutoff = static_cast<uint>(PerBytecodeRecompilationCutoff) / 8;
+  uint method_cutoff = static_cast<uint>(PerMethodRecompilationCutoff) / 2 + 1;
+  Deoptimization::DeoptReason per_bc_reason =
+      Deoptimization::reason_recorded_per_bytecode_if_any(reason);
+  ciMethod* trap_method =
+      Deoptimization::reason_is_speculate(reason) ? _method : nullptr;
```
taofengliu: `Deoptimization::reason_is_speculate(reason)` is always false.

src/hotspot/share/jeandle/jeandleVMCallback.cpp
```diff
@@ -810,6 +812,98 @@ bool JeandleVMCallback::update_call_site(int64_t id, int dest, bool need_attache
   return true;
 }
 
+// ---------------------------------------------------------------------------
+// Profile-guided devirtualization
+// ---------------------------------------------------------------------------
+
+static void record_profile_receiver(ciKlass* receiver) {
```
taofengliu: Merge main branch, there's a `record_klass_metadata` function we can use.

src/hotspot/share/jeandle/jeandleVMCallback.cpp
```diff
@@ -487,6 +490,7 @@ bool JeandleVMCallback::record_inline_result(int scope_id, int bci, uintptr_t ca
     // IsOkToInline has already prepared this tree, so a successful result only
     // commits metadata in the same successful-inline order as LLVM's InlineScopes.
     comp->commit_inline_tree_for_callee(scope_id, bci, callee);
+    comp->accumulate_trap_counts_from_mdo(callee);
```
taofengliu: If we accumulate trap counts when inline success, the abstract interpreter of inlinee cannot see the accumulated trap counts.

src/hotspot/share/jeandle/jeandleVMCallback.cpp
lfeng14: Fixed. merged the latest main, removed record_profile_receiver(), and reused record_klass_metadata().

src/hotspot/share/jeandle/jeandleVMCallback.cpp
lfeng14: Fixed. Trap counts are now accumulated in JeandleAbstractInterpreter::interpret() before each root or inlinee method is parsed, restoring the existing behavior. The accumulation after a successful inline has been removed.

src/hotspot/share/jeandle/jeandleProfile.cpp
lfeng14: Profile devirtualization currently uses only Reason_class_check and Reason_bimorphic, so this condition is not reached. It is kept for future speculative support and uses the compilation root method, as in C2.

taofengliu (approved):

PR #589, opened 2026-08-21
feat: add getObjectSize intrinsic

eapenzhan:
## Related issue(s):

N/A

`Instrumentation.getObjectSize` currently falls back to the native entry when compiled by Jeandle. This PR adds `_getObjectSize` lowering with the same layout-helper calculation used by C1 and C2.

The lowering:

- uses the canonical `jeandle.load_klass`, `jeandle.layout_helper`, and `jeandle.arraylength` JavaOps so metadata and constant folding remain available;
- preserves the full category-2 operand stack through the null-check path;
- computes the layout-helper approximation for instances and the aligned runtime size for arrays;
- leaves the intrinsic-disabled native fallback unchanged.

It also adds enabled/disabled path, IR, object-code, semantic, and focused JMH coverage.

### Verification

Validated on AArch64 Linux against `performance` at `2ac1206b47`, using Jeandle LLVM `performance` at `149ba00573`:

- fastdebug and release `images` builds: passed;
- `compiler/jeandle/intrinsic/TestGetObjectSize.java`: 1/1 passed;
- `serviceability/jvmti/GetObjectSizeClass.java`: 1/1 passed with Jeandle enabled;
- intrinsic wiring and diff checks: passed;
- independent cold review: no blocking findings.

`java/lang/instrument/GetObjectSizeIntrinsicsTest.java` reports 2/9 on this branch. The seven failures occur first in the test's `-Xint` native action: Jeandle policy disables compressed oops/class pointers, while the test hardcodes a 16-byte array header and expects 4112 bytes instead of the native result of 4120. The failure occurs before JIT or intrinsic execution.

### Performance

AArch64 release JMH, 3 interleaved forks, 5 x 1 s warmup, 8 x 1 s measurement, one pinned CPU, SerialGC, 256 MiB heap, and compressed oops/class pointers disabled consistently in all modes. Lower is better.

| Shape | Jeandle intrinsic | Jeandle disabled | C2 |
| --- | ---: | ---: | ---: |
| instance | 3.285 ns/op | 82.690 ns/op | 3.028 ns/op |
| byte[0] | 3.619 ns/op | 82.974 ns/op | 3.300 ns/op |
| byte[1024] | 3.628 ns/op | 83.049 ns/op | 3.316 ns/op |
| Object[1024] | 3.953 ns/op | 83.211 ns/op | 3.344 ns/op |

All matrices completed without runner warnings; fork CV was at most 1.361%. The pinned CPU used the `performance` governor, and its physical-core sibling averaged 99.949% idle with no nice load during the monitored run. The intrinsic is 21.05x-25.18x faster than the Jeandle native fallback and remains 8.49%-18.20% slower than C2 for these shapes.


eapenzhan:
It's better after #590

Issue #605, opened 2026-08-27
feat: support late inlining in Jeandle


minchao1024:
**The feature you'd like**

Jeandle should support postponing selected inline candidates until after early optimization passes.

This allows methods that benefit from a cleaner intermediate representation—such as selected string, boxing, vector, replay-directed, and oversized `@ForceInline` methods—to be considered in a later inlining phase instead of always being inlined eagerly or rejected.

The inliner should also stop scheduling additional candidates when the root LLVM IR exceeds the configured instruction budget.

**Implementation you've considered**

Replace the boolean inline-policy callback with a decision-based callback that can return:

- `Deny`
- `InlineNow`
- `InlineLater`
- `HitNodeCountCutoff`

Propagate the late-inline state through the Jeandle inline policy and:

- delay selected inline candidates during the eager phase;
- retry delayed candidates after early optimization;
- allow `@ForceInline` methods exceeding normal depth or bytecode-size limits to use post-parse inlining;
- use a slightly larger instruction budget for late-inline candidates;
- record late-inline state in replay data and inline-tree metadata;
- notify the LLVM inline driver when the instruction cutoff is reached.

The corresponding LLVM implementation should consume these decisions and drive the eager and late inlining rounds.

**Additional context**

Intrinsic processing is handled before regular inlining and is outside the scope of this change.

`JeandleIncrementalInline` controls delayed handling of `@ForceInline` methods that exceed the normal depth or bytecode-size limits. Special-purpose delayed inlining may be controlled by its corresponding optimization flags.

The initial LLVM IR instruction cutoff is provisional and should be tuned using representative Jeandle workloads.

PR #606, opened 2026-08-27
feat: support late inlining in Jeandle

minchao1024:
## Related issue(s):

issue #605 

Replace the boolean inline-policy callback with a decision-based
interface that distinguishes immediate, delayed, denied, and
instruction-cutoff results.

Propagate late-inline state through the Jeandle inline policy and delay
selected string, boxing, vector, replay-directed, and oversized
@ForceInline candidates during eager inlining.

Track late-inline decisions in the inline tree and replay data, and
allow late candidates additional room in the root LLVM IR instruction
budget.

Report instruction-cutoff decisions to the LLVM inline driver so it
can stop eager scheduling or transition to the late-inline phase.

### DIFF
--- a/src/hotspot/share/jeandle/jeandleCompilation.cpp
+++ b/src/hotspot/share/jeandle/jeandleCompilation.cpp
@@ -407,9 +407,13 @@ llvm::Value* JeandleCompilation::find_or_insert_oop(ciObject* oop) {
   return global_oop_handle;
 }
 
-bool JeandleCompilation::over_inlining_cutoff() const {
+bool JeandleCompilation::over_inlining_cutoff(bool is_late_inline) const {
   assert(_llvm_module != nullptr, "llvm module must exist");
-  if (JeandleNodeCountInliningCutoff == 0) {
+  intx cutoff = JeandleNodeCountInliningCutoff;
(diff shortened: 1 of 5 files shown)


src/hotspot/share/jeandle/jeandleCompilation.cpp
```diff
@@ -897,21 +966,36 @@ bool JeandleInlineTree::try_to_inline(JeandleCompilation* comp,
...
   assert(caller != nullptr, "caller method must exist");
   reason = JeandleInlineReason::InlineHot;
+  should_delay = false;
+  hit_node_count_cutoff = false;
+
+  if (comp->over_inlining_cutoff(is_late_inline)) {
+    reason = JeandleInlineReason::NodeCountInliningCutoff;
+    hit_node_count_cutoff = true;
+    return false;
+  }
```
taofengliu, line 998: C2 seems do this check only when `callee_method->code_size() > MaxTrivialSize`.
And this maybe done after `jeandle_reject_or_delay_desired_method_limit` in `try_to_inline`, then we can mark more delayed callees in `jeandle_reject_or_delay_desired_method_limit`, like C2

Issue #610, opened 2026-09-01
Support folding of elements in stable arrays


yirenchuanmu:
**Goals**
1. Support folding of elements in stable arrays.
2. Passed jtreg tests: test/hotspot/jtreg/compiler/stable

**Description**

```
    @Stable
    static int[] array = {10, 20, 30, 40, 50};

    static int test() {
        return array[2];
    }

    public static void main(String[] args) {
        System.out.println(test());
        array[2] = 99;
        System.out.println(test());
        array[2] = 88;
        System.out.println(test());
    }
```
Since array is annotated with `@Stable`, the compiler treat array[2] as a stable value and fold it to the constant 30.
Therefore, test() always return 30.

PR #611, opened 2026-09-01
enhance: Support folding of elements in stable arrays

yirenchuanmu:
## Related issue(s):

issue #610

1. Support folding of elements in stable arrays.
2. Passed jtreg tests: test/hotspot/jtreg/compiler/stable

### DIFF
--- a/src/hotspot/share/jeandle/jeandleCompiledCode.cpp
+++ b/src/hotspot/share/jeandle/jeandleCompiledCode.cpp
@@ -677,6 +677,16 @@ int JeandleCompiledCode::parse_stackmap_prologue(StackMapParser::record_iterator
   return third.getSmallConstant();
 }
 
+void JeandleCompiledCode::record_stable_array(int oop_id, int dimension) {
+  int& old_dimension = _stable_array_dimensions[oop_id];
+  old_dimension = MAX2(old_dimension, dimension);
+}
+
+int JeandleCompiledCode::stable_array_dimension(int oop_id) const {
+  auto it = _stable_array_dimensions.find(oop_id);
+  return it == _stable_array_dimensions.end() ? 0 : it->second;
+}
+
 JeandleStackMap* JeandleCompiledCode::parse_stackmap(StackMapParser& stackmaps,
                                                      StackMapParser::record_iterator& record,
                                                      StackMapParser::RecordAccessor::location_iterator& location,
--- a/src/hotspot/share/jeandle/jeandleVMCallback.cpp
+++ b/src/hotspot/share/jeandle/jeandleVMCallback.cpp
@@ -112,15 +112,28 @@ uintptr_t record_klass_metadata(ciKlass* klass) {
   return reinterpret_cast<uintptr_t>(encoding);
 }
 
-bool constant_field(int oop_id, int offset, ciField** field, ciConstant* con) {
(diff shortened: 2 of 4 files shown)


src/hotspot/share/jeandle/jeandleCompiledCode.cpp
```diff
@@ -677,6 +677,16 @@ int JeandleCompiledCode::parse_stackmap_prologue(StackMapParser::record_iterator
   return third.getSmallConstant();
 }
 
+void JeandleCompiledCode::record_stable_array(int oop_id, int dimension) {
+  int& old_dimension = _stable_array_dimensions[oop_id];
+  old_dimension = MAX2(old_dimension, dimension);
+}
```
taofengliu, line 683: Do we need to record the array dimension in `JeandleCompilation::find_or_insert_oop` when the inserted oop is a stable array?

src/hotspot/share/jeandle/jeandleVMCallback.cpp
```diff
@@ -116,15 +116,28 @@ uintptr_t record_klass_metadata(ciKlass* klass) {
   return reinterpret_cast<uintptr_t>(encoding);
 }
 
-bool constant_field(int oop_id, int offset, ciField** field, ciConstant* con) {
+bool constant_field(int oop_id, int offset, ciField** field, ciConstant* con, int* stable_dimension) {
```
```suggestion
bool constant_field(int oop_id, int offset, ciField*& field, ciConstant& con, int& stable_dimension) {
```
Define it like this.

src/hotspot/share/jeandle/jeandleCompiledCode.cpp
yirenchuanmu: We don't need to record the array dimension in `find_or_insert_oop()`.

When loading the array object, we determine whether its stable dimension needs to be recorded.

In the first load, `%2` is the array object, and its stable dimension is recorded when the stable array conditions are met.
In the second load, `%7` is an array element, and the recorded stable dimension is used for constant folding via the array object's OopId.

```
// stable array conditions
if (FoldStableValues && field->is_stable() && field->type()->is_array_klass()) {
      stable_dimension = field->type()->as_array_klass()->dimension();
}

// LLVM IR
bci_0:                                            ; preds = %entry
  %0 = load ptr addrspace(1), ptr @oop_handle_java.lang.Class_0, align 8
  %1 = getelementptr inbounds i8, ptr addrspace(1) %0, i64 192
  %2 = load atomic ptr addrspace(1), ptr addrspace(1) %1 unordered, align 8, !java-klass !4, !java-klass-exact !5
  %3 = icmp eq ptr addrspace(1) %2, null
  br i1 %3, label %bci_4_null_check_fail, label %bci_4_null_check_pass, !make.implicit !5

bci_4_boundary_check_pass:                        ; preds = %bci_4_null_check_pass
  %array_element_base = getelementptr inbounds i8, ptr addrspace(1) %2, i32 24
  %array_element_address = getelementptr inbounds i32, ptr addrspace(1) %array_element_base, i32 2
  %7 = load atomic i32, ptr addrspace(1) %array_element_address unordered, align 4
  call hotspotcc void @jeandle.safepoint_poll() [ "deopt"(i64 0, i32 5, i32 5, i64 65546, i32 %7, i64 327695, ptr %OrigPcSlot) ]
  ret i32 %7
```

taofengliu (approved):

PR #612, opened 2026-09-02
fix: carry invokeBasic target signature

xiejf2020:

`_invokeBasic` may expose an adapted wrapper signature that differs from the actual `MethodHandle` target signature. Previously, CHA metadata reused the wrapper information, which could lead Jeandle to emit an incorrectly typed direct call. This change passes the concrete target method information through the VM callback so the compiler can reconstruct the call using the real target signature. It also adds a jtreg regression test covering a `void` wrapper invoking a non-`void` target. The test passes with a Jeandle fastdebug build.


src/hotspot/share/jeandle/jeandleVMCallback.cpp
```suggestion
      return {reinterpret_cast<uintptr_t>(target->holder()->constant_encoding()) | 1,
```
Fix this by the way.

xiejf2020: Adopted in eb5232d930ad22c6c3152b07769b91dcf88c439e. The new commit moves every writer (frontend DeclaredHolder and _linkTo*) to
constant_encoding and re-wraps the incoming Klass* through the CI layer at both VM callback entries, so the field now has a single contract. And CHAOptInfo won’t be unclear about responsibilities anymore, avoiding the mix-up between ciInstanceKlass and InstanceKlass.

taofengliu (approved):

PR #615, opened 2026-09-03
enhance: reimplement float16 conversion intrinsic

rcjhd:
### What changed
The Float16 intrinsics now use LLVM constrained floating-point casts:

  - experimental_constrained_fpext for float16ToFloat
  - experimental_constrained_fptrunc for floatToFloat16

  This replaces the previous hand-written integer NaN reconstruction path.

  ### Why

  The previous NaN-aware implementation generated additional integer operations and a select for every conversion, increasing latency and code size. The initial unconstrained FP-cast version could also be folded by LLVM and failed to preserve Java’s signaling-NaN behavior.

  Constrained casts keep the conversion boundary intact, allow the backend to select the target hardware conversion instructions, and keep NaN behavior consistent with the interpreter and C2 paths.

  ### Performance comparison

  Local x86_64, Intel i7-10700; JMH 1.37, 5 warmup iterations, 8 measurement iterations, 3 forks. Values are ns/op; lower is better.

|  Conversion  |  Current Jeandle on               |  Old NaN Jeandle on                 |       Jeandle off                |   C2 on (-UseSuperWord)|
|-----------| -----------| -----------| -----------| -----------|
|   f16 → f32, finite            |     0.743             |      1.011            |       2.981                |     0.442                |
|  f16 → f32, NaN            |    0.762             |      1.135            |        2.941               |       0.441               | 
|   f32 → f16, finite            |    0.885             |      1.312               |       3.722                 |      0.375                |  
|  f32 → f16, NaN             |     0.890              |        1.318            |         2.955              |      0.378                |         

  Compared with the old NaN arithmetic implementation, the new Jeandle path is approximately 26–33% faster. C2’s default results are much lower because the benchmark loop is vectorized. 

So the above C2 data is with vectorization disabled, the remaining gap is mainly due to loop-level optimization and ABI conversion overhead, rather than NaN handling.


rcjhd:
According to the LLVM spec (https://llvm.org/docs/LangRef.html#constrained-floating-point-intrinsics), functions that use constrained FP intrinsics are not allowed to contain any normal floating-point operations, and they must carry the strictfp attribute. That's a fair amount of overhead to deal with.

It also feels like a rather non-mainstream feature — still somewhat experimental — so using it carries some uncertainty and risk. After all, the only benefit was avoiding some unnecessary NaN semantic handling; the existing implementation sacrifices a little bit of performance, but compared to not enabling the intrinsics at all, it's already good enough.

I will close this PR.

PR #618, opened 2026-09-08
feat: add Class intrinsics

xiejf2020:
## Related issue(s):

issue #

This PR adds Jeandle intrinsic lowering for frequently used `java.lang.Class`
queries and adapts the lowering to the current Jeandle JavaOp and type-analysis
infrastructure.

### What this PR does

The following Class operations are lowered through Jeandle intrinsics:

- `Class.cast`
- `Class.isInstance`
- `Class.isAssignableFrom`
- `Class.getSuperclass`
- `Class.getModifiers`
- `Class.getClassAccessFlags`
- `Class.isArray`
- `Class.isPrimitive`
- `Class.isInterface`
- `Class.isHidden`

The implementation handles constant Class mirrors, primitive and `void`
mirrors, arrays, hidden classes, interfaces, unloaded classes, dynamic mirror
queries, null arguments, and class unloading.

A shared `klass_value_for_mirror` helper centralizes conversion from a Class
mirror to its corresponding Klass value. Static final Class mirrors are exposed
directly to Class intrinsic lowering, while unrelated static constants remain
normal loads so that downstream LLVM constant-field folding and type-analysis
passes can optimize them.

The frontend no longer duplicates generic constant folding already handled by
LLVM. Remaining frontend folds are limited to cases required for Java semantics
or intrinsic metadata construction.

### Why we need it

Class queries are common in reflection, serialization, framework dispatch, and
runtime type-checking paths. Without intrinsic lowering, these operations pay
the full Java call and metadata access overhead.

This change provides:

- lower latency for dynamic Class queries;
- direct handling of constant Class mirrors;
- better optimization opportunities for LLVM;
- a clear separation between Java semantic lowering and LLVM optimization;
- explicit coverage for primitive, hidden, unloaded, and class-unloading cases.

The implementation relies on the existing infrastructure in
`jeandle-llvm/main`, including Class mirror/Klass metadata folding and the
updated JavaType analysis. No additional LLVM patch is required in this PR.

### Testing

Fastdebug validation was performed with the matching LLVM configuration:

- LLVM: `MinSizeRel`, `LLVM_ENABLE_ASSERTIONS=ON`
- JDK: fastdebug

Results:

- `TestClassQueries.java`: passed
- `TestClassConstantUnloading.java`: passed
- LLVM constant-field-folding tests: 35/35 passed
- LLVM type-check-elimination tests: 7/7 passed

Release JDK and LLVM builds were also completed successfully.

### Performance

JMH was run with:

- release JDK and release LLVM;
- `ClassQueries` benchmarks;
- `kind=OBJECT`;
- one fork;
- two warmup iterations;
- three measurement iterations;
- single thread;
- unit: ns/op, lower is better.

| Benchmark | C2 | Jeandle intrinsic off | Jeandle intrinsic on |
|---|---:|---:|---:|
| Dynamic-query geometric mean (10 benchmarks) | 0.839 | 47.03 | 1.069 |
| `cast` | 1.604 | 45.71 | 1.444 |
| `isInstance` | 1.295 | 51.50 | 1.406 |
| `getSuperclass` | 1.328 | 46.72 | 1.505 |
| `constantCastObject` | 0.371 | 45.14 | 0.662 |
| `constantGetSuperclass` | 0.371 | 47.44 | 0.668 |
| `constantIsInstanceHit` | ≈0 | 44.55 | 0.354 |

With Class intrinsics enabled, the dynamic-query geometric mean improves by
approximately 44x compared with Jeandle with the Class intrinsics disabled.
It is approximately 27% slower than C2 on this dynamic subset.

Most constant Class queries are folded to near-zero cost. The remaining
non-folded constant cases are below 1 ns/op with the intrinsic enabled.

### DIFF
--- a/src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
+++ b/src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
@@ -34,11 +34,15 @@
 #include "jeandle/jeandleUtils.hpp"
 
 #include "jeandle/__hotspotHeadersBegin__.hpp"
+#include "ci/ciEnv.hpp"
 #include "ci/ciField.hpp"
+#include "ci/ciInstance.hpp"
 #include "ci/ciInstanceKlass.hpp"
+#include "ci/ciKlass.hpp"
 #include "ci/ciMethod.hpp"
 #include "ci/ciSignature.hpp"
 #include "ci/ciSymbol.hpp"
+#include "classfile/vmClasses.hpp"
 #include "classfile/javaClasses.hpp"
 #include "classfile/vmIntrinsics.hpp"
 #include "jeandle/jeandle_globals.hpp"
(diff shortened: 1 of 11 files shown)


src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
```diff
@@ -828,6 +859,605 @@ bool JeandleIntrinsicLowering::lower_java_op(const char* java_op_name,
...
+  if (load == nullptr) {
+    return nullptr;
+  }
+  llvm::Value* address = load->getPointerOperand()->stripPointerCasts();
+  llvm::GlobalVariable* handle = llvm::dyn_cast<llvm::GlobalVariable>(address);
+  if (handle == nullptr) {
+    return nullptr;
+  }
+  return _interp->_compiled_code.oop_for_handle_name(handle->getName());
+}
```
taofengliu: Optimization is difficult here because, after passing through a block with multiple predecessors, all values on the stack become PHI nodes. Further optimization requires analysis support from LLVM.
Therefore, I suggest removing this interface and all related optimizations.

src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
```diff
@@ -828,6 +859,605 @@ bool JeandleIntrinsicLowering::lower_java_op(const char* java_op_name,
...
+  if (mirror_type != nullptr) {
+    if (mirror_type->is_primitive_type()) {
+      return llvm::ConstantPointerNull::get(klass_type);
+    }
+    if (mirror_type->is_klass() && mirror_type->is_loaded()) {
+      return constant_klass_value(mirror_type->as_klass());
+    }
+  }
+  return _interp->call_java_op("jeandle.load_mirror_klass", {mirror});
+}
```
taofengliu: Optimization is difficult here because, after passing through a block with multiple predecessors, all values on the stack become PHI nodes. Further optimization requires analysis support from LLVM.
Therefore, I suggest removing this interface and all related optimizations.
Just use `jeandle.load_mirror_klass`.

taofengliu (commented):
We can do more folding in ConstantFoldingPass

PR #619, opened 2026-09-10
feat: implement equalsB and equalsC intrinsics

NeverRaR:
## Related issue(s):

issue #


src/hotspot/share/jeandle/templatemodule/template.ll
```diff
@@ -1330,4 +1330,34 @@ slow_path:
   ret void
 }
 
+define hotspotcc i32 @jeandle.memcmp(ptr noundef readonly captures(none) %0, ptr noundef readonly captures(none) %1, i64 noundef %2) "lower-phase"="1" #0 {
```
taofengliu, line 1333: Use addrspace(1) as pointer type.
Use vector instructions to optimize the comparison, or make sure it can be optimized by the auto vectorizer.

PR #623, opened 2026-09-14
feat: implement Object.notify and notifyAll

benshuai5D:
### 一. Implementation Architecture

Using Jeandle intrinsics:

```text
Object.notify() and Object.notifyAll()
    -> JeandleIntrinsicLowering::lower_object_notify()
    -> JeandleRuntimeRoutine::monitor_notify()
    -> quick_notify() fast path
    -> ObjectSynchronizer::notify() only if the fast path fails
```

The implementation ultimately reuses HotSpot's monitor implementation, keeping its behavior consistent with C2.

### 二. Tips

1. **Advantage over a regular call:** avoids part of the overhead of a regular native call.

2. **`ObjectSynchronizer::quick_notify` requires two pieces of information:**

   - `JavaThread`
   - receiver

   Therefore, the `lower_object_notify` method mainly handles obtaining the `JavaThread` and receiver.

3. `notify` may throw `IllegalMonitorStateException`, enter the VM, trigger a safepoint, and access monitor and Java object state. Therefore:

   ```cpp
   CallSiteAttributeMetadata attrs = {
       CTRL_NEEDS_EXCEPTION_EDGE,
       MEM_READ | MEM_WRITE | MEM_NEEDS_GC_STATE
   };
   ```

4. **Fast and slow paths of `notify()`:**

   - Fast path: `quick_notify()` directly operates on the monitor in the Java thread state without switching from Java to VM state, and requires that no safepoint occur during execution.
   - Therefore, `SafepointSynchronize::is_synchronizing()` is used to select the fast or slow path.

### 三. Performance Results

| Benchmark | Jeandle | Jeandle Disabled | C2 Original Fallback | C2 With Whitelist | Jeandle / Disabled | Jeandle / Fixed C2 |
|---|---:|---:|---:|---:|---:|---:|
| notifyOne | 112.19M | 14.66M | 14.38M | 117.26M | 7.65x | 95.68% |
| notifyAll | 112.20M | 14.72M | 14.46M | 113.37M | 7.62x | 98.97% |

> The C2 intrinsic list was missing support for `Object.notify` and `Object.notifyAll`, which is a defect in the community implementation.


src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
```diff
@@ -828,6 +836,29 @@ bool JeandleIntrinsicLowering::lower_java_op(const char* java_op_name,
...
+
+  llvm::Value* receiver = _interp->_jvm->peek_value(0).value();
+  static constexpr CallSiteAttributeMetadata attrs = {
+      CTRL_NEEDS_EXCEPTION_EDGE, MEM_READ | MEM_WRITE | MEM_NEEDS_GC_STATE};
+  llvm::FunctionCallee callee = notify_all
+      ? JeandleRuntimeRoutine::monitor_notify_all_callee(module)
+      : JeandleRuntimeRoutine::monitor_notify_callee(module);
+  emit_callsite(callee, llvm::CallingConv::Hotspot_JIT,
+                {receiver, current_thread}, attrs);
+  _interp->_jvm->apop();
```
taofengliu: I think we should pop the receiver before `emit_callsite`. Because the deopt bundle carries `should_reexecute = false`, we cannot leave the receiver in the stack if we don't reexecute.

src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
```diff
@@ -828,6 +836,29 @@ bool JeandleIntrinsicLowering::lower_java_op(const char* java_op_name,
...
 
+bool JeandleIntrinsicLowering::lower_object_notify(vmIntrinsics::ID id) {
+  const bool notify_all = id == vmIntrinsics::_notifyAll;
+  assert(notify_all || id == vmIntrinsics::_notify, "unexpected intrinsic");
+
+  llvm::Module& module = _interp->_module;
+  llvm::IRBuilder<>& builder = _interp->_ir_builder;
+  llvm::Function* current_thread_fn = module.getFunction("jeandle.current_thread");
+  assert(current_thread_fn != nullptr, "jeandle.current_thread JavaOp must exist");
+  llvm::CallInst* current_thread = builder.CreateCall(current_thread_fn);
```
taofengliu: Use `_interp->call_java_op`

src/hotspot/share/jeandle/jeandleRuntimeRoutine.hpp
```diff
@@ -54,6 +54,18 @@
...
+      llvm::Type::getVoidTy(context),\
+      llvm::PointerType::get(context, llvm::jeandle::AddrSpace::JavaHeapAddrSpace),\
+      llvm::PointerType::get(context, llvm::jeandle::AddrSpace::CHeapAddrSpace))    \
+                                                                                    \
+  def(monitor_notify_all,\
+      JeandleRuntimeRoutine::monitor_notify_all,\
+      llvm::Type::getVoidTy(context),\
+      llvm::PointerType::get(context, llvm::jeandle::AddrSpace::JavaHeapAddrSpace),\
+      llvm::PointerType::get(context, llvm::jeandle::AddrSpace::CHeapAddrSpace))    \
+                                                                                    \
```
taofengliu: Format `\`, Align the code in each column.

src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
benshuai5D: Thanks for pointing this out.

The receiver is now popped before emit_callsite(), keeping the deoptimization state consistent when should_reexecute is false.

src/hotspot/share/jeandle/jeandleIntrinsicLowering.cpp
benshuai5D: done，uses the existing `_interp->call_java_op("jeandle.current_thread", {})` helper instead of manually looking up the JavaOp.

src/hotspot/share/jeandle/jeandleRuntimeRoutine.hpp
benshuai5D: fixed，thanks

taofengliu (approved):

Issue #624, opened 2026-09-14
Build fails with `fatal error: llvm/ADT/DenseMap.h: No such file or directory` when using `--with-jeandle-llvm=~/...`


sendaoYan:
### Summary

Configuring with a `~`-prefixed path fails during `make images` because the
literal `~` is never expanded, so the compiler cannot find the LLVM headers.

```
* For target hotspot_variant-server_libjvm_objs_ciEnv.o:
In file included from .../src/hotspot/share/ci/ciEnv.cpp:56:
.../src/hotspot/share/jeandle/jeandleCompilation.hpp:25:10: fatal error: llvm/ADT/DenseMap.h: No such file or directory
   25 | #include "llvm/ADT/DenseMap.h"
```

### Environment

- OS: Linux (x86_64)
- Configure command:
  ```bash
  bash configure --with-jeandle-llvm=~/software/jeandle-llvm
  time make images test-image
  ```
- jeandle-llvm is installed and complete, e.g.:
  ```
  ~/software/jeandle-llvm/include/llvm/ADT/DenseMap.h   # exists
  ~/software/jeandle-llvm/bin/llvm-config --includedir  # -> .../include
  ```

### Root cause

Bash only performs tilde expansion when `~` is at the **start** of a word.
In `--with-jeandle-llvm=~/software/jeandle-llvm` the `~` follows `=`, so it is
not expanded and configure records the literal value:

```
build/linux-x86_64-server-release/spec.gmk:
  JEANDLE_LLVM_DIR := ~/software/jeandle-llvm
```

`make/hotspot/lib/JvmFlags.gmk` then emits `-I$(JEANDLE_LLVM_DIR)/include`,
i.e. `-I~/software/jeandle-llvm/include`, and the compiler (invoked by `make`
through `sh`) does not expand the mid-argument `~` either, so it looks for a
directory literally named `~`. The same problem would also hit the link step,
which references `$(JEANDLE_LLVM_DIR)/bin/llvm-config`.

### Related problem: non-existent path is silently accepted

There is also no validation that `--with-jeandle-llvm` actually points to an
existing directory. The only check is that the option is provided at all
(`make/autoconf/jvm-features.m4` errors only when `JEANDLE_LLVM_DIR` is empty).
So `--with-jeandle-llvm=/non-existent` passes configure without complaint and
only fails much later during the build (header not found / `llvm-config` not
found), which is a poor failure mode.

### Expected behavior

[...]

PR #625, opened 2026-09-14
fix: jeandle-jdk: expand '~' and validate existence for --with-jeandle-llvm

sendaoYan:
fix: expand tilde and validate path in --with-jeandle-llvm (#625)

## Related issue(s):

issue #624 

`--with-jeandle-llvm=~/...` stores the literal `~` because the shell does not expand a `~` that follows `=` on the command line. The build then passes `-I~/software/jeandle-llvm/include` to the compiler, which fails with `llvm/ADT/DenseMap.h: No such file or directory`.

This commit runs JEANDLE_LLVM_DIR and OPENJDK_BUILD_JEANDLE_LLVM_DIR through UTIL_FIXUP_PATH, which expands a leading `~`, makes the path absolute, and fails early at configure time when the directory does not exist (previously a non-existent path was silently accepted and only failed during the build).


sendaoYan:
Before this PR, pointing `--with-jeandle-llvm=` to a non-existent path does not cause `bash configure` to throw an error.

After this PR, pointing `--with-jeandle-llvm=` to a non-existent path will cause `bash configure` to throw an error.


image

taofengliu (approved):

PR #626, opened 2026-09-16
feat: implement StringUTF16.getChar/putChar intrinsics

rcjhd:

Implement StringUTF16.getChar/putChar intrinsics.

### 1. Implementation approach

- Uses an i8 GEP plus an offset to the base address of the array data area;
- Uses an i16 GEP plus the character index, with LLVM automatically scaling by 2 bytes;
- getChar uses a single i16 load, then zero-extends to i32;
- putChar truncates the input to i16, then uses a single i16 store;
- Retains explicit null checks and constant-access bailout.

Because StringUTF16 uses the target platform's native byte order, this implementation is consistent with the memory layout on AArch64/x86.

### 2. Something interesting

StringUTF16.getChar/putChar use trusted caller semantics, consistent with C2, and do not perform additional bounds checks.

The Java fallback was implemented in units of bytes:

Explicitly performed index << 1;
Loaded two i8 values separately and combined them manually;
Stored two i8 values separately and handled endianness manually.

The intrinsic implementation directly uses i16 access, avoiding manual byte splitting and combination. Compared with manual i8 implemention, the AArch64 machine code changed from:

lsl w10, w2, \# 1
add x10, x1, w10, sxtw
ldrh/strh ...

optimized to:

add x10, x1, w2, sxtw \# 1
ldrh/strh ...

Therefore, index scaling can be directly fused into the addressing instruction, and the machine code form is closer to C2.

### 3. Performance and conclusion

The test environment was an AArch64 release JDK, with JMH configured as 5 warmup iterations, 8 measurement iterations, 3 forks, and 1 second per round, using the default compiler blackhole.

|Benchmark |Jeandle Off |Jeandle On |Jeandle improvement |C2 Off| C2 On |C2 improvement|
|----|----|----|----|----|----|----|
|getCharDirect |1.130 ± 0.008 |0.831 ± 0.009 |26.5% |1.358 ± 0.070 |1.081 ± 0.027 |20.4%|
|putCharDirect |1.405 ± 0.008 |1.189 ± 0.005 |15.4% |1.435 ± 0.005 |0.961 ± 0.001| 33.0%|

The unit is ns/op; lower is better.

Conclusion:

The implementation significantly reduces the execution overhead of the Jeandle intrinsic; getChar's Jeandle performance is better than C2; putChar is still slower than C2, mainly due to C2's more compact control flow and safepoint/deoptimization layout;
The current modification resolves the lsl + add addressing fusion issue, **but explicit null checks in dynamic index scenarios are still not eliminated by ImplicitNullChecksPass.**


test/hotspot/jtreg/compiler/jeandle/intrinsic/TestStringUTF16CharAccess.java
```diff
@@ -0,0 +1,148 @@
...
+    private static final String PASSED = "TestStringUTF16CharAccess PASSED";
+    private static final String GET_NAME = "java.lang.StringUTF16.getChar";
+    private static final String PUT_NAME = "java.lang.StringUTF16.putChar";
+
+    public static void main(String[] args) throws Exception {
+        if (args.length == 0) {
+            run(true, null);
+            run(false, "-XX:ControlIntrinsic=-_getCharStringU,-_putCharStringU");
+        } else {
+            semantics(args[0].equals("disabled"));
```
taofengliu:
```suggestion
            semantics(args[0].equals("enabled"));
```

test/hotspot/jtreg/compiler/jeandle/intrinsic/TestStringUTF16CharAccess.java
```diff
@@ -0,0 +1,148 @@
...
+
+    private static boolean contains(Path path, String marker) {
+        try {
+            return Files.readString(path).contains(marker);
+        } catch (Exception e) {
+            return false;
+        }
+    }
+
+    private static void semantics(boolean exerciseMalformedCalls) throws Exception {
```
taofengliu:
```suggestion
    private static void semantics(boolean enabled) throws Exception {
```
This is more readable

test/hotspot/jtreg/compiler/jeandle/intrinsic/TestStringUTF16CharAccess.java
```diff
@@ -0,0 +1,148 @@
...
+        Asserts.assertEquals(0x1234, (int) (char) get.invoke(null, value, 1));
+
+        put.invoke(null, value, 0, 0xabcd);
+        Asserts.assertEquals(0xabcd, (int) (char) get.invoke(null, value, 0));
+
+        // The intrinsic follows C2's trusted-caller contract and does not
+        // perform bounds checks; malformed arguments are tested only through
+        // the disabled fallback path.
+        if (exerciseMalformedCalls) {
+            expectFailure(get, NullPointerException.class, null, 0);
```
taofengliu: Test expected `NullPointerException` when intrinsics are enabled, too.

taofengliu (commented):
The new JMH benchmark breaks the entire `test/micro` build.

`test/micro/org/openjdk/bench/java/lang/StringUTF16CharAccessIntrinsic.java` imports `java.lang.JeandleStringUTF16CharCaller`, a class that only exists in the jtreg tree (test/hotspot/jtreg/compiler/jeandle/intrinsic/java/lang/). However, `make/test/BuildMicrobenchmark.gmk` compiles all of test/micro (547 files) in a single SetupJavaCompilation invocation whose classpath contains only the JMH jars (JMH_COMPILE_JARS) — nothing from the jtreg tree is on it. The import therefore cannot resolve, and the whole build-microbenchmark target fails, so benchmarks.jar is no longer produced for any benchmark.

Side note (separate from this PR): a clean micro build is currently broken on main as well — #549's VectorizedMismatchIntrinsic.java needs --add-exports java.base/jdk.internal.misc=... and jdk.internal.util=... in BuildMicrobenchmark.gmk, which was never added. Incremental builds mask this by reusing the old batch markers, which is likely why this wasn't caught earlier.

test/hotspot/jtreg/compiler/jeandle/intrinsic/TestStringUTF16CharAccess.java
rcjhd: Revised.

test/hotspot/jtreg/compiler/jeandle/intrinsic/TestStringUTF16CharAccess.java
rcjhd: Revised.

test/hotspot/jtreg/compiler/jeandle/intrinsic/TestStringUTF16CharAccess.java
rcjhd: Done.

rcjhd:
> The new JMH benchmark breaks the entire `test/micro` build.
> 
> `test/micro/org/openjdk/bench/java/lang/StringUTF16CharAccessIntrinsic.java` imports `java.lang.JeandleStringUTF16CharCaller`, a class that only exists in the jtreg tree (test/hotspot/jtreg/compiler/jeandle/intrinsic/java/lang/). However, `make/test/BuildMicrobenchmark.gmk` compiles all of test/micro (547 files) in a single SetupJavaCompilation invocation whose classpath contains only the JMH jars (JMH_COMPILE_JARS) — nothing from the jtreg tree is on it. The import therefore cannot resolve, and the whole build-microbenchmark target fails, so benchmarks.jar is no longer produced for any benchmark.
> 
> Side note (separate from this PR): a clean micro build is currently broken on main as well — #549's VectorizedMismatchIntrinsic.java needs --add-exports java.base/jdk.internal.misc=... and jdk.internal.util=... in BuildMicrobenchmark.gmk, which was never added. Incremental builds mask this by reusing the old batch markers, which is likely why this wasn't caught earlier.

I directly removed `test/micro/org/openjdk/bench/java/lang/StringUTF16CharAccessIntrinsic.java`. It looks like it has too big an impact on the existing JMH build framework. After weighing the trade-offs, I don't think there's a particular need to keep this jmh use case. I made minimal changes to get existing `VectorizedMismatchIntrinsic.java` to compile. Now the microbenchmark can be built successfully.

taofengliu (approved):

Issue #559, opened 2026-07-24
Build fails on AArch64 environment with a high-version GCC.


rcjhd:
**Describe the bug**
I configure JDK with a high-version GCC in my aarch64 environment, which is
```
* C Compiler:     Version 12.3.1 (at /usr/bin/gcc)
* C++ Compiler:   Version 12.3.1 (at /usr/bin/g++)
```
However, it shows a build fail with error infos below:
```
In file included from /usr/include/features.h:503,
                 from /usr/include/bits/libc-header-start.h:33,
                 from /usr/include/stdint.h:26,
                 from /usr/lib/gcc/aarch64-openEuler-linux/12/include/stdint.h:9,
                 from /home/jihaodi/jeandle-jdk/src/hotspot/share/utilities/debug.hpp:34,
                 from /home/jihaodi/jeandle-jdk/src/hotspot/share/memory/allocation.hpp:29,
                 from /home/jihaodi/jeandle-jdk/src/hotspot/share/classfile/classLoaderData.hpp:28,
                 from /home/jihaodi/jeandle-jdk/src/hotspot/share/precompiled/precompiled.hpp:34:
In function 'ssize_t read(int, void*, size_t)',
    inlined from 'bool read_fully(const char*, char*, size_t)' at /home/jihaodi/jeandle-jdk/src/hotspot/os_cpu/linux_aarch64/vm_version_linux_aarch64.cpp:188:29,
    inlined from 'static void VM_Version::get_compatible_board(char*, int)' at /home/jihaodi/jeandle-jdk/src/hotspot/os_cpu/linux_aarch64/vm_version_linux_aarch64.cpp:218:19:
/usr/include/bits/unistd.h:28:10: error: argument 2 is null but the corresponding size argument 3 value is [1, 18446744073709551615] [-Werror=nonnull]
   28 |   return __glibc_fortify (read, __nbytes, sizeof (char),
      |          ^~~~~~~~~~~~~~~
/usr/include/bits/unistd-decl.h: In static member function 'static void VM_Version::get_compatible_board(char*, int)':
/usr/include/bits/unistd-decl.h:29:16: note: in a call to function 'ssize_t __read_alias(int, void*, size_t)' declared with attribute 'access (write_only, 2, 3)'
   29 | extern ssize_t __REDIRECT_FORTIFY (__read_alias, (int __fd, void *__buf,
      |                ^~~~~~~~~~~~~~~~~~
At global scope:
[...]

PR #628, opened 2026-09-16
fix: suppress false-positive GCC -Wnonnull in vm_version_linux_aarch64.cpp

sendaoYan:
## Related issue(s):

issue #559 

Building fastdebug on Linux aarch64 with openEuler's GCC 12.3.1 and glibc fortify hardening fails. glibc annotates `read()` with `__attribute__((access(write_only, 2, 3)))`, so the compiler requires a non-null buffer whenever the size argument is non-zero.

In `read_fully()` the buffer is only guarded by an `assert()` (which the compiler cannot use to prove non-nullness) and the pointer is derived from `snprintf()`'s return value, so GCC emits `-Wnonnull`, which is treated as an error in fastdebug builds (`WARNINGS_AS_ERRORS=true`).

The buffer is actually `_cpu_desc`, a static 4096-byte array that can never be null, so this is a false positive. Suppress the warning locally around the `::read` call with `PRAGMA_NONNULL_IGNORED`.

Clean backport of JDK-8339331, risk is low.


sendaoYan:
|包名|版本|来源|
|gcc（默认，已装）|12.3.1|update|
|gcc-toolset-14-gcc|14.2.1|everything|
|gcc-toolset-7-gcc|7.3.0|EPOL|

openeuler的官方yum源未提供gcc10版本下载，因此无法通过降级环境版本至gcc10来规避该问题

Issue #631, opened 2026-09-17
JVM aborts with "Error value was Success" when jeandle-llvm is built with -DLLVM_ENABLE_ASSERTIONS=ON


sendaoYan:
**Describe the bug**

A release build of jeandle-jdk crashes with SIGABRT when compiling a method
that has an exception handler table (any method with try/catch), for example
`java.lang.invoke.MethodHandle::<clinit>`:

```
> ./build/local-release/jdk/bin/java -Xcomp -version
Program aborted due to an unhandled Error:
Error value was Success. (Note: Success values must still be checked prior to being destroyed).
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGABRT (0x6) at pc=0x0000ffff8f8e1600 (sent by kill), pid=2688157, tid=2688171
#
# JRE version: OpenJDK Runtime Environment (21.0.7) (build 21.0.7-internal-adhoc.yansendao.jeandle-jdk)
# Java VM: OpenJDK 64-Bit Server VM (21.0.7-internal-adhoc.yansendao.jeandle-jdk, compiled mode, tiered, compressed oops, compressed class ptrs, g1 gc, linux-aarch64)
# Problematic frame:
# C  [libc.so.6+0x81600]
#
# Core dump will be written. Default location: /home/yansendao/git/jeandle-jdk/core.2688157
#
# An error report file with more information is saved as:
# /home/yansendao/git/jeandle-jdk/hs_err_pid2688157.log
#
# Compiler replay data is saved as:
# /home/yansendao/git/jeandle-jdk/replay_pid2688157.log
#
# If you would like to submit a bug report, please visit:
#   https://bugreport.java.com/bugreport/crash.jsp
#
Segmentation fault (core dumped)
```

[hs_err_pid2688280.log](https://github.com/user-attachments/files/32340546/hs_err_pid2688280.log)
[replay_pid2688280.log](https://github.com/user-attachments/files/32340545/replay_pid2688280.log)

**Minimal reproduce steps**

1. Build jeandle-llvm with `-DLLVM_ENABLE_ASSERTIONS=ON`.
2. Build jeandle-jdk as a release build.
3. Run:

       $ build/local-release/jdk/bin/java -Xcomp -Xms64M -Xmx64M -version

[build.sh](https://github.com/user-attachments/files/32340615/build.sh)

**Expected behavior**
[...]

PR #632, opened 2026-09-17
fix: always check DataExtractor cursor in exception handler table (#631)

sendaoYan:
## Related issue(s):

issue #631

- Replace assert(data_cursor, ...) with guarantee(data_cursor, ...) when decoding the .gcc_except_table section.
- Fixes a SIGABRT in release builds: compiling any method with an exception handler table aborted the VM with "Error value was Success".

Note: llvm::DataExtractor::Cursor holds an llvm::Error that starts as an unchecked success value and must be checked before destruction. The asserts were compiled out in release builds, so the cursor was destroyed with an unchecked llvm::Error. llvm::Error::~Error() aborts only when LLVM_ENABLE_ABI_BREAKING_CHECKS is enabled, which happens when jeandle-llvm is built with -DLLVM_ENABLE_ASSERTIONS=ON (the default is WITH_ASSERTS). guarantee() is always evaluated, so it both marks the error checked and keeps the parse validation active in release builds.


sendaoYan:
```bash
time for name in guarantee consumeError ; do for i in `seq 10` ; do { time numactl -N 1 -m 1 ./build/local-release/images/jdk-$name/bin/java -Xcomp -version; } &> build/version-$name-$i.log ; done ; don
```

The `build/local-release/images/jdk-consumeError /bin/java` is build from fix this issue by consume the llvm:Error by by `llvm::consumeError(data_cursor.takeError());` at the end of `JeandleCompiledCode::pd_build_exception_handler_table()`

Below `real` value get from time command shows that use guarantee instead of assert do not affect the performance.

```
yansendao@xuzhu:[jeandle-jdk]> grep real build/version-guarantee-* -h
real    0m19.367s
real    0m19.421s
real    0m19.539s
real    0m19.417s
real    0m19.324s
real    0m19.325s
real    0m19.308s
real    0m19.344s
real    0m19.379s
real    0m19.496s
yansendao@xuzhu:[jeandle-jdk]> grep real build/version-consumeError-* -h
real    0m19.443s
real    0m19.379s
real    0m19.820s
real    0m19.559s
real    0m19.510s
real    0m19.349s
real    0m19.550s
real    0m19.429s
real    0m19.445s
real    0m19.359s
```

sendaoYan:
riscv64 fix solution has been verified by build the linux-riscv64 jeandle-jdk and run `java -Xcomp -version`

```
qemu-riscv64-static build/local-riscv64-release/images/jdk/bin/java -Xcomp -version
openjdk version "21.0.7-internal" 2025-04-15
OpenJDK Runtime Environment (build 21.0.7-internal-adhoc.yansendao.jeandle-jdk)
OpenJDK 64-Bit Server VM (build 21.0.7-internal-adhoc.yansendao.jeandle-jdk, compiled mode)
```

PR #633, opened 2026-09-18
feat(intrinsics): implement System.currentTimeMillis and nanoTime in Jeandle

benshuai5D:
#### Performance Results

JMH used the existing `SystemTime` benchmark with CPU 8 pinned on NUMA node 0, one thread, 3 forks, 3 × 1-second warmup iterations, and 5 × 1-second measurement iterations.

| Benchmark | Jeandle Enabled | Jeandle Disabled | C2 | Jeandle / Disabled Throughput | Jeandle / C2 Throughput |
|---|---:|---:|---:|---:|---:|
| `currentTimeMillis` | 28.445 ± 0.414 ns/op | 50.948 ± 0.809 ns/op | 28.488 ± 0.222 ns/op | 1.791x | 100.2% |
| `nanoTime` | 28.667 ± 0.775 ns/op | 51.829 ± 1.005 ns/op | 28.858 ± 1.026 ns/op | 1.808x | 100.7% |

Compared with Jeandle with the intrinsics disabled:

- `currentTimeMillis` throughput improves by approximately `79.1%`.
- `nanoTime` throughput improves by approximately `80.8%`.

Compared with C2:

- `currentTimeMillis` reaches approximately `100.2%` of C2 throughput.
- `nanoTime` reaches approximately `100.7%` of C2 throughput.

#### Overall Design

- Implements Jeandle intrinsic support for `System.currentTimeMillis()` and `System.nanoTime()`.
- Consistent with C2, the intrinsics directly call the corresponding HotSpot OS runtime functions:
  - `currentTimeMillis` → `os::javaTimeMillis`
  - `nanoTime` → `os::javaTimeNanos`
- Both intrinsics share `lower_native_time_func()` and differ only in the runtime callee passed to it.
- The runtime calls use the C ABI and are marked as GC-leaf and `nounwind`. Their `i64` return values are pushed onto the JVM long operand stack.
- The calls use `MEM_OBSERVES_EXTERNAL_STATE` to prevent LLVM from treating them as pure functions and eliminating, merging, or reordering them.


test/micro/org/openjdk/bench/java/lang/SystemTimeIntrinsic.java
taofengliu, line 1: This may be duplicate with `test/micro/org/openjdk/bench/java/lang/SystemTime.java`

